PR #249, opened 2025-02-04
add issues from subconcepts to librarian concept pages

harrisonpim:
already deployed, eg https://djhoyps9k9bag.cloudfront.net/Q989.html

### DIFF
--- a/static_sites/concept_librarian/__main__.py
+++ b/static_sites/concept_librarian/__main__.py
@@ -1,4 +1,5 @@
 import shutil
+from collections import defaultdict
 from pathlib import Path
 from typing import Set
 
@@ -71,6 +72,37 @@ def main():
 
     console.log(f"Fetched {len(concepts)} concepts")
 
+    # Get recursive subconcepts for each concept
+    wikibase_id_to_subconcepts = defaultdict(set)
+    # Create lookup dict for faster access
+    wikibase_id_to_concept = {c.wikibase_id: c for c in concepts}
+
+    for concept in track(
+        concepts,
+        description="Getting recursive subconcepts",
+        transient=True,
+    ):
+        result = set()
+        wikibase_ids_to_process = set(concept.has_subconcept)
+        processed_wikibase_ids = set()
+
+        while wikibase_ids_to_process:
+            current_wikibase_id = wikibase_ids_to_process.pop()
+            if current_wikibase_id in processed_wikibase_ids:
+                continue
+
+            processed_wikibase_ids.add(current_wikibase_id)
+            if current_wikibase_id in wikibase_id_to_concept:
+                current_concept = wikibase_id_to_concept[current_wikibase_id]
+                result.add(current_concept)
+                # Add this concept's subconcepts to be processed
+                wikibase_ids_to_process.update(current_concept.has_subconcept)
+
+        wikibase_id_to_subconcepts[concept.wikibase_id] = result
+    console.log(
+        f"Built a list of subconcepts for {len(wikibase_id_to_subconcepts)} concepts"
+    )
+
     issues: list[ConceptStoreIssue] = []
     for check in [
         validate_related_relationship_symmetry,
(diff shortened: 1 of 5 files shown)


static_sites/concept_librarian/__main__.py
```diff
@@ -52,24 +51,63 @@ def get_affected_concept_ids(issues: list[ConceptStoreIssue]) -> Set[WikibaseID]
     return concept_ids
 
 
+def get_all_subconcepts(
+    concept_id: WikibaseID,
+    visited: set,
```
kdutia, line 56: **suggestion:** it's not super clear to me what the purpose of `visited` is here from this function alone. a docstring could be handy?

static_sites/concept_librarian/__main__.py
```diff
@@ -52,24 +51,63 @@ def get_affected_concept_ids(issues: list[ConceptStoreIssue]) -> Set[WikibaseID]
...
+
+    visited.add(concept_id)
+    result = set()
+
+    if concept_id not in wikibase_id_to_concept:
+        return result
+
+    current_concept = wikibase_id_to_concept[concept_id]
+
+    for subconcept_id in current_concept.has_subconcept:
```
kdutia, line 70: **suggestion:** the naming of `has_subconcept` here is unclear imo, as it returns multiple concept IDs. I saw that `related_concepts` is another property – could this be renamed to something like `subconcepts`?

static_sites/concept_librarian/__main__.py
```diff
@@ -52,24 +51,63 @@ def get_affected_concept_ids(issues: list[ConceptStoreIssue]) -> Set[WikibaseID]
...
-    concept_ids = wikibase.get_concept_ids()
-    for wikibase_id in track(
-        concept_ids,
-        description="Fetching all concepts from wikibase",
+    with console.status("Fetching all of our concepts from wikibase"):
+        concepts = wikibase.get_concepts()
+    console.log(f"✅ Fetched {len(concepts)} concepts from wikibase")
+
+    wikibase_id_to_recursive_subconcept_ids = defaultdict(lambda: set[WikibaseID]())
+    wikibase_id_to_concept = {c.wikibase_id: c for c in concepts}
```
kdutia, line 91: **suggestion:** this seems like a mapping that's useful in other places (I think I might have already used it or seen it used elsewhere?). 

could this be a `@property` in `WikibaseSession`?

static_sites/concept_librarian/__main__.py
```diff
@@ -85,10 +123,12 @@ def main():
         validate_concept_label_casing,
     ]:
         issues.extend(check(concepts))
-        console.log(f'Ran "{check.__name__}"')
+        console.log(f'🔬 Ran "{check.__name__}"')
 
     problematic_concepts = get_affected_concept_ids(issues)
-    console.log(f"Found {len(issues)} issues in {len(problematic_concepts)} concepts")
+    console.log(
+        f"❗ Found {len(issues)} issues in {len(problematic_concepts)} concepts"
```
kdutia, line 130: **polish:** could this error message call them problematic concepts, as the code does?

```suggestion
        f"❗ Found {len(issues)} issues in {len(problematic_concepts)} problematic concepts"
```

kdutia (approved):
looks great to me 😌 

as someone who's stepped away from this code a bit i had the benefit of looking at it with fresh eyes, and i've made some minor, non-blocking clarity suggestions

static_sites/concept_librarian/__main__.py
harrisonpim: good shout! will add

static_sites/concept_librarian/__main__.py
harrisonpim: this one is based on [the property name in wikibase](https://climatepolicyradar.wikibase.cloud/wiki/Property:P1) and it's eeeeeverywhere in our code at the moment. agree that it could be clearer, but i think that's something to address in a separate PR

static_sites/concept_librarian/__main__.py
harrisonpim: i think this might be too bulky to be a property (getting the concepts themselves involves making a big sequence of requests), but maybe it could be a method... will have a think

static_sites/concept_librarian/__main__.py
kdutia: sounds good to me!

static_sites/concept_librarian/__main__.py
kdutia: ah - when i wrote this i was imagining that `WikibaseSession` stores internal state of its concepts but realised that's not true. ignore me!

PR #250, opened 2025-02-05
Simplify structure of knowledge graph datasets in argilla

harrisonpim:
New system uses a single common knowledge-graph workspace, with datasets named by their QIDs. This makes it easier for us to push/pull those datasets because:

- no ambiguity about which workspace a concept's labelled passages will exist in
- no chance of datasets' names becoming out of sync when a concept's preferred label is updated

image

This PR includes:
- the script that I used to pull/re-push the datasets. We won't need to run it again, but maybe it's worth persisting it in the git history here, at least until we refactor the scripts directory.
- A few small quality of life improvements which reflect the new structure.

See [this thread in slack](https://climate-policy-radar.slack.com/archives/C05SASSJBMH/p1738670591698739) for a bit more nuance on potential adverse side effects from this migration.

I also ran the following to revoke access to the original copies of the datasets. 

```sh
workspaces=("risks" "sectors" "fossil-fuels" "ghgs" "policy-instruments" "impacted-groups" "impacted-groups-relabel")
users=("sarah" "sion" "henry" "roshan" "kyra" "tabea" "jesse" "mark" "fred" "julie" "erichcastro" "camilomartinez" "anne")

for workspace in "${workspaces[@]}"; do
    for user in "${users[@]}"; do
        argilla workspaces --name $workspace delete-user $user
    done
done
```

The original datasets are still there - I haven't deleted them, just in case. But most (non-admin) argilla users should now only see a list of QID datasets in the `knowledge-graph` workspace


mpjuhasz (approved):
Looks great 🚀

PR #252, opened 2025-02-10
Librarian check for graph depth and node descendants

mpjuhasz:
## What this PR does?
- adds a new check for validating the depth - descendant balance of a concept
- adds traversal methods to the graph, e.g. finding the longest depth, aggregating the number of descendants

The idea came from an issue whereby "Adaptation" got added below "Public Sector" resulting in way too many descendants for some concepts and bloating them. This check is meant to find these spurious links, by checking the depth / descendants ratio. 

image

NOTE: I've also fixed an issue with `EmptyConcept`. Its inheritence from `Concept` was breaking some downstream functions, so I've made various changes to get this working.

## How was this tested?
By running on all concepts in the concept store, and inspecting results -- it picks up the above example, but nothing else.

Furthermore, by regenerating the librarian locally, and incepting results.

### DIFF
--- a/static_sites/concept_librarian/checks.py
+++ b/static_sites/concept_librarian/checks.py
@@ -46,15 +46,17 @@ class MultiConceptIssue(ConceptStoreIssue):
 
 
 def format_concept_link(concept: Concept | EmptyConcept) -> str:
-    """Format a concept as an HTML link"""
-    style = "text-blue-600 dark:text-blue-400 hover:text-blue-800 dark:hover:text-blue-300 underline underline-offset-4"
-    display_text = (
-        concept.wikibase_id if isinstance(concept, EmptyConcept) else str(concept)
-    )
-    wikibase_url = concept.wikibase_url if isinstance(concept, Concept) else ""
-    return (
-        f"<a href='{wikibase_url}' target='_blank' class='{style}'>{display_text}</a>"
-    )
+    """
+    Format a concept as an HTML link
+
+    If the concept is an EmptyConcept, it will be formatted as a plain string.
+    """
+    if isinstance(concept, EmptyConcept):
+        return concept.wikibase_id
+    else:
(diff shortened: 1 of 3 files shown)


static_sites/concept_librarian/checks.py
```diff
@@ -338,3 +343,110 @@ def validate_concept_label_casing(
...
+    }
+
+    for concept in concepts:
+        id = concept.wikibase_id
+        assert id is not None, "Concepts should have a Wikibase ID"
+        depth = longest_depths[id]
+        n_descendants = number_of_descendants_map[id]
+
+        if depth != 0 and 200 / depth < n_descendants:
+            issues.append(concept)
```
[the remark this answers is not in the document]
mpjuhasz: Good question, I've added some documentation to clarify why 🙌

static_sites/concept_librarian/checks.py
```diff
@@ -372,8 +375,18 @@ def validate_concept_depth_and_descendant_balance(
...
+        # limit of this ratio for well-behaving concepts, which seemed to be bounded by
+        # n_descendants = 200 / depth, meaning that we're expecting e.g. max 200 descendants
+        # for a concept at level 1, and 40 for one at level 5.
         if depth != 0 and 200 / depth < n_descendants:
-            issues.append(concept)
+            issues.append(
+                ConceptIssue(
+                    concept=concept,
+                    issue_type="concept_depth_and_descendant_balance",
+                    message=f"{format_concept_link(concept)} is too deep in the hierarchy with {n_descendants} descendants",
```
harrisonpim, line 388: it might be better to frame this as more of a warning than a problem that _must_ be fixed

harrisonpim (approved):
looks great!

static_sites/concept_librarian/checks.py
mpjuhasz: Is there a way to flag a `ConceptIssue` as a warning? Or are you saying we shouldn't be returning a concept issue at all? I'm not sure I understand what you mean

PR #255, opened 2025-02-11
Labelling librarian

mpjuhasz:
> [!NOTE] 
> @harrisonpim and I have added a librarian for argilla akin to the concept one, and most of this is joint work through pairing sessions.

## What this PR does
Creates a new, shiny static site for our ✨ **Labelling Librarian** ✨ 

This includes:
- two pages:
    - landing page: contains all of the datasets with issues. This view is filterable for issue type. The datasets are sorted by the number of issues found. Upon clicking the dataset name, it takes the user to the dataset page
    - dataset page: more detailed view of all the dataset- and passage-level issues for the given dataset. The passage level issues are all presented, highlighting the issue type and the spurious span. Passage level issues are filterably by issue type.
- the following issue types by our own ideas and requests from the policy team (D for dataset-level and P for passage-level):
    - empty dataset (D): no labelled data in the dataset
    - too many discards (D): more than 5% of the passages are discarded
    - too few positives (D): only 20% of the passages contain positives
    - high non-alphabetic ratio (P): the span contains at least 50% non-alphabetic characters
    - long span (P): the span is longer than 50 characters
    - span border in word (P): the span's border is in the middle of the word

The infra has been created for this with Pulumi. 

## How was this tested
By serving the generated site locally.

### DIFF
--- a/static_sites/labelling_librarian/checks.py
+++ b/static_sites/labelling_librarian/checks.py
@@ -20,7 +20,7 @@ def dataset_to_labelled_passages_with_cache(
     if dataset_name in DATASET_CACHE:
         return DATASET_CACHE[dataset_name]
 
-    labelled_passages = dataset_to_labelled_passages(dataset, unescape_html=False)
+    labelled_passages = dataset_to_labelled_passages(dataset)
     merged_passages = create_gold_standard_labelled_passages(labelled_passages)
 
     DATASET_CACHE[dataset_name] = merged_passages


harrisonpim (approved):
hell yeah hell yeah 🕺

static_sites/labelling_librarian/README.md
```diff
@@ -0,0 +1,84 @@
+# Labelling Librarian
+
+A web-based tool that helps concept developers maintain the quality and consistency of our [labelled data](https://argilla.labs.climatepolicyradar.org/).
```
kdutia, line 3: as this is a public repo, do we want the url to our argilla instance to be in here?

static_sites/labelling_librarian/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+The site will be available at `http://localhost:8080`.
+
+## Deployment
+
+The site is deployed using AWS CloudFront and S3.
+
+### Deploy the infrastructure
+
+This step is only required if you're making changes to the infrastructure itself - skip it if you're just updating the site content, and move straight on to [Pushing new data to s3](#pushing-new-data-to-s3).
```
kdutia, line 33: to remove the cost of change later on, is it worth combining these instructions with the one for the concept librarian?

kdutia (approved):
🕺🏽 lgtm! some non-blocking thoughts

static_sites/labelling_librarian/README.md
harrisonpim: IMO including the url is fine - there are lots of methods which would allow a motivated person to discover subdomains of cpr.org (and lots of those are much more straightforward than reading through all of our documentation). if/when people do visit the site, access to the tool itself is auth'd so they would have to act in a much more malicious way to gain access... 👹

static_sites/labelling_librarian/README.md
harrisonpim: I think it's fine for the time being - IMO this is a case where being overly DRY would actually make it _harder_ to use these docs, rather than easier, given the differences with the vibe_check tool. Def worth revisiting to create a totally generic module when the number of these copy-pasted READMEs is higher though!

kdutia: ah cool, i didn't realise these instructions were different (i didn't check the others)

PR #258, opened 2025-02-12
Add setup for local test wikibase instance

olaughter:
Docker compose file with config and just commands for spinning up a test instance of wikibase. Resulting in this locally:

image

No data or tests against it yet, but this seemed like a good point to share back.

## Reference

This was setup based on an adaption of the advice found [here](https://github.com/wmde/wikibase-release-pipeline/blob/main/deploy/README.md), and through adapting the [linked dockerfile](https://github.com/wmde/wikibase-release-pipeline/blob/main/deploy/docker-compose.yml). These adaptations where done to make it easier to use from a purely test perspective.

### What adaptations where made?

1. To remove the need to get an ssl certificate when spinning up the stack. Our use for this instance is very much a local & ci instance, being able to access it without needing a secure connection means we don't need to bother letsencrypt for certificates every run (which likely would have rate limited us).
2. Portable domain resolution. The [guidance on local setup](https://github.com/wmde/wikibase-release-pipeline/blob/main/deploy/README.md#can-i-host-wbs-deploy-locally) suggests making changes on the host machine to resolve the domains. But making it so we can skip this step makes it easier to spin up on different machines and contexts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -177,3 +177,5 @@ scratch/
 
 
 *.DS_Store
+tests/local_wikibase/config/*
+!tests/local_wikibase/config/.gitkeep
--- a/justfile
+++ b/justfile
@@ -1,6 +1,7 @@
 set dotenv-load
 export WANDB_DIR := "./data/wandb"
 import "tests/local_vespa/local_vespa.just"
+import "tests/local_wikibase/local_wikibase.just"
 
 # Set the default command to list all available commands
 default:
--- a/tests/local_wikibase/README.md
+++ b/tests/local_wikibase/README.md
@@ -1,0 +1,14 @@
+# Local Wikibase
+
+## Setup for tests
+
+This folder has the config for a local wikibase instance that is used for testing.
+
+## Reference
+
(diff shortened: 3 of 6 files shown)


THOR300 (commented):
This I'm great, I'm just wondering whether we need to feed in some data into the local instance before considering this done? 

Just as I'm aware how many problems there were previously trying to represent our data in a local instance due to all the plugins required (perhaps our testing requirements are different though as we won't be editing the data) 🤔

olaughter:
> feed in some data into the local instance before considering this done

I'd much prefer to do that as a separate ticket, maybe even across several. Using this instance for testing the wikibase_to_s3 pipeline will have very different requirements to using it to test things like the concept librarian. And the question of how to feed in data still feels like quite an open question to me as there seems to be so many options to choose from. Also on top of that, deciding wether to copy our live wikibase data, or recreate a subset of it... Basically I think we can deal with data separately after this

THOR300 (approved):
LGTM - thanks for the reply re test data!

PR #260, opened 2025-02-17
deploy a prefect flow to rebuild and push the concept librarian site

harrisonpim:
Until now we've been manually running the commands to repopulate the concept librarian (and labelling librarian, and vibe-check) whenever we're asked by the policy team. That's boring and silly and inefficient - we could just have a computer run the commands on a schedule for us, so that the sites are never out of date.

This PR adds a new deployment to our prefect setup which runs:
- `just generate-static-site concept_librarian`
- `aws s3 sync static_sites/concept_librarian/dist s3://cpr-knowledge-graph-concept-librarian --delete`

every night at midnight. We'll see whether this is the right cadence based on feedback from the policy team.

At the moment, the deployment only builds/pushes the concept librarian, but it should be easy enough to extend to the other sites when we've shown that the approach works!


flows/deploy_static_sites.py
```diff
@@ -0,0 +1,67 @@
...
+
+    os.environ["WIKIBASE_PASSWORD"] = wikibase_password
+    os.environ["WIKIBASE_USERNAME"] = wikibase_username
+    os.environ["WIKIBASE_URL"] = wikibase_url
+
+
+@task
+def generate_static_site(app_name: str):
+    """Generate the static site using the existing command."""
+    subprocess.run(["just", "generate-static-site", app_name], check=True)
```
olaughter, line 32: **suggestion**: Don't consider this suggestion blocking, but given the just command itself calls a python file, would it be better to call the python directly from here instead?

olaughter (approved):
Nice!

flows/deploy_static_sites.py
harrisonpim: very good point! will amend

PR #261, opened 2025-02-17
add the labelling librarian to the static site generation flow

harrisonpim:
If all goes well, this should run the labelling librarian static site generation flow at the same time as the concept librarian one 🕺

I've added the two necessary params to SSM
- `/Argilla/APIKey`
- `/Argilla/APIURL`

The API key corresponds to the existing `prefect-bot-admin` user


mpjuhasz (approved):
Amazing 🚀 🎉

PR #263, opened 2025-02-19
Implementing changes to speed up indexing of KG pipeline

THOR300:
This Pull Request: 
---
- Is an attempted bugfix for our knowledge graph indexing pipeline. 
- The issue is that we are spinning up a host for every document and only yielding batches of 50 docs at a time in ecs which is leading to a very long ingest. (E.g. working on 50 docs, spinning up hosts for each, computing, spinning down and going again).
- The initial theory around what was causing this was around the unpacking of the generator on line 901 of index.py but I think the issue is actually that we are intending to run the process as a sub flow but are instead running as a deployment. We are also asynchronously running on individual documents as opposed to batches of documents. 

Theory: 
- We were using the parameter `as_subflow` in the `run_deployment` call, I believe we thought this meant it would run the process as a flow as opposed to a deployment. Looking at the prefect code I believe this just actually links the deployment and in the ui shows it as if it were a sub flow but still runs it as a deployment. 
- Run batches asynchronously in separate containers as opposed to creating batches and then unpacking them to run in parallel. 

Changes: 
- Attempted fix so we run as a flow / deployment when we intend to. 
- Parallelise on batches of documents in separate containers not documents. 
- Updates batch size: Batch size should be increased as this will run all the batches in parallel, e.g. if we have 10,000 [ docs ] / 50 [ docs / batch ] we'd have 200 indexers running in parallel.

Plan: 
- Merge this PR and kick off a staging or sandbox run once deployed. 
- Validate whether we are running as a flow or deployment. 

Diagram of the hoped improvement: 
image


flows/index.py
```diff
@@ -888,45 +934,33 @@ async def index_by_s3(
...
-
-                if isinstance(result, Exception):
-                    logger.error(
-                        f"failed to process document `{document_import_id}`: {str(result)}",
-                    )
-                    continue
-
-                logger.info(f"processed batch documents #{documents_batch_num}")
+        logger.info("gathering indexing tasks")
+        _ = await asyncio.gather(*indexing_tasks, return_exceptions=True)
```
harrisonpim, line 962: is it fine for the results of this gather to be discarded? is there any value in logging or retrying failures at this level?

harrisonpim (approved):
Approving to unblock 🚀 

I've done my best to pick apart and understand these changes, but i'm still a bit of a prefect n00b so i won't be offended if you want to wait for a more informed approval before merging/running :) though for what it's worth, these changes do seem sensible to me!

One minor question, mostly for my own learning

flows/index.py
THOR300: Good shout I'll add back in.

flows/index.py
```diff
@@ -888,45 +934,33 @@ async def index_by_s3(
...
+        logger.info("gathered indexing tasks")
 
 
 def run_partial_updates_of_concepts_for_document_passages_as(
```
THOR300, line 966: Revert

mpjuhasz (approved):
nice one 👌

jesse-c:
> * We were using the parameter `as_subflow` in the `run_deployment` call, I believe we thought this meant it would run the process as a flow as opposed to a deployment. Looking at the prefect code I believe this just actually links the deployment and in the ui shows it as if it were a sub flow but still runs it as a deployment.

FWIW this isn't right, IIRC. It wasn't trying to run it as a Flow. The control flow for running as genuinely a Deployment vs a function was to allow for testing.

PR #265, opened 2025-02-20
librarian flows need the awscli

harrisonpim:
Our librarian static site deployment flows are still failing because they can't run the `aws s3 sync` command 🤦‍♂️ 

This PR adds the AWS CLI to the dependencies in `pyproject.toml` so that it should be installed when the docker container is built.


mpjuhasz (approved):
LGTM

PR #267, opened 2025-02-20
labelling librarian check for low IAA

harrisonpim:
adds a dataset-level check for low inter-annotator agreement

also makes a couple of minor tweaks to the templates to ensure that dark mode styles are properly applied 😎


mpjuhasz (approved):
niice 🚀 🕺

PR #273, opened 2025-02-25
Catching raised exceptions when no passages for a document.

THOR300:
This Pull Request: 
---
- Catches raised exceptions, particularly when no passages are found for a document. 
- When we were running `run_partial_updates_of_concepts_for_batch` asynchronously the value errors that were thrown when there were no passages found for a document did not propagate and stop the program due to the use of the `asyncio.gather`. Now that we are awaiting the result the exception is being thrown and thus must be caught.
- I considered being tight on the exception handling and setting a ValueError, but then we might stop the entire program for another type of error and the prior solution didn't have granular exception handling so this is no step backwards.

### DIFF
--- a/flows/index.py
+++ b/flows/index.py
@@ -855,26 +855,25 @@ async def run_partial_updates_of_concepts_for_batch(
     """Run partial updates for concepts in a batch of documents."""
 
     logger = get_run_logger()
+    logger.info(
+        f"Updating concepts for batch of documents, documents in batch: {len(documents_batch)}."
+    )
     for i, document_importer in enumerate(documents_batch):
-        logger.info(
-            f"Updating concepts for batch of documents, {batch_size} indexing tasks for batch {documents_batch_num}."
-        )
+        try:
+            _ = await run_partial_updates_of_concepts_for_document_passages(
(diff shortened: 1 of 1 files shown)


harrisonpim (approved):

jesse-c:
> * When we were running `run_partial_updates_of_concepts_for_batch` asynchronously the value errors that were thrown when there were no passages found for a document did not propagate and stop the program due to the use of the `asyncio.gather`.

To clarify, if that's the issue, you can change `return_exceptions` to `False` [1][2]. This particular bit of 

[1] https://docs.python.org/3/library/asyncio-task.html#asyncio.gather
[2] https://github.com/climatepolicyradar/knowledge-graph/blob/main/flows/index.py#L623

> Now that we are awaiting the result the exception is being thrown and thus must be caught.

I'm confused by this, since the result was already being awaited?

Anyway, this change does seem okay, though just with a clarified understanding. We don't want 1 `DocumentImporter` in the batch taking down the rest of the batch. The batch is done sequentially, and not concurrently, so there's no gathering, and thus we could've used `return_exceptions=True` to not take down other `DocumentImporter`s in the batch.

PR #275, opened 2025-02-26
Removing flow decorator.

THOR300:
This Pull Request: 
---
- Removes the flow decorator from a function. 
- Below we are validating the size of the parameters before invoking a flow or deployment, thus this can't be a flow itself.


mpjuhasz (approved):
👌

PR #276, opened 2025-02-26
Feature/run deployments in batches to allow smaller document batches

THOR300:
This Pull Request: 
---
- Updates the indexing flow to run deployments / indexing tasks in batches. 
- E.g. run 20 indexers in parallel, wait for them all to finish and start a new batch. 

This is being trialled as we are experiencing some odd phenomena in prefect cloud where deployments / tasks are stuck in pending. I believe this may be due to the batch size being to big and causing the flow to fail as we exceed the max input data size. But am not totally sure (chasing up with prefect via an email)... 

The original plan was to get all the document_batches and just split them across hosts. Then each host would just process let's say 1/50th of the entire corpus rather than spinning up and down. Doing this however requires a large batch size of ~10,000 documents as we have so many tagged concepts in the labelled passages dir ~650,000. 

I've validated that we can run an ingest by running just one classifier which has resulted in concepts being returned by the backend. 

https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/7661cde2-b3e4-4909-bd4f-a6bdb08daf1b?entity_id=7661cde2-b3e4-4909-bd4f-a6bdb08daf1b


jesse-c:
> This is being trialled as we are experiencing some odd phenomena in prefect cloud where deployments / tasks are stuck in pending. 

Could we do trials in staging? You can use the tip I provided before our holidays, where you use https://github.com/climatepolicyradar/knowledge-graph/actions/workflows/prefect_deploy_staging.yml with your branch name. 

If I'm understanding correctly, production is currently broken for batches of at least > 1 size, where the deployments are stuck in pending? We could revert those changes, and do a (full) pipeline run, so as to unblock the Policy (and DSci?) team. Whilst it was slow before, it was working. Even if it takes 3 days, we could've started a (slow) run on Monday and it would've been done by now?

Then, we could do trial in staging (of batch sizes that aren't 1 document?), and then merge in to production once we know it works?

Does this sound like a reasonable approach?

> I believe this may be due to the batch size being to big and causing the flow to fail as we exceed the max input data size.

Prefect returns an error, if this is the case, so I would say this probably isn't the issue!

jesse-c:
From Anne:

> We'll go bug-hunting on Monday and Tuesday, assuming we can see all concepts on all document behind the feature flag by then. [1]

IMO this gives weight to doing a working, but slow run (after reverting the changes), so it's ready by Monday for them.

[1] https://climate-policy-radar.slack.com/archives/C078K2QNU0L/p1740652969504719

THOR300:
> From Anne:
> 
> > We'll go bug-hunting on Monday and Tuesday, assuming we can see all concepts on all document behind the feature flag by then. [1]
> 
> IMO this gives weight to doing a working, but slow run (after reverting the changes), so it's ready by Monday for them.
> 
> [1] https://climate-policy-radar.slack.com/archives/C078K2QNU0L/p1740652969504719

Yeh this was something I was considering, I feel we're so close though! This run ran in 18 mins, and this one 5 mins for 5500 and 350 labelled passages respectively! 

This also means worst comes to worst I just manually or via cli run one classifier at a time and we should still be much faster than prior. If it took 3.5 days before we'd be looking at roughly a week for the old indexing flow as we've added loads of concepts since then!

THOR300:
> > This is being trialled as we are experiencing some odd phenomena in prefect cloud where deployments / tasks are stuck in pending.
> 
> Could we do trials in staging? You can use the tip I provided before our holidays, where you use https://github.com/climatepolicyradar/knowledge-graph/actions/workflows/prefect_deploy_staging.yml with your branch name. A bonus too is that I found this a much faster dev iteration cycle than merging into production.
> 
> If I'm understanding correctly, production is currently broken for batches of at least > 1 size, where the deployments are stuck in pending? We could revert those changes, and do a (full) pipeline run, so as to unblock the Policy (and DSci?) team. Whilst it was slow before, it was working. Even if it takes 3 days, we could've started a (slow) run on Monday and it would've been done by now?
> 
> Then, we could do trial in staging (of batch sizes that aren't 1 document?), and then merge in to production once we know it works?
> 
> Does this sound like a reasonable approach?
> 
> > I believe this may be due to the batch size being to big and causing the flow to fail as we exceed the max input data size.
> 
> Prefect returns an error, if this is the case, so I would say this probably isn't the issue!

Ahh yes love that idea, I'll give that a go! Fingers crossed we have labelled passages there!

jesse-c:
> > From Anne:
> > > We'll go bug-hunting on Monday and Tuesday, assuming we can see all concepts on all document behind the feature flag by then. [1]
> > 
> > 
> > IMO this gives weight to doing a working, but slow run (after reverting the changes), so it's ready by Monday for them.
> > [1] https://climate-policy-radar.slack.com/archives/C078K2QNU0L/p1740652969504719
> 
> Yeh this was something I was considering, I feel we're so close though! [This](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/5e440463-c472-44b3-a739-5f1304374b17?entity_id=5e440463-c472-44b3-a739-5f1304374b17) run ran in 18 mins, and this one 5 mins for 5500 and 350 labelled passages respectively!

TBH I'm a little reticent since I spent quite a bit of time fixing up this pipeline just to get it working, when as a team we thought it was working. 😅 

> This also means worst comes to worst I just manually or via cli run one classifier at a time and we should still be much faster than prior. If it took 3.5 days before we'd be looking at roughly a week for the old indexing flow as we've added loads of concepts since then!

This could be done with it reverted too, no?

jesse-c:
>  Fingers crossed we have labelled passages there!

If not, you could run inference. There were labelled passages in there, that I was using, when I was using this development flow of using staging.

THOR300:
Deployed to staging and kicked off a run, looks like the fix is working: 
 https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/0a9ba6ca-a9ed-4915-ae06-44cb7d53ed1e?entity_id=0a9ba6ca-a9ed-4915-ae06-44cb7d53ed1e

flows/index.py
jesse-c, line 1: **fix:** In `run_partial_updates_of_concepts_for_batch_flow_or_deployment`, could you remove the parameter size log lines, since it causes lag in the Prefect webapp, and they were only added for debugging?

jesse-c, line 1: **polish:** In `run_partial_updates_of_concepts_for_batch_flow_or_deployment`, there's the `as_subflow` argument that is now used in a different way than originally added, IIRC, and there's the usage of the `as_subflow` argument to `run_deployment`, which I find confusing.

Could the argument to `run_partial_updates_of_concepts_for_batch_flow_or_deployment` be refactored to be something like `as_deployment`?

jesse-c, line 1: **polish:** With all these changes around batching, the `batch_size` argument to `run_partial_updates_of_concepts_for_batch` is now unused [1], could you please remove it?

[1] https://github.com/climatepolicyradar/knowledge-graph/blob/main/flows/index.py#L851

flows/index.py
```diff
@@ -42,6 +42,7 @@
 from src.span import Span
 
 DEFAULT_BATCH_SIZE = 500
```
jesse-c, line 44: **polish:** This batch size is now vague, and the new one is specific. Could this be renamed to be specific? The arguments that use it as a default value may want to be updated too, to be specific.
```suggestion
DEFAULT_DOCUMENTS_BATCH_SIZE = 500
```

jesse-c (commented):
It's taken me a minute to understand this. We have the first batching on all the documents [1] and then this new 2nd batching on top of that.

So we're limiting how many batches of documents we're running concurrently, right? 

Previously when I refactored this pipeline, 1 of my biggest gripes about it was that was confusing in nested batching.

Instead of nested batching, could we use a Prefect concurrency limit? We'd have it on `run_partial_updates_of_concepts_for_batch_flow_or_deployment` so that there's ≤ `|DEFAULT_INDEXING_TASK_BATCH_SIZE|` of them at one point in time?

I'd feel much happier with going with that solution and starting a big run for Monday.

[1] https://github.com/climatepolicyradar/knowledge-graph/blob/main/flows/index.py#L976

jesse-c:
Are you able to see successful runs? It just lags for me [1][2]. :(

[1]
CleanShot 2025-02-27 at 14 19 19
[2] https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/8c666087-315b-43df-b473-15a38f661b08?entity_id=8c666087-315b-43df-b473-15a38f661b08

THOR300:
> Are you able to see successful runs? It just lags for me [1][2]. :(
> 
> [1] CleanShot 2025-02-27 at 14 19 19 [2] https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/8c666087-315b-43df-b473-15a38f661b08?entity_id=8c666087-315b-43df-b473-15a38f661b08

So before I believe we weren't logging out these errors at the top level. 

image

Whereas now we are seeing error logs due to the `ValueErrors` raised when we can't find document passages in vespa. These are caught however and aren't "throwing" as it were. I think vespa just considers the flow a failure due to the error log. 

image

jesse-c:
Nice! I remember being able to see a majority of flows passing, but I think now it's not a flow any more. Anyway, yay for Prefect lagging.

THOR300:
> Nice! I remember being able to see a majority of flows passing, but I think now it's not a flow any more. Anyway, yay for Prefect lagging.

I'm seeing pretty similar results for the previous 3 day big prod run (similarly due to translated docs)

image

flows/index.py
THOR300: Yeh that's a great shout to rename, our usage of the parameter was effectively incorrect. It just links a deployment to the parent flow, it doesn't actuate whether to run as a flow or deployment. 

I can remove the parameter `as_subflow` from the `run_deployment` call as this basically defaults to true anyway.

THOR300:
> The small improvements look good. What do you think about the Prefect concurrency limit suggestion?

I think given we need to get this in to run before the weekend I would view it as a follow up PR! Running asnyc in batches seems to be working for now 💭

jesse-c (approved):
> > The small improvements look good. What do you think about the Prefect concurrency limit suggestion?
> 
> I think given we need to get this in to run before the weekend I would view it as a follow up PR! Running asnyc in batches seems to be working for now 💭

We don't necessarily have to. We can say that we need time. I can't emphasise enough how much I don't want to trend towards the state before https://github.com/climatepolicyradar/knowledge-graph/pull/226/files#diff-d8e21d60e20c4297775625ad008c76836b54f513f7e86a651872e2b70d10a30b, with nested batches. 😅

A follow-up PR, as in just after this one?

PR #278, opened 2025-02-27
Huggingface backup

mpjuhasz:
## What this PR does
- creating a prefect flow for backing up data into Huggingface
- add a new `HuggingfaceSession` class to manage these sessions
- transforming all of these datasets to the `LabelledPassage` format

After the v2 upgrade to argilla, we've faced the following issue: the settings of the datasets doesn't migrate properly from v1. This is not an issue for pulling the datasets and transforming them into our existing `LabelledPassage` objects. It does however cause issues, with using built-in argilla functionality that relies on this attribute being set. 
> NOTE: I'm excluding the struggles for trying to make these datasets work. This instead is a workaround.

Instead of trying to fix argilla's broken internal code, I worked around it in a way that relies on our own, controllable, and trustworthy classes. Hence, Argilla will ever only interface with `LabelledPassage` which is our central model for storing labelled data. This is turned into a HF dataset for each of the datasets.

## Important considerations / decisions:
- **cluttering our HF workspace**: I've tried organising them in collections but seems like that's a UI only feature. There's an open question whether this is something we care about, or accept -- my suggestion being that we do accept it.
- **backup has limited ability to restore Argilla service**: due to the intermediary formatting, this doesn't 100% substitute snapshots. I.e. we won't have a simple command to recreate our workspace (like we do with snapshots). We will however have all data retained in a very usable format. At the unlikely event of our argilla server failing, with some extra work this is also capable of restoring the dataset.
- **this is only for knowledge-graph datasets at the moment:** hence why I've put this in this repo. If we want this an argilla-wide service, that raises new questions about where this flow should live.

## How was this tested
Pushing and pulling dataset to and from HF successfully.


flows/data_backup.py
```diff
@@ -0,0 +1,39 @@
...
+def data_backup():
+    """Flow to deploy all our static sites."""
+    setup_environment()
+
+    hf_session = HuggingfaceSession()
+
+    datasets = get_all_datasets("knowledge-graph")
+    for dataset in datasets:
+        labelled_passages = dataset_to_labelled_passages(dataset)
+        hf_session.push(dataset.name, labelled_passages)
```
THOR300, line 35: **query**: Does this entirely overwrite the existing version of the dataset in huggingface? 

I.e. if we had removed some examples from a dataset and pushed to huggingface, would they be removed?

THOR300 (approved):
This is such a clean PR! 🙌

THOR300:
backup has limited ability to restore Argilla service

**query**: Could we not just push a version of the dataset to huggingface as is / in the original format; without converting to a `LabelledPassages` array and then unpacking into the dict?

flows/data_backup.py
mpjuhasz: the huggingface dataset is [version controlled](https://huggingface.co/docs/datasets/v1.8.0/share_dataset.html#dataset-versioning), so this practically just creates a new latest. We can still access the previous versions of the datasets by specifying a revision.

mpjuhasz:
> query: Could we not just push a version of the dataset to huggingface as is / in the original format; without converting to a LabelledPassages array and then unpacking into the dict?

@THOR300 so that's what I wanted to try first, and this is where the schema issue comes into play: to push the Argilla dataset object to the hub, we need to use their built-in method. This, however relies on pulling the fields out of the `dataset.settings` which contains the schema. This attribute didn't get populated during our migration, and is empty for all datasets, causing an error when pushing the dataset to the hub. 

I have tried backfilling the settings, because we know exactly what they should look like. This is forbidden however, as one cannot change the settings of an already published dataset. When I overwrite the settings locally, that helps with local transformations, but then I've faced another bug: there's an implicit assumption, that no two fields anywhere in the dataset can have the same name. In these datasets we have a metadata field called `text` and a top level field called `text`. When argilla tries flattening the dataset, these two conflict. Since this is a published dataset, neither the field-name, nor the metadata field could be overwritten, when I tried solving it that way. 

This is where I got to yesterday, when I started thinking about workarounds. @harrisonpim had a good suggestion which would resolve the last issue: by going back to argilla v1, and overwriting all datasets with an updated metadata-field name and then doing the migration again, we could resolve the name-conflict issue. This however involves transforming and updating the datasets in argilla v1, then remigrating again, and then we'd still have to use the local-overwrite trick for the v2 dataset settings.

Given that this felt more arduous, and doesn't guarantee that no other issues pop up after we remigrated the datasets, I've chosen the above workaround, which does hinder our ability to do an exact backup easily, bit it also moves us away from relying on Argilla code, which in light of all of the above, I considered a good thing.

flows/data_backup.py
THOR300: Thanks! That's great.

THOR300:
> > query: Could we not just push a version of the dataset to huggingface as is / in the original format; without converting to a LabelledPassages array and then unpacking into the dict?
> 
> @THOR300 so that's what I wanted to try first, and this is where the schema issue comes into play: to push the Argilla dataset object to the hub, we need to use their built-in method. This, however relies on pulling the fields out of the `dataset.settings` which contains the schema. This attribute didn't get populated during our migration, and is empty for all datasets, causing an error when pushing the dataset to the hub.
> 
> I have tried backfilling the settings, because we know exactly what they should look like. This is forbidden however, as one cannot change the settings of an already published dataset. When I overwrite the settings locally, that helps with local transformations, but then I've faced another bug: there's an implicit assumption, that no two fields anywhere in the dataset can have the same name. In these datasets we have a metadata field called `text` and a top level field called `text`. When argilla tries flattening the dataset, these two conflict. Since this is a published dataset, neither the field-name, nor the metadata field could be overwritten, when I tried solving it that way.
> 
> This is where I got to yesterday, when I started thinking about workarounds. @harrisonpim had a good suggestion which would resolve the last issue: by going back to argilla v1, and overwriting all datasets with an updated metadata-field name and then doing the migration again, we could resolve the name-conflict issue. This however involves transforming and updating the datasets in argilla v1, then remigrating again, and then we'd still have to use the local-overwrite trick for the v2 dataset settings.
> 
> Given that this felt more arduous, and doesn't guarantee that no other issues pop up after we remigrated the datasets, I've chosen the above workaround, which does hinder our ability to do an exact backup easily, bit it also moves us away from relying on Argilla code, which in light of all of the above, I considered a good thing.

Sounds like we've made the right decision considering the options! Backups as well have proven to be reliable via volume so we can always live with that solution as well in tandem with the huggingface backups!

kdutia (approved):
looks great - a couple of nonblocking comments

PR #281, opened 2025-03-04
Handle redirected concepts and add ancestry field

harrisonpim:
This PR:
- adds an optional ancestry field to a concept, which is populated with all of its parents, recursively. These should enable proper filtering in product.
- when fetching concepts, IDs for merged concepts will now be overwritten with their redirect targets

### DIFF
--- a/src/concept.py
+++ b/src/concept.py
@@ -38,6 +38,13 @@ class Concept(BaseModel):
             "disambiguate it from other concepts with similar labels"
         ),
     )
+    definition: Optional[str] = Field(
+        default=None,
+        description=(
+            "A more exhaustive definition of the concept, which should be enough for a "
+            "human labeller to identify instances of the concept in a given text. "
+        ),
+    )
     wikibase_id: Optional[WikibaseID] = Field(
         default=None, description="The Wikibase ID for the concept"
     )
@@ -51,8 +58,12 @@ class Concept(BaseModel):
     related_concepts: list[WikibaseID] = Field(
         default_factory=list, description="List of related concept IDs"
     )
-    definition: Optional[str] = Field(
-        default=None, description="The definition of the concept"
+    ancestry: Optional[list[WikibaseID]] = Field(
+        default=None,
+        description=(
+            "A list of all parent concept IDs, recursively, including that of the "
+            "concept itself"
+        ),
     )
     labelled_passages: list[LabelledPassage] = Field(
         default_factory=list,
--- a/src/wikibase.py
+++ b/src/wikibase.py
@@ -121,6 +121,7 @@ class WikibaseSession:
         wikibase_id: WikibaseID,
         timestamp: Optional[datetime] = None,
         include_labels_from_subconcepts: bool = False,
+        include_ancestry: bool = False,
     ) -> Concept:
         """
         Get a concept from Wikibase by its Wikibase ID
@@ -130,6 +131,7 @@ class WikibaseSession:
             If not provided, the latest version of the concept will be fetched.
         :param bool include_labels_from_subconcepts: Whether to include the labels
             from subconcepts in the concept
+        :param bool include_ancestry: Whether to include the concept's complete ancestry
         :return Concept: The concept with the given Wikibase ID
         """
 
@@ -267,7 +269,37 @@ class WikibaseSession:
             concept.alternative_labels = list(all_positive_labels)
             concept.negative_labels = list(all_negative_labels)
 
+        if include_ancestry:
+            concept.ancestry = self.get_ancestry(wikibase_id)
+
         return concept
+
+    def get_ancestry(self, wikibase_id: WikibaseID) -> list[WikibaseID]:
+        """Fetch the complete ancestry of a concept"""
+        response = self.session.get(
+            url=self.api_url,
+            params={
+                "action": "wbgetentities",
+                "format": "json",
+                "ids": wikibase_id,
+                "props": "claims",
+            },
+        ).json()
+
+        entity = response["entities"][wikibase_id]
+        ancestry = []
+        if "claims" in entity:
+            for claim in entity["claims"].values():
+                for statement in claim:
+                    if statement["mainsnak"]["snaktype"] == "value":
+                        property_id = statement["mainsnak"]["property"]
+                        value = statement["mainsnak"]["datavalue"]["value"]
+                        if property_id == self.subconcept_of_property_id:
+                            ancestry.append(value["id"])
+                            ancestry.extend(self.get_ancestry(value["id"]))
+
+        # Duplicates can occur in the ancestry, so they need to be removed
+        return list(set(ancestry))
 
     def get_concept_ids(self) -> list[WikibaseID]:
         """


src/concept.py
```diff
@@ -51,8 +58,12 @@ class Concept(BaseModel):
     related_concepts: list[WikibaseID] = Field(
         default_factory=list, description="List of related concept IDs"
     )
-    definition: Optional[str] = Field(
-        default=None, description="The definition of the concept"
+    ancestry: Optional[list[WikibaseID]] = Field(
```
kdutia, line 61: Is 'ancestry' a widely-used term? I needed to read this PR to figure out what it meant. If it's not, it could be replaced by something more descriptive like `recursive_parents`? (Not too happy with that name either, but i think it expresses the point i'm trying to make!)

src/wikibase.py
```diff
@@ -267,58 +272,123 @@ def get_concept(
...
+        if "claims" in entity:
+            for claim in entity["claims"].values():
+                for statement in claim:
+                    if statement["mainsnak"]["snaktype"] == "value":
+                        property_id = statement["mainsnak"]["property"]
+                        value = statement["mainsnak"]["datavalue"]["value"]
+                        if property_id == self.subconcept_of_property_id:
+                            resolved_id = self._resolve_redirect(value["id"])
+                            ancestry.append(resolved_id)
+                            ancestry.extend(self.get_ancestry(resolved_id))
```
kdutia, line 306: Is there potential that this could infinite loop if there is a cylical 'subconcept of' relationship?

src/wikibase.py
```diff
@@ -267,58 +272,123 @@ def get_concept(
...
-        # An extra precaution against infinite loops, suitable up to 1M concepts (500*2000)
-        MAX_PAGE_REQUESTS = 2000
+        MAX_PAGE_REQUESTS = 2000  # Suitable up to 1M pages (500*2000)
 
-        params = {
+        base_params = {
             "action": "query",
             "format": "json",
-            "list": "allpages",  # See https://www.mediawiki.org/wiki/API:Allpages
+            "list": "allpages",
```
kdutia, line 324: The link to API docs was removed from this. Personally I found it useful as Wikibase docs are hard to navigate but it could also live in the docstring? Wdyt?

src/wikibase.py
```diff
@@ -267,58 +272,123 @@ def get_concept(
...
+
+        :return dict[WikibaseID, WikibaseID]: The redirects, e.g
+            {"Q123": "Q456", "Q456": "Q789"}
+        """
+        pages = self._get_pages(extra_params={"apfilterredir": "redirects"})
+        redirects = {}
+
+        # For each redirect, we need to find the wikibase ids of the source and target.
+        # We process the pages in batches of 50
+        for batch in [pages[i : i + 50] for i in range(0, len(pages), 50)]:
```
kdutia, line 375: What's the reason that this is 50? Would we ever want to change it (maybe via a function kwarg)?

tests/local_wikibase/README.md
```diff
@@ -30,24 +30,6 @@ http://wdqs-frontend.localhost/
 
 http://quickstatements.localhost/
 
-### Programmatic Access
```
kdutia, line 33: what's the reason for removing these docs? do the docs need more attention?

kdutia (commented):
overall looks great! ✨  most are minor comments, but there's one about a potential infinite loop so not explicitly approving for that reason. (also sorry, i forgot conventional comments style commenting this time!)

src/concept.py
```diff
@@ -51,8 +58,12 @@ class Concept(BaseModel):
...
+        default=None,
+        description=(
+            "A list of all parent concept IDs, recursively, including that of the "
+            "concept itself"
```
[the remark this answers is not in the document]
harrisonpim: yeah, we could totally do this without! i included it because it matches the complete set of values that i expect will end up in a field in vespa for querying, but i agree, it makes the name a bit misleading. if others agree, i'll revert to the version without its own ID

tests/local_wikibase/README.md
harrisonpim: deleted because this was the only place that the get_all_properties() method was actually used! could replace this with a different example if you think it's needed?

src/wikibase.py
harrisonpim: good point - will re-add

src/concept.py
harrisonpim: it's not afaik, but i've used the word a few times in internal discussions and it always seems help make the idea click! will have a think about alternatives though

src/wikibase.py
harrisonpim: no specific reason tbh 🤷 it's a number, and it works! 

will add a kwarg with 50 as the default 🫡

src/concept.py
harrisonpim: i've replaced `ancestry` with `recursive_parent_concepts` for the time being - just shout if you have a better idea before this is merged!

tests/local_wikibase/README.md
kdutia: it could be worth replacing with something like "you can access the test instance using the `WikibaseSession` class", with a link to the implementation?

or just leave it - i feel like you have a much better sense of who will use which docs in this repo than I do! (and there are lots of great docs, and well-documented code, already)

src/concept.py
kdutia: lgtm!

src/wikibase.py
kdutia: i think the escape hatch can probably be built using a `max_number_of_levels` parameter with a sensible default? e.g. if it's 10, even if there's a loop it'll only try 10 times before stopping anyway

src/wikibase.py
```diff
@@ -161,6 +155,18 @@ def get_concept(
...
+            page_id_response.get("entities", {})
+            .get(wikibase_id, {})
+            .get("redirects", [])
+        )
+        if redirects:
+            logger.warning(
+                f"Made a request to {self.base_url} for {wikibase_id} but was "
+                f"redirected to {redirects.get('to')}"
+            )
+            wikibase_id = redirects.get("to")
```
jesse-c, line 169: **question:** This looks like a singular value, but above, for `"redirects"` it defaults to a list, instead of a dict that would have a `to` key?

jesse-c, line 169: **suggestion:** We could store the original/non-redirected ID, in case it helps with debugging.

src/concept.py
```diff
@@ -51,8 +58,9 @@ class Concept(BaseModel):
     related_concepts: list[WikibaseID] = Field(
         default_factory=list, description="List of related concept IDs"
     )
-    definition: Optional[str] = Field(
-        default=None, description="The definition of the concept"
+    recursive_parent_concepts: Optional[list[WikibaseID]] = Field(
```
jesse-c: **thought:** We lose some fidelity in having a flat list, e.g. a←b←c, a→d, becomes a has [b,c,d], where the lo-fi is that c is a grandparent. Would we be happier later on if we opt for a complex data structure that is hi-fi?

**question:** What is a "related" concept now (https://github.com/climatepolicyradar/knowledge-graph/blob/main/src/concept.py#L51-L53)?

We have subconcept of/has, related, ancestry relationships now and it's starting to get deep, which may be perfectly fine. How long until we ship a proper graph? Haha

src/wikibase.py
```diff
@@ -267,58 +273,133 @@ def get_concept(
...
+                "action": "wbgetentities",
+                "format": "json",
+                "ids": wikibase_id,
+                "props": "claims",
+            },
+        ).json()
+
+        entity = response["entities"][wikibase_id]
+        recursive_parent_concepts = [wikibase_id]
+        if "claims" in entity:
```
jesse-c, line 348: **thought:** Does Wikibase prevent infinite parents? E.g. a→b→c→a .. would be infinite? If not, you could keep track of visited concepts, and not visit them again (or possible another solution).

src/concept.py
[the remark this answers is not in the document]
harrisonpim: think your instincts were correct - i've removed the self concept ID from this field 🙂

jesse-c (approved):
LGTM

Will be interesting to see this pan out in different uses.

Could probably do with some unit testing, by separating out some bits, like the recursive bit. If it's separated from network effects, that would make it nicer to use (it may be already!)

src/wikibase.py
harrisonpim: great idea! i've made the change

the `wikibase._get_recursive_relationships()` method now takes a `max_depth` kwarg (default 50). at that depth, we should raise a warning and break the loop, returning the values accumulated so far

src/wikibase.py
jesse-c: I see you've covered this in https://github.com/climatepolicyradar/knowledge-graph/pull/281#discussion_r1979900434 now!

harrisonpim: @kdutia raised a similar point above! i've added a break clause if we hit a `max_depth`, which should return the values accumulated so far with a warning

src/wikibase.py
harrisonpim: great point! that's a vestigial `[]` from a previous version of the code 😅

i've updated it to default to a dict

harrisonpim: if we get stuck

src/concept.py
harrisonpim: > What is a "related" concept now?

`related_to` relationships are still valid! The two hierarchical relationship types (`has_subconcept`, `subconcept_of`) allow us to express super clean, fully entailed relationships (e.g. [carbon dioxide (Q223)](https://climatepolicyradar.wikibase.cloud/wiki/Item:Q223) is a subconcept of [greenhouse gas (Q218)](https://climatepolicyradar.wikibase.cloud/wiki/Item:Q218)). The `related_to` relationships allow us to capture more of the real-world, associative connections (e.g. the relationship between [greenhouse gas (Q218)](https://climatepolicyradar.wikibase.cloud/wiki/Item:Q218) and [the greenhouse effect (Q753)](https://climatepolicyradar.wikibase.cloud/wiki/Item:Q753))

These new recursive properties just extend the chain of what we can see from the strict hierarchies for a single concept.

> How long until we ship a proper graph?

yeah, i've been talking to @jamesgorrie about this a bit recently! It's a hard balance to strike, providing the relevant context without needing to pull the _entire_ graph every time you want to check on a single concept. IMO what we have here still gives us a decent local view of what's going on with a single concept, without needing to pull in everything.

On the specifics here, I think the flat list is all we need in order to enable our full sophisticated vespa filtering, so that's all i've tried to capture... If you have ideas/needs which go beyond that, I'm all ears!

harrisonpim:
great point about testing @jesse-c - i'm going to ram this through for now, but i'll come back with a bunch of tests when i've thought about how to implement them more carefully 🫡

src/concept.py
jesse-c: That's a reasonable approach for now!

PR #285, opened 2025-03-05
Refactor argilla v2 into `ArgillaSession`

mpjuhasz:
## What this PR does
We want to follow the same pattern with Argilla that we did for Huggingface, and Wikibase, our other two external data stores in this repo, namely that each of them have a `<service_name>Session` class, which is responsible for all transactions between the data store and the KG repo. This PR refactors all of the argilla v2 code into the `ArgillaSession` class which follows this pattern.

It also readds the data-backup flow deployment, now that the bug of trying to initialise the argilla client before pulling the credentials from ssm is fixed, and client connection is more safely managed.

### DIFF
--- a/deployments.py
+++ b/deployments.py
@@ -29,6 +29,7 @@ from flows.inference import (
     run_classifier_inference_on_batch_of_documents,
 )
 from flows.wikibase_to_s3 import wikibase_to_s3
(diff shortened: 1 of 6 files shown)


harrisonpim (approved):
🎉

PR #289, opened 2025-03-05
feat: Clarify behaviour around updates/removals

jesse-c:
This is part of understanding when updates/removals are happening.

TOWARDS PLA-481

### DIFF
--- a/flows/count_family_document_concepts.py
+++ b/flows/count_family_document_concepts.py
@@ -80,7 +80,14 @@ async def partial_update_family_document_concepts_counts(
     concepts_counts_with_names: dict[str, int],
     vespa_search_adapter: VespaSearchAdapter,
 ) -> None:
-    """Update document concept counts in Vespa via partial updates."""
+    """
+    Update document concept counts in Vespa via partial updates.
+
+    Similar to index.get_updated_passage_concepts, during the update
+    we remove all the old concepts related to a model. This is as it
+    was decided that holding out dated concepts counts on the document
+    in Vespa for a model is not useful.
+    """
 
     response: VespaResponse = vespa_search_adapter.client.update_data(
         schema="family_document",
@@ -88,7 +95,7 @@ async def partial_update_family_document_concepts_counts(
         data_id=document_import_id,
         fields={
             "concept_counts": concepts_counts_with_names
-        },  # Note the schema is misnamed in vespa
+        },  # Note the schema is misnamed in Vespa
     )
 
     if not response.is_successful():
@@ -142,7 +149,6 @@ async def load_update_document_concepts_counts(
     vespa_search_adapter: VespaSearchAdapter | None = None,
 ) -> dict[str, int]:
     """Load and aggregate concept counts from document URIs."""
-
     logger = get_run_logger()
 
     concepts_counts: Counter[Concept] = Counter()
@@ -190,6 +196,7 @@ async def load_update_document_concepts_counts(
 
     cm, vespa_search_adapter = get_vespa_search_adapter(vespa_search_adapter)
 
+    # Serialise them
     concepts_counts_with_names = {
         f"{concept.wikibase_id}{CONCEPT_COUNT_SEPARATOR}{concept.preferred_label}": count
         for concept, count in concepts_counts.items()


THOR300 (commented):
Great to push out the boiler plate code early! 

From what I can see your suggestion is that we wait until the additive indexing pipeline is done and then kick off a deindex pipeline to remove concepts from vespa that are no longer needed as well as removal of the labelled passage outputs? 

It might be good to have a quick design chat about this tomorrow morning to wizz over ideas. 

For the case where we've updated a classifier I was imagining that the way we would do this would be as follows: 
- During indexing we have the new concepts to add for a document passage as well as the existing concepts. 
- In my head we would just as part of indexing consolidation the new and current concepts into one concepts array and then set that in vespa whether that be by partial update or a complete overwrite of the data in the field?

For the case where we have completely deleted a concept from wikibase I haven't fully considered that scenario yet!

jesse-c:
> From what I can see your suggestion is that we wait until the additive indexing pipeline is done and then kick off a deindex pipeline to remove concepts from vespa that are no longer needed as well as removal of the labelled passage outputs?

It doesn't need to run after, it could run before, or at any point disconnected from indexing. I was picturing this:

> This would be triggered from the Wikibase to S3 backup automatically and by DSci after updating the classifier spec manually.

> It might be good to have a quick design chat about this tomorrow morning to wizz over ideas.

Sure!

> For the case where we've updated a classifier I was imagining that the way we would do this would be as follows:

This changeset isn't about the updated a classifier, but on a concept being removed in Wikibase and/or a concept not having a production-ready classifier. I have another changeset that will account for empty results, that will alter the inference and the indexing pipelines.

> * During indexing we have the new concepts to add for a document passage as well as the existing concepts.
> * In my head we would just as part of indexing consolidation the new and current concepts into one concepts array and then set that in vespa whether that be by partial update or a complete overwrite of the data in the field?

The existing indexing does something akin to [this at the moment](https://github.com/climatepolicyradar/knowledge-graph/blob/feat%2Fstore-no-inference-results/flows/index.py#L730). I'm still reading it to remember how it fully works.

IIRC Indexing doesn't have all the concepts at once for a document, since they're stored as a separate document-concept (`1:1`) instances, as opposed to document-concepts (`1:m`), so I don't think that would work? In the existing approach, it does merging of concepts. I need to remember how the [grouping](https://github.com/climatepolicyradar/knowledge-graph/blob/feat%2Fstore-no-inference-results/flows/index.py#L595-L598) works.

> For the case where we have completely deleted a concept from wikibase I haven't fully considered that scenario yet!

That's okay, I've got it covered in the Notion doc and this changeset!

THOR300:
> > * During indexing we have the new concepts to add for a document passage as well as the existing concepts.
> > * In my head we would just as part of indexing consolidation the new and current concepts into one concepts array and then set that in vespa whether that be by partial update or a complete overwrite of the data in the field?
> 
> The existing indexing does something akin to [this at the moment](https://github.com/climatepolicyradar/knowledge-graph/blob/feat%2Fstore-no-inference-results/flows/index.py#L730). I'm still reading it to remember how it fully works.
> 
> IIRC Indexing doesn't have all the concepts at once for a document, since they're stored as a separate document-concept (`1:1`) instances, as opposed to document-concepts (`1:m`), so I don't think that would work? In the existing approach, it does merging of concepts. I need to remember how the [grouping](https://github.com/climatepolicyradar/knowledge-graph/blob/feat%2Fstore-no-inference-results/flows/index.py#L595-L598) works.
> 

Yes great, what I was referring to was this function here where we combine the existing concepts from vespa with those from the labelled passages for the document / passage. 

In my head at that point we could update the logic to no longer be additive, clearing down existing concepts where a model matches those being added 🤔 

https://github.com/climatepolicyradar/knowledge-graph/blob/main/flows/index.py#L515

flows/index.py
```diff
@@ -1045,9 +1083,16 @@ async def index_labelled_passages_from_s3_to_vespa(
 
     logger.info(f"running with config: {config}")
 
-    if classifier_specs is None:
-        logger.info("no classifier specs. passed in, loading from file")
-        classifier_specs = parse_spec_file(config.aws_env)
+    match (classifier_specs, change):
```
THOR300, line 1086: **_confirmation:_**

So to confirm the `CHANGE.add` type does the following: 
- Update the concepts in vespa to reflect the latest version of the labelled passages, also including removal of any existing concepts for a model that are suitable for delete. 

And `CHANGE.remove` does the following: 
- Remove s3 labelled passages and concept counts for a classifier? 

**_thought:_** If we deleted a classifier:version would it potentially just be easier to remove all files at a particular prefix all at once after clearing down vespa?

jesse-c: Yup, that's it! I tried to lay all that out in the PR description. I missed the removing the concepts counts files though.

> thought: If we deleted a classifier:version would it potentially just be easier to remove all files at a particular prefix all at once after clearing down vespa?

Possibly!

jesse-c: Pushed a commit that accounts for removing concepts counts, and also renamed the enum value to be explicit (also, reflected in PR description).

jesse-c: Ugh, the tricky part is deleting concepts counts 😬

flows/index.py
```diff
@@ -1022,6 +1083,7 @@ async def index_labelled_passages_from_s3_to_vespa(
     classifier_specs: list[ClassifierSpec] | None = None,
     document_ids: list[str] | None = None,
     config: Config | None = None,
+    change: Change | None = Change.ADD_AND_UPDATE,
```
THOR300, line 1086: suggestion: Could we just implement a separate flow for classifier deletion? Judging by how deep we are passing the change param down, it could make our lives easier to just have a separate classifier removal flow that removals all occurrences from the concepts field of the document passage table and the concept counts field of the family table?

flows/index.py
```diff
@@ -649,30 +676,55 @@ async def run_partial_updates_of_concepts_for_document_passages(
...
-            serialised_concepts_counts = json.dumps(
-                {str(k): v for k, v in concepts_counts.items()}
-            )
+                match change:
+                    case Change.ADD_AND_UPDATE:
+                        concepts_counts.update(concepts_models)
+                    case Change.REMOVE:
+                        concepts_counts.subtract(concepts_models)
+
+                # Possibly remove the inference results from S3
```
THOR300, line 686: Yeah I'm not majorly worried about s3 removal at the moment, I can't seem to think of any worrying reason if they persist as we largely either rely on params passed into the flow or the config files for the model config 🤔

flows/index.py
```diff
@@ -526,27 +533,46 @@ def get_updated_passage_concepts(
     It is also, not possible to duplicate a Concept object in the concepts array as we
     are removing all instances where the model is the same.
     """
-    if not passage.concepts:
-        return [concept.model_dump(mode="json") for concept in concepts]
+    match change:
```
THOR300, line 536: suggestion: When we implement the work we split into two or more PR's

flows/index.py
jesse-c: That's how I originally started, with attempt no. 1, then I realised I can reuse all this work. I have thought about refactoring what would be shared, but more focused on getting something working. That could be done later.

flows/index.py
jesse-c: The default run isn't discerning, as in, we don't pass params, and it just grabs everything [1]. So we'd have to change that behaviour, if we didn't delete things?

[1] https://github.com/climatepolicyradar/knowledge-graph/blob/feat%2Fdeindex/flows/index.py#L1111-L1113

flows/index.py
jesse-c: Resolving as also brought up on [Slack](https://climate-policy-radar.slack.com/archives/C08F40KPLNR/p1741279239592469?thread_ts=1741193276.013729&cid=C08F40KPLNR).

flows/index.py
jesse-c: It's back to a separate, new pipeline!

flows/index.py
jesse-c: Resolving as I don't think this is doable without those changes! Happy to consider otherwise though 🙇🏻

flows/count_family_document_concepts.py
```diff
@@ -80,15 +80,22 @@ async def partial_update_family_document_concepts_counts(
     concepts_counts_with_names: dict[str, int],
     vespa_search_adapter: VespaSearchAdapter,
 ) -> None:
-    """Update document concept counts in Vespa via partial updates."""
+    """
+    Update document concept counts in Vespa via partial updates.
```
THOR300, line 84: comment: Nice that this effectively already has the update/delete behaviour (e.g. removing old count values)!

THOR300 (approved):
LGTM

PR #290, opened 2025-03-06
Adding a limit in to the passage query.

THOR300:
This Pull Request: 
---
- Is a bug fix for the indexer where we were listing out passages for documents from vespa with a default limit of 5000 hits. 
- Some passages have more than 5000 hits as per the example in the linked linear issue where we were receiving 10922 hits. 
- 50,000 is the maximum number of hits that we allow for a response so setting to that as we don't want to be accidentally hitting up against this limit again. 
- I've checked over the KG pipeline and this is the only place we run a `select * ....` vespa query and thus we shouldn't need to make any other query updates to set limits. 

50,000 is the maximum we set ourselves for the deployment: 
- [Infra](https://github.com/climatepolicyradar/navigator-infra/blob/main/vespa/navigator_app/search/query-profiles/default.xml#L3).
- Confirmed in this [ticket](https://linear.app/climate-policy-radar/issue/PLA-480/confirm-whether-we-have-documents-in-vespa-with-over-50000-passages) that some documents do exceed 50,000 so I'm chasing up a fix with vespa.

How was this tested? 
- Did a run in [staging](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/e2c78418-4b4a-456c-a2c7-1325f11e2300?entity_id=e2c78418-4b4a-456c-a2c7-1325f11e2300) on the document and verified that we see the error there. 
- [Deployed](https://github.com/climatepolicyradar/knowledge-graph/actions/runs/13696195270) the bug fix PR to staging. 
- Reran the indexing [flow](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/5ec98532-471e-460f-b4da-80fc8ee33b20?entity_id=5ec98532-471e-460f-b4da-80fc8ee33b20) and no errors were seen.
- Validated via a vespa query that we have data.

PR #291, opened 2025-03-06
Feature/pla 482 investigate whether we can fix the max hits issue by

THOR300:
This Pull Request: 
---

> Is an alternative to the following [PR](https://github.com/climatepolicyradar/knowledge-graph/pull/290) where we query vespa on text block id rather than upping the limit to 50,000 hits as we have documents that can be 500k passages in length! 

- Is a follow up on a bug where we were hitting the maximum number of hits for a query. 
- The PR aims to get around this by no longer querying vespa for all of the passages of a document as some of these can have as many as half a million hits. 
- Instead we retrieve the passage that we are looking to update one at a time. 
- Bumps pyvespa so we can use the query builder for a nicer query definition. 

Testing: 
- Deployed to staging and kicked off a [run](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/6b5a4fa4-339e-4a6b-ae61-10b3c8f175bf?entity_id=6b5a4fa4-339e-4a6b-ae61-10b3c8f175bf) for a document. Successfully passed with no errors.
- Lowered the max hits to prove that we are iterating via text blocks and [reran](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/5829397c-fd9b-4582-81bd-420d9c937cfd?entity_id=5829397c-fd9b-4582-81bd-420d9c937cfd).


flows/index.py
```diff
@@ -364,6 +365,55 @@ def get_document_passages_from_vespa(
     ]
 
 
+def get_document_passage_from_vespa(
+    text_block_id: str,
+    document_import_id: DocumentImportId,
+    vespa_search_adapter: VespaSearchAdapter,
+    limit_hits: int = 5000,
```
THOR300, line 372: thought: We could remove this parameter altogether now...

flows/index.py
```diff
@@ -13,6 +13,7 @@
 from typing import Any, ContextManager, TypeAlias, TypeVar
 
 import boto3
+import vespa.querybuilder as qb
```
jesse-c, line 16: **praise:** Nice to see this being used!

flows/index.py
```diff
@@ -364,6 +365,52 @@ def get_document_passages_from_vespa(
...
+        f"id:doc_search:family_document::{document_import_id}"
+    ) & qb.QueryField("text_block_id").contains(text_block_id)
+
+    yql = qb.select("*").from_("document_passage").where(condition)
+
+    vespa_query_response: VespaQueryResponse = vespa_search_adapter.client.query(
+        body={
+            "yql": yql.__str__(),
+        }
+    )
```
jesse-c, line 390: **polish:** From looking at https://pyvespa.readthedocs.io/en/latest/query.html#Example-1---matches,-order-by-and-limit, I _think_ we can simplify this?
```suggestion
    yql = qb.select("*").from_("document_passage").where(condition)

    vespa_query_response: VespaQueryResponse = vespa_search_adapter.client.query(yql=yql)
```

flows/index.py
```diff
@@ -364,6 +365,52 @@ def get_document_passages_from_vespa(
...
+
+    if (status_code := vespa_query_response.get_status_code()) != HTTP_OK:
+        raise QueryError(status_code)
```
jesse-c, line 393: **polish:** Fred put me onto this built-in helper [1][2]:

```suggestion
    if not response.is_successful():
        raise QueryError(response.get_status_code())
```

[1] https://pyvespa.readthedocs.io/en/latest/query.html#Error-handling
[2] https://github.com/climatepolicyradar/knowledge-graph/blob/feat%2Fdeindex/flows/count_family_document_concepts.py#L94-L95

flows/index.py
```diff
@@ -712,21 +693,37 @@ def get_vespa_search_adapter(
...
+    Extract non-schema namespaced ID (last element after "::")
+
+    Example:
+    "CCLW.executive.10014.4470.623" from document passage ID like
+    "id:doc_search:document_passage::CCLW.executive.10014.4470.623".
+    """
+    splits = hit_id.split("::")
+    if len(splits) != 2:
+        raise ValueError(f"received {len(splits)} splits, when expecting 2: {splits}")
+    return splits[1]
```
jesse-c, line 707: **fix:** There's a type alias for this!
```suggestion
def get_data_id_from_vespa_hit_id(hit_id: VespaHitId) -> VespaDataId:
    """
    Extract non-schema namespaced ID (last element after "::")

    Example:
    "CCLW.executive.10014.4470.623" from document passage ID like
    "id:doc_search:document_passage::CCLW.executive.10014.4470.623".
    """
    splits = hit_id.split("::")
    if len(splits) != 2:
        raise ValueError(f"received {len(splits)} splits, when expecting 2: {splits}")
    return splits[1]
```

flows/index.py
```diff
@@ -712,21 +693,37 @@ def get_vespa_search_adapter(
...
+
+
 async def partial_update_text_block(
     text_block_id: TextBlockId,
-    document_passages: list[tuple[VespaHitId, VespaPassage]],
     concepts: list[VespaConcept],
+    document_import_id: str,
```
jesse-c, line 713: **fix:** Another type alias to be used:
```suggestion
    document_import_id: DocumentImportId,
```
Before these were added, it was a bunch of tuples and strings, that made it hard to understand.

jesse-c (commented):
LGTM. Just some small changes first! 🙏🏻

flows/index.py
```diff
@@ -384,13 +384,11 @@ def get_document_passage_from_vespa(
...
-        body={
-            "yql": yql.__str__(),
-        }
+        yql=yql
     )
 
-    if (status_code := vespa_query_response.get_status_code()) != HTTP_OK:
-        raise QueryError(status_code)
+    if not vespa_query_response.is_successful():
+        raise QueryError(vespa_query_response.status_code)
```
jesse-c, line 391: **quibble:** It would be good to still use the `get_status_code` [1] function, so as to not reach into an "internal" field, since presumably they've added that function as an entry point for some reason? But maybe not! 😅 

[1] https://github.com/vespa-engine/pyvespa/blob/76205515ece4a6dfbc4d34b15f745e9dafd277d2/vespa/io.py#L28

jesse-c (approved):
LGTM

Thanks for the polish and fixes!

PR #295, opened 2025-03-10
Use the correct ID for "healthcare sector"

harrisonpim:
The healthcare sector classifier was attached to an old, redirected ID. I've trained a new one with the appropriate ID, which should make the results discoverable in product after re-running inference and indexing. 

This PR updates the classifier spec files, and makes a small tweak to the train script to use the new recursive relationship functionality when fetching the concept from wikibase.


THOR300 (approved):
Looks great, as a validation you could manually trigger an inference run in sandbox, labs or staging in prefect to confirm that the classifier:version is outputting sensible results.

PR #296, opened 2025-03-10
Only run inference on english docs and write out with correct file stem 

THOR300:
This Pull Request: 
---
Is a fix for the following:
- We are running inference on english and non english documents. 
- This means that inference is taking longer than required and is resulting with us indexing incorrect concepts that were for example run over french documents where the file stem matches the document id but the language is not english. 
- We are also saving translated documents to a file name with the suffix translated_en which means they are not being correctly indexed. 

Please see a diagram of the problem at this link: 

 [POL-2385](https://linear.app/climate-policy-radar/issue/POL-2385/classifiers-not-working-on-translated-documents#comment-a3d1103a)

### DIFF
--- a/flows/inference.py
+++ b/flows/inference.py
@@ -9,7 +9,6 @@ from typing import Final, Optional, Set, Tuple, TypeAlias
 
 import boto3
 import prefect.artifacts as artifacts
-import wandb
 from botocore.client import ClientError
 from cpr_sdk.parser_models import BaseParserOutput, BlockType
 from cpr_sdk.ssm import get_aws_ssm_param
@@ -18,9 +17,9 @@ from prefect.concurrency.asyncio import concurrency
 from prefect.deployments import run_deployment
 from prefect.task_runners import ConcurrentTaskRunner
 from pydantic import SecretStr
-from wandb.sdk.wandb_run import Run
 
-from flows.utils import SlackNotify
+import wandb
+from flows.utils import SlackNotify, remove_translated_suffix
 from scripts.cloud import (
     AwsEnv,
     ClassifierSpec,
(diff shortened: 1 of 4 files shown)


flows/inference.py
```diff
@@ -393,7 +399,7 @@ async def run_classifier_inference_on_document(
         store_labels(
             config=config,
             labels=doc_labels,
-            file_stem=file_stem,
+            file_stem=remove_translated_suffix(file_stem),
```
THOR300, line 402: @mpjuhasz has the concern here that we lose track of the source document.

THOR300: Idea that we could update the indexer to accept _translated_en documents and index these by stripping the suffix.

THOR300: If we did want to add a check to the indexer that the document had the correct language we could load it from s3.

THOR300: OR update the labelled passages object.

THOR300: OR prune and trust the above logic.

mpjuhasz (approved):
looks good 👌

flows/index.py
```diff
@@ -587,13 +590,13 @@ async def run_partial_updates_of_concepts_for_document_passages(
         for batch_num, batch in enumerate(batches, start=1):
             logger.info(f"processing partial updates batch {batch_num}")
 
+            document_import_id = remove_translated_suffix(document_importer[0])
```
jesse-c, line 593: **question:** Why's this needed? Is it since until now, it was reading from the English translation file, and in Vespa, it's just th ID without that suffix that's needed? I think that a comment here would be helpful to anyone reading this code in the future.

jesse-c (approved):

flows/index.py
THOR300: Essentially the document importer contains the file stem. This could be like "CCLW.exec.1.1_translated_en".

We need to remove the suffix to get the document id as this is what we use to query vespa (e.g. passages will only match on CCLW.exec.1.1.

PR #297, opened 2025-03-10
Refactoring document id to file stem in the inference pipeline

THOR300:
This Pull Request: 
---
- Is a refactor to update the naming of document id in the **_INFERENCE_** pipeline to being file stem where applicable. 
- This is as we are referring to objects as documents ids when they are not which has lead to processing [issues](https://linear.app/climate-policy-radar/issue/POL-2385/classifiers-not-working-on-translated-documents#comment-a3d1103a). 
- Why use file stems? This is the attribute that is used in the `Path` object, e.g.:

```python
>>> Path("s3://..../file.json").stem
>>> "file"
```


flows/inference.py
```diff
@@ -153,30 +158,35 @@ def get_latest_ingest_documents(config: Config) -> list[str]:
...
         )
     elif use_new_and_updated:
         requested_document_ids = get_latest_ingest_documents(config)
     elif requested_document_ids is None:
-        return current_bucket_ids
+        return current_bucket_file_stems
 
-    missing_from_bucket = list(set(requested_document_ids) - set(current_bucket_ids))
+    missing_from_bucket = list(
+        set(requested_document_ids) - set(current_bucket_file_stems)
```
mpjuhasz, line 188: So do I understand right based on the fixme, that we're not suffixing the requested document ids here with `_translated_en` to see if they're among the stems, because that comes in a later PR? Or is it that the `requested_document_ids` already contain the suffix?

mpjuhasz (approved):
Minor clarifying comment -- if I got that right, all looks good to me 👌

flows/inference.py
THOR300: Yeah correct, we've identified that we need to check for translated docs here but I wanted this Pr to be solely a refactor so we update the functionality in one of the stacked Pr's.

flows/inference.py
```diff
@@ -104,18 +104,23 @@ def get_bucket_paginator(config: Config, prefix: str):
     )
 
 
-def list_bucket_doc_ids(config: Config) -> list[str]:
-    """Scan configured bucket and return all IDs."""
+def list_bucket_file_stems(config: Config) -> list[str]:
```
jesse-c, line 107: **polish:** A type alias would be nice, so it's reason to understand all the strings being passed around.

jesse-c (approved):

PR #299, opened 2025-03-11
Bugfix: Find all file stems for configured document ids

THOR300:
This Pull Request: 
---

> Is a bug fix for the following...

- When we provide document ids from the `new_and_updated_documents.json` or the prefect ui we are reading files from s3 with `${document_id}.json`. This means we miss translated files, e.g. that could be the french document and we miss a `${document_id}_translated_en.json`.
- Thus, we need to update the interface to inference to pick up translated files as well. 

## Testing

- Deployed to staging and kicked off a [run](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/49a7b4c1-9654-4dd0-b3c4-a25a51f53796?entity_id=0e4cc07f-6269-4c9d-9b01-c44c2eb68896).
- As we can see we even though we only specify one document id we run on two documents as there is a non-translated and translated document in s3.
- We then see the `ValueError` raised as we don't run inference on the non-english document.


flows/utils.py
```diff
@@ -64,3 +66,41 @@ def remove_translated_suffix(file_name: str) -> str:
...
+    document_id: str, bucket_name: str, prefix: str
+) -> list[str]:
+    """
+    Get the file stems for a document ID.
+
+    This function is used to get the file stems for a document ID. For example we would
+    find any translated documents in a directory for a document id as follows:
+
+    Example:
+    "CCLW.executive.1.1" -> ["CCLW.executive.1.1_en_translated", "CCLW.executive.1.1"]
```
THOR300, line 96: translate_en

flows/utils.py
```diff
@@ -64,3 +66,41 @@ def remove_translated_suffix(file_name: str) -> str:
     E.g. "CCLW.executive.1.1_en_translated" -> "CCLW.executive.1.1"
     """
     return re.sub(r"(_translated(?:_[a-zA-Z]+)?)$", "", file_name)
+
+
+# TODO: Add test
```
THOR300, line 71: Holding off on writing tests until some of this work gets reviewed in case we go down a different path.

mpjuhasz (approved):
🚀

flows/utils.py
```diff
@@ -64,3 +66,46 @@ def remove_translated_suffix(file_name: str) -> str:
...
+    except ClientError as e:
+        if e.response["Error"]["Code"] == "404":
+            return False
+        raise
+
+
+# TODO: Add test
+def get_file_stems_for_document_id(
+    document_id: str, bucket_name: str, prefix: str
+) -> list[str]:
```
jesse-c, line 88: **polish:** A type alias would be nice, so it's reason to understand all the strings being passed around.

flows/utils.py
```diff
@@ -64,3 +66,46 @@ def remove_translated_suffix(file_name: str) -> str:
...
+        return True
+    except ClientError as e:
+        if e.response["Error"]["Code"] == "404":
+            return False
+        raise
+
+
+# TODO: Add test
+def get_file_stems_for_document_id(
+    document_id: str, bucket_name: str, prefix: str
```
jesse-c, line 87: **fix:** There's asn existing type alias to be used for `document_id`.

flows/utils.py
```diff
@@ -64,3 +66,46 @@ def remove_translated_suffix(file_name: str) -> str:
...
+    find any translated documents in a directory for a document id as follows:
+
+    Example:
+    "CCLW.executive.1.1" -> ["CCLW.executive.1.1_translated_en", "CCLW.executive.1.1"]
+    """
+    stems = [document_id]
+
+    if "translated" in document_id:
+        # If we have a translated suffix then this is not considered an import id.
+        return stems
```
jesse-c, line 102: **question:** Why would this ever happen? The document IDs are only ever provided, right? You mentioned this in the PR description:

> When we provide document ids from the new_and_updated_documents.json or the prefect ui we are reading files from s3 with ${document_id}.json. 

I think it would be good to remove this, or raise an exception, or change the behaviour. 

Where this function is called, it expects at a minimum an array of length 1, that would have a document ID (with no translation prefix)?

jesse-c (approved):

flows/utils.py
THOR300: Yeah agreed this shouldn't be needed.

PR #300, opened 2025-03-11
feat: Save empty inference results

jesse-c:
This is so downstream, we know to do removals.

TOWARDS PLA-481


THOR300 (approved):
Great to get this in!

PR #301, opened 2025-03-11
feat: Add de-indexing pipeline and concept store's CDN clean-up 

jesse-c:
This pipeline reads what's the state in S3, and then undoes that from Vespa (hence, the "de-index") and then removes the S3 objects as well.

It uses the same interface as other pipelines, except that you must specify the classifiers specs, as we don't want to ever allow a full removal of all concepts.

This approach avoids a full scan of Vespa, to remove a concept.

It also includes cleaning up the concept store's CDN and also then trigger a de-index.

# FAQ

**Resiliency on partial updates failures**

De-indexing accounts for if updating a document passage fails. When removing a concept from the document passage fails, it'll still be counted, and written to S3.

The usage here is intended to re-run the de-indexing pipeline again, for the concept(s) and family document(s), if the failure wasn't related to the pipeline itself. This could be a transient network failure. This very generously could be called idempotency.

**I don't see changes for removing concepts from concepts counts**

The concepts counts-related pipeline does a wholesale overwrite, after it's collected the concepts counts from S3. That means, a concepts counts object for concepts will be deleted in de-indexing, and thus won't be seen during the counting pipeline, and will be removed from the family document.

---

TOWARDS PLA-481
TOWARDS PLA-469


jesse-c:

- #289
    - #300
        - #301 ◀

flows/deindex.py
```diff
@@ -0,0 +1,1231 @@
+import asyncio
```
THOR300, line 1: **thought:** This module feels like we're actually just editing the `index.py` indexing pipeline to clear down existing concepts that aren't found in s3 for a classifier, would it not be easier to just edit that pipeline so the changes are clearer? 

When I saw the name `deindex.py` my mind jumped to this being a deployment to handle the case when a concept has been entirely deleted in wikibase.

flows/deindex.py
```diff
@@ -0,0 +1,1231 @@
...
+
+            logger.info(f"gathering partial updates tasks for batch {batch_num}")
+            results = await asyncio.gather(
+                *partial_update_tasks, return_exceptions=True
+            )
+            logger.info(
+                f"gathered partial {len(results)} updates tasks for batch {batch_num}"
+            )
+
+            concepts_counts = calculate_concepts_counts_from_results(results, batch)
```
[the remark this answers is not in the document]
jesse-c: Haha yup, I did consider that. It would make when changes happen even more confusing than it already is.

flows/deindex.py
jesse-c: I did have that at one point, and you suggested to make it a separate pipeline[^1]. In that PR, which at merging was completely changed, since I separated it into several PRs, I had several attempts tried where the first was a separate, new pipeline. Then I tried it together. Lastly, I separated it back out since it was a better implementation, and from your suggestion. I gave updates on these different attempts too on Slack

[^1]: https://github.com/climatepolicyradar/knowledge-graph/pull/289#discussion_r1983679352

deployments.py
```diff
@@ -107,17 +108,24 @@ def create_deployment(
 
 create_deployment(
     flow=run_partial_updates_of_concepts_for_document_passages,
-    description="Co-ordinate updating inference results for concepts in Vespa",
+    description="Co-ordinate adding/updating inference results for concepts in Vespa",
```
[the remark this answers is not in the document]
jesse-c: Indexing does add/updates and de-indexing does removes/updates. If a concept was already indexed on a family document and > family passage within it, but the inference results changed, that would count as an update

flows/deindex.py
jesse-c: If I can communicate this better, please do lmk!

THOR300: Yeah let's catch up in our chat, this comment was more referring to the clear down of a concept.json from the cdn when an entire classifier is removed in wikibase as opposed to the updates of the concepts field on all the passages in vespa

jesse-c: Sure!

FWIW the `QXXX.json` in the CDN isn't deleted in this pipeline. That happens in https://github.com/climatepolicyradar/knowledge-graph/pull/302.

tests/flows/test_deindex.py
```diff
@@ -1001,7 +1002,7 @@ async def test_run_partial_updates_of_concepts_for_document_passages(
         vespa_search_adapter=local_vespa_search_adapter,
     )
 
-    assert document_remove_1_post.concept_counts is None
+    assert document_remove_1_post.concept_counts == concept_counts__document_remove
```
THOR300, line 1005: From my understanding the document passage will have the concepts on it that were provided on line 915 of this module and thus they should be there at the end as we aren't actually updating the family index as part of the `run_partial_updates_of_concepts_for_document_passages` function?

tests/flows/conftest.py
```diff
@@ -121,22 +121,32 @@ def vespa_app(
     print("\nSetting up Vespa connection...")
     app = Vespa(mock_vespa_credentials["VESPA_INSTANCE_URL"])
 
-    subprocess.run(
```
THOR300: Reverting the changes to this file and `tests/flows/fixtures/CCLW.executive.4934.1571.json` was helpful for me during debugging the test to isolate the issue. We can add back in if required.

tests/flows/test_deindex.py
```diff
@@ -978,8 +979,8 @@ async def test_run_partial_updates_of_concepts_for_document_passages(
         vespa_search_adapter=local_vespa_search_adapter,
     )
 
-    assert passage_remove_1_pre == passage_remove_1_post
-    assert passage_remove_2_pre == passage_remove_1_post
+    assert passage_remove_1_pre != passage_remove_1_post
+    assert passage_remove_2_pre != passage_remove_1_post
```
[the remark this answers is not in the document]
jesse-c: Nope, as the ones in the fixtures on disk shouldn't be modified

tests/flows/conftest.py
jesse-c: Will do! Thanks for the explanation.

tests/flows/test_deindex.py
jesse-c: Ah, yes! I'll change it to read the updated objects from S3.

flows/boundary.py
```diff
@@ -0,0 +1,894 @@
...
+        documents_batch=documents_batch,
+        documents_batch_num=documents_batch_num,
+        cache_bucket=cache_bucket,
+        concepts_counts_prefix=concepts_counts_prefix,
+        partial_update_flow=partial_update_flow,
+    )
+
+
+@flow
+async def updates_by_s3(
```
jesse-c, line 817: **note:** The name was very indexing specific, when it was merged last week, so I've corrected it to be generic for both indexing and de-indexing.

tests/flows/fixtures/CCLW.executive.4934.1571.json
```diff
@@ -46,126 +46,7 @@
...
-                            "author": "['Argentina']",
-                            "link": "",
-                            "author_is_party": "1.0",
-                            "date": "2019-11-27 00:00:00",
-                            "world_bank_region": "Latin America & Caribbean",
-                            "prediction": "True"
-                        }
-                    }
-                ]
+                "labelled_passages": []
```
jesse-c, line 49: **note:** I've added this, and the other fixture change, back, since I'm pretty sure when they were reverted before to get the test passing, mean that the test was passing, but it wasn't actually testing correct things.

jesse-c:
**update:** Reported success in [Slack](https://climate-policy-radar.slack.com/archives/C08F40KPLNR/p1742993869172699?thread_ts=1742918874.859289&cid=C08F40KPLNR).

tests/local_vespa/local_vespa.just
```diff
@@ -33,8 +33,9 @@ vespa_feed_data:
 	vespa feed tests/local_vespa/test_documents/document_passage.json
 
 # Deletes some test data into local Vespa instance
+#
```
jesse-c, line 36: **note:** This change makes it not print it out when running the task

tests/local_vespa/local_vespa.just
```diff
@@ -23,7 +23,7 @@ vespa_healthy:
 # Deploys schema to dev instance
 vespa_deploy_schema:
 	vespa config set target local
-	@vespa deploy tests/local_vespa/test_app --wait 300
+	@vespa deploy tests/local_vespa/test_app --wait 180
```
jesse-c, line 26: **note:** 5 minutes is very long, so I've halved it.

justfile
```diff
@@ -15,11 +15,11 @@ install +OPTS="":
 
 # test the project
 test +OPTS="":
-    poetry run pytest {{OPTS}}
+    poetry run pytest --disable-pytest-warnings --color=yes {{OPTS}}
 
 # test the project, excluding tests that rely on a local vespa instance
 test-without-vespa +OPTS="":
-    poetry run pytest {{OPTS}} -m 'not vespa'
+    poetry run pytest --disable-pytest-warnings --color=yes {{OPTS}} -m 'not vespa'
```
jesse-c, line 22: **note:** It's otherwise so noisy, but happy to revert this. The `color` is to try and make it look nicer in GitHub workflows.

jesse-c: **update:** Much nicer!

CleanShot 2025-03-27 at 16 33 27

flows/boundary.py
```diff
@@ -0,0 +1,1416 @@
...
+        raise ValueError("there was at least 1 task that failed")
+
+
+# Index -------------------------------------------------------------------------
+
+
+# FIXME: From what I can tell this function should be identical to the related function
+# deindex.py. In deindex.py we need to remove_translated_suffix from the
+# document_import_id and in this function I'm assuming we should add the updates to how
+# we calculate concepts counts from results.
```
jesse-c, line 928: **question:** @THOR300, I don't think I understand this, do you mind expanding on it?

tests/flows/test_deindex.py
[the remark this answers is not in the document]
olaughter: Worth ticketing the fixture rethink?

olaughter (approved):
🙌

flows/boundary.py
THOR300: @jesse-c yes this was a note I put to myself during one of the refactor PR's. 

The indexing pipeline was using an old code path for storing concept counts in s3 whereas the new deindex pipeline was using an updated methodology. At a glance I was under the impression that the indexing pipeline could potentially make use of the same updated code (that assumption may have been wrong)

jesse-c: I see! From looking at where it's done for [indexing](https://github.com/climatepolicyradar/knowledge-graph/blob/add-deindex-pipeline/flows/boundary.py#L1018-L1039) and for [de-indexing](https://github.com/climatepolicyradar/knowledge-graph/blob/add-deindex-pipeline/flows/boundary.py#L1340-L1353), there may be a chance for alignment.

tests/flows/test_deindex.py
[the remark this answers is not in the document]
jesse-c: I'll add a todo for myself to try and coherently write-up some thoughts, to double-check I'm not just annoyed for no good reason haha

flows/deindex.py
```diff
@@ -0,0 +1,449 @@
...
+        raise ValueError(
+            f"Failed to process document batch in clean-ups task batch. Batches with failures: #{failed_batches_str}"
+        )
+
+
+@flow(
+    on_failure=[SlackNotify.message],
+    on_crashed=[SlackNotify.message],
+)
+async def deindex_labelled_passages_from_s3_to_vespa(
```
THOR300, line 353: @jesse-c just to confirm my understanding of this, so we do the following: 

- remove all the old labelled passages and concept counts for a classifier in s3 
- run de-indexing into vespa clearing down the concept on the labelled passages that were tagged by the current primary version / latest 

And thus, post run we should have only the "primary" or latest labelled passages in s3?

jesse-c: Depending on the classifier specs. parameter (e.g. `Q770:latest`, yes. If it's `latest` for the alias (AKA version), then it will do a big wipe. This would just be triggered off of a concept being removed from the concept store.

Otherwise, if it's as specific alias (e.g. `Q770:v4`), it will remove the artifacts in S3 and the spans from Vespa. We don't know what version is in Vespa, so we can't remove a span _iff_ it's not the latest. The indexing pipeline though does overwrites, so getting the primary (AKA latest) in is expected to be done through there.

De-indexing really is just getting data out, and not getting it up-to-date with latest inference results or classifier specs.

PR #302, opened 2025-03-11
feat: Clean-up and de-index concepts

jesse-c:
We now wipe the concepts from S3 (the CDN).

Along with this, we trigger the new de-indexing pipeline, to remove the concepts' inference and concepts counts objects, and labelled passages (spans) in Vespa.

TOWARDS PLA-481
TOWARDS PLA-469

PR #303, opened 2025-03-11
refactor: Share code between flows for systems boundaries

jesse-c:
I found it easy to think about this as the boundaries, and where we traverse them. This isn't the final step in sharing, but is a good start imo.

TOWARDS PLA-481


jesse-c:

- #289
    - #300
        - #301
            - #302
                - #303 ◀

PR #304, opened 2025-03-11
Refactoring document id to file stem in the indexing pipeline

THOR300:
This Pull Request: 
---
- Is a refactor to update the naming of document id in the **_INDEXING_** pipeline to being file stem where applicable. 
- This is as we are referring to objects as documents ids when they are not which has lead to processing [issues](https://linear.app/climate-policy-radar/issue/POL-2385/classifiers-not-working-on-translated-documents#comment-a3d1103a). 
- Why use file stems? This is the attribute that is used in the `Path` object, e.g.:

```python
>>> Path("s3://..../file.json").stem
>>> "file"
```


mpjuhasz (approved):
LGTM 💯

jesse-c (approved):

PR #307, opened 2025-03-18
Adding a script for cleaning the labelled passages directory.

THOR300:
This Pull Request: 
---
- Adds a script for cleaning up the labelled passages directory. 
- This is required as we have outputs that are from non-english languages (e.g. we have run inference on french docs for example).

### DIFF
--- a/scripts/clean_labelled_passages_directory.py
+++ b/scripts/clean_labelled_passages_directory.py
@@ -4,10 +4,10 @@
 # The script will:
 # 1. Find all the non-english language documents from the embeddings input prefix.
 # 2. Delete the non-english language documents from the labelled passages prefix;
-#   using the document id that corresponds to the file stem.
+#   using the corresponding file stems.
 #
 # E.g. embeddings_input/CCLW.exec.1.1["languages"] != ["en"]
-#   -> delete labelled_passages/CCLW.exec.1.1.json
+#   -> delete labelled_passages/Q123/v1/CCLW.exec.1.1.json
 
 import json
 from pathlib import Path
--- a/scripts/clean_labelled_passages_directory.py
+++ b/scripts/clean_labelled_passages_directory.py
@@ -41,12 +41,7 @@ def json_generator(
 ) -> Generator[tuple[Dict[str, Any], str], None, None]:
     """Generator that yields JSON objects from S3 bucket."""
     for key in list_s3_objects(bucket_name, prefix):
-        try:
-            yield load_json_from_s3(bucket_name, key), key
-        except json.JSONDecodeError:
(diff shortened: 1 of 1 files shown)


scripts/clean_labelled_passages_directory.py
```diff
@@ -0,0 +1,105 @@
...
+    bucket_name: str, prefix: str
+) -> Generator[tuple[Dict[str, Any], str], None, None]:
+    """Generator that yields JSON objects from S3 bucket."""
+    for key in list_s3_objects(bucket_name, prefix):
+        try:
+            yield load_json_from_s3(bucket_name, key), key
+        except json.JSONDecodeError:
+            print(f"Skipping {key}: Invalid JSON format")
+        except Exception as e:
+            print(f"Skipping {key}: {e}")
```
mpjuhasz, line 49: Are we expecting these errors as normal behaviour? I'm just wondering, if we're catching all of them, are we not running the risk of leaving non-English documents in the prefix unnoticed. Wouldn't it be better to throw these errors?

scripts/clean_labelled_passages_directory.py
```diff
@@ -0,0 +1,105 @@
...
+
+def extract_non_english_language_documents(
+    bucket_name: str, prefix: str
+) -> list[tuple[Dict[str, Any], str]]:
+    """Extract the 'languages' key from JSON objects in S3 bucket."""
+    typer.echo(f"Extracting non-English language documents from {bucket_name}/{prefix}")
+    non_english_documents = []
+    for json_obj, key in json_generator(bucket_name, prefix):
+        languages = json_obj.get("languages", [])
+        if languages and languages != ["en"]:
```
mpjuhasz, line 60: Which are the documents with an empty / null `languages` field?

mpjuhasz (approved):
Looks good, two minor comments -- with the first one more important than the other around how we're throwing errors.

scripts/clean_labelled_passages_directory.py
THOR300: Agreed, makes total sense!

scripts/clean_labelled_passages_directory.py
THOR300: So in the model of a `ParserOuptut` the [languages field ](https://github.com/climatepolicyradar/cpr-sdk/blob/main/src/cpr_sdk/parser_models.py#L188)can be nullable. 

Upon catch up over huddle with @mpjuhasz we've agreed that we only want to run indexing / inference on documents that explicitly have language english. This solution relies on the creation of a translated document with the language of english for the documents with a nullable language. The code has been checked and this is supported.

PR #313, opened 2025-03-20
Find correct translated document paths 

THOR300:
This Pull Request: 
---
- Is a bug fix for the inference and indexing pipeline where we're not picking up translated files. 
- This is as the labelled passages are at paths like:
   `labelled_passages/Q787/v4/CCLW.executive.1813.2418_translated_en.json`
  and we're looking under the following prefix in staging / production:
    `labelled_passages/CCLW.executive.1813.2418_translated_en.json`

Ran in staging on a document that previously had no results for the translated document: 
- https://app.dev.climatepolicyradar.org/documents/cipe-resolution-december-19-2002-no-123-2002-review-of-the-guidelines-for-national-policies-and-measures-to-reduce-greenhouse-gas-emissions-law-n-120-2002_a151?cfn=air%20pollution%20risk
image

### DIFF
--- a/flows/index.py
+++ b/flows/index.py
@@ -802,6 +802,10 @@ def s3_paths_or_s3_prefixes(
                 cache_bucket=cache_bucket,
                 prefix=prefix,
             )
+
+            logger.info(
+                f"Identifier {len(document_paths)} documents to process from {len(document_ids)} document IDs"
+            )
             return S3Accessor(paths=document_paths, prefixes=None)
 
         case (None, list()):
--- a/flows/index.py
+++ b/flows/index.py
@@ -28,7 +28,7 @@ from vespa.io import VespaQueryResponse, VespaResponse
 from flows.inference import DOCUMENT_TARGET_PREFIX_DEFAULT
 from flows.utils import (
     SlackNotify,
-    get_all_document_paths_from_document_ids,
+    get_all_labelled_passage_paths_from_document_ids,
     remove_translated_suffix,
 )
 from scripts.cloud import (
@@ -812,7 +812,7 @@ def s3_paths_or_s3_prefixes(
             # Run on specified documents, for the specified classifier
             logger.info("run on specified documents, for the specified classifier")
 
-            document_paths = get_all_document_paths_from_document_ids(
+            document_paths = get_all_labelled_passage_paths_from_document_ids(
                 document_ids=document_ids,
                 classifier_specs=classifier_specs,
                 cache_bucket=cache_bucket,
--- a/flows/index.py
+++ b/flows/index.py
@@ -821,7 +821,7 @@ def s3_paths_or_s3_prefixes(
             )
 
             logger.info(
-                f"Identifier {len(document_paths)} documents to process from {len(document_ids)} document IDs"
+                f"Identified {len(document_paths)} documents to process from {len(document_ids)} document IDs"
             )
             return S3Accessor(paths=document_paths, prefixes=None)
 
--- a/flows/index.py
+++ b/flows/index.py
@@ -28,7 +28,7 @@ from vespa.io import VespaQueryResponse, VespaResponse
 from flows.inference import DOCUMENT_TARGET_PREFIX_DEFAULT
 from flows.utils import (
     SlackNotify,
-    get_all_labelled_passage_paths_from_document_ids,
+    add_labelled_passage_paths_from_translated_files_if_existent,
     remove_translated_suffix,
 )
 from scripts.cloud import (
@@ -813,11 +813,13 @@ def s3_paths_or_s3_prefixes(
             # Run on specified documents, for the specified classifier
             logger.info("run on specified documents, for the specified classifier")
 
-            document_paths = get_all_labelled_passage_paths_from_document_ids(
-                document_ids=document_ids,
-                classifier_specs=classifier_specs,
-                cache_bucket=cache_bucket,
-                labelled_passages_prefix=prefix,
+            document_paths = (
+                add_labelled_passage_paths_from_translated_files_if_existent(
+                    document_ids=document_ids,
+                    classifier_specs=classifier_specs,
+                    cache_bucket=cache_bucket,
+                    labelled_passages_prefix=prefix,
+                )
             )
 
             logger.info(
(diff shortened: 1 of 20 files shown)


flows/utils.py
```diff
@@ -98,13 +103,72 @@ def get_file_stems_for_document_id(
     Example:
     "CCLW.executive.1.1" -> ["CCLW.executive.1.1_translated_en", "CCLW.executive.1.1"]
     """
+    logger = get_run_logger()
+    logger.info(f"Checking for translated files for document ID: {document_id}")
```
THOR300, line 107: Remove merge opening PR.

flows/utils.py
```diff
@@ -87,7 +91,7 @@ def s3_file_exists(bucket_name: str, file_key: str) -> bool:
 
 
 def get_file_stems_for_document_id(
-    document_id: DocumentImportId, bucket_name: str, prefix: str
+    document_id: DocumentImportId, bucket_name: str, document_key: str
```
THOR300, line 94: Use DocumentKey

flows/utils.py
```diff
@@ -1,15 +1,19 @@
...
 from botocore.exceptions import ClientError
 from prefect.settings import PREFECT_UI_URL
 from prefect_slack.credentials import SlackWebhook
 
+from scripts.cloud import ClassifierSpec
+
 # Example: CCLW.executive.1813.2418
 DocumentImportId: TypeAlias = str
 DocumentStem: TypeAlias = str
+DocumentKey: TypeAlias = str
```
THOR300, line 16: Can we import from anywhere?

tests/flows/test_utils.py
```diff
@@ -98,9 +100,55 @@ def test_get_file_stems_for_document_id(test_config, mock_bucket_documents) -> N
...
+                file_name,
+            ),
+            Body=body,
+            ContentType="application/json",
+        )
+
+    # Get all the document paths for classifiers and documents ids.
+    document_paths = get_all_labelled_passage_paths_from_document_ids(
+        document_ids=["CCLW.executive.1.1", "CCLW.executive.10083.rtl_190"],
+        classifier_specs=[classifier_spec],
```
THOR300, line 143: Add another classifier here?

flows/utils.py
```diff
@@ -101,10 +105,65 @@ def get_file_stems_for_document_id(
...
-            stems.append(stem)
+            file_key=translated_file_key.__str__(),
+        )
+        if file_exists:
+            stems.append(translated_file_key.stem)
+
     return stems
+
+
+def get_all_labelled_passage_paths_from_document_ids(
```
THOR300, line 123: Rename function.

flows/utils.py
```diff
@@ -101,10 +104,65 @@ def get_file_stems_for_document_id(
...
-            stems.append(stem)
+            file_key=translated_file_key.__str__(),
+        )
+        if file_exists:
+            stems.append(translated_file_key.stem)
+
     return stems
+
+
+def add_labelled_passage_paths_from_translated_files_if_existent(
```
olaughter, line 122: **refactor:** This function name is very verbose and yet also misleading. "add_" makes it sound like something is being written, rather then retrieved. I also think the `_if_existent` suffix is an implementation detail that doesn't make sense to add here

olaughter: Am I reading it right that it actually does both translated and non translated too, because it calls `get_file_stems_for_document_id`? I just went searching for the non translation paths version of this function based on the function name 🥲

flows/inference.py
```diff
@@ -195,8 +195,9 @@ def determine_file_stems(
 
     requested_document_stems = []
     for doc_id in requested_document_ids:
+        document_key = os.path.join(config.document_source_prefix, f"{doc_id}.json")
         requested_document_stems += get_file_stems_for_document_id(
-            doc_id, config.cache_bucket, config.document_source_prefix
+            doc_id, config.cache_bucket, document_key
```
olaughter, line 200: **question:** Is the translated paths functionality only needed for indexing?

flows/index.py
```diff
@@ -813,24 +813,18 @@ def s3_paths_or_s3_prefixes(
...
-                    classifier_spec.alias,
-                    f"{file_stem}.json",
-                )
-                for classifier_spec in classifier_specs
-                for file_stem in file_stems
-            ]
+            logger.info(
+                f"Identified {len(document_paths)} documents to process from {len(document_ids)} document IDs"
+            )
             return S3Accessor(paths=document_paths, prefixes=None)
```
olaughter, line 828: **praise:** good log message

olaughter: **suggestion:** It's only used on this case statement, do you think it would be worth adding this to the outcome of the other cases as well? Either as a drive by or its own pr?

olaughter (approved):
lgtm, some points to resolve below but I've approved so you are unblocked here

flows/utils.py
THOR300: Yeah it's not a great function name! 

How does this sound: `get_labelled_passage_paths_with_translated`?

THOR300: Yeah essentially it's taking the document ids, checking if we have any labelled passages for a translated version of the document and then returning all the paths. So yes it's doing non/translated.

flows/inference.py
THOR300: Believe so: 

inference runs on: 

```
    current_bucket_file_stems = list_bucket_file_stems(config=config)
    validated_file_stems = determine_file_stems(
        config=config,
        use_new_and_updated=use_new_and_updated,
        requested_document_ids=document_ids,
        current_bucket_file_stems=current_bucket_file_stems,
    )
```

indexing runs on: 
```
    s3_accessor = s3_paths_or_s3_prefixes(
        classifier_specs,
        document_ids,
        config.cache_bucket,
        config.document_source_prefix,
    )
```

flows/index.py
THOR300: Ooh good thought, ;looking at the other cases they use prefixes that are then passed to generators so I'm not sure how much we would get from logging out "we got 20 prefixes" type thing.

THOR300: Worth noting as well I've checked over those cases and I'm confident that we will catch translated documents as they list out all the file keys under a prefix.

PR #315, opened 2025-03-20
Argilla to HF backup permission fix

mpjuhasz:
## What this PR does
- makes all backed up HF datasets private
- creates a new argilla users `prefect-owner-bot` with credentials in bitwarden
- creates a new secret in SSM, `/Argilla/Owner/APIKey`, which is the above user's key
- makes the backup script use this key with `owner` permissions to be able to list users


kdutia (approved):
lgtm, but i think the PR description might be out of date (we're not refactoring into `ArgillaSession` here, from what i can see!)

PR #319, opened 2025-03-25
build: Reflect latest formatting

jesse-c:
I'm temporarily reverted the Pyright version change, since it added a bunch of (reasonable) errors.

FIXES PLA-486

### DIFF
--- a/scripts/cloud.py
+++ b/scripts/cloud.py
@@ -24,8 +24,7 @@ class ClassifierSpec(BaseModel):
     )
     alias: str = Field(
         description=(
-            "The alias tag for the version to use for inference. "
-            "e.g 'latest' or 'v2'"
+            "The alias tag for the version to use for inference. e.g 'latest' or 'v2'"
         ),
         default="latest",
         min_length=1,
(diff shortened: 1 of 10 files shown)


kdutia (approved):
🎨

scripts/cloud.py
```diff
@@ -159,7 +158,7 @@ def is_logged_in(aws_env: AwsEnv, use_aws_profiles: bool) -> bool:
         aws_env_as_profile = aws_env if use_aws_profiles else None
 
         sts = get_sts_client(aws_env_as_profile)
-        sts.get_caller_identity()
+        sts.get_caller_identity()  # pyright: ignore[reportAttributeAccessIssue]
```
THOR300, line 161: thought: There is a fix for these type hints if we want. It requires adding this package as a dependency `mypy-boto3-sts` and using the `STSClient` stub as a return type for the get client function. 

```python
from mypy_boto3_sts.client import STSClient

def get_sts_client(
    aws_env: Optional[AwsEnv],
) -> STSClient:
```

THOR300 (approved):
Looks great just one non-blocking comment.

PR #323, opened 2025-04-01
fix: Sort numerically, and not lexicographically

jesse-c:
I noticed this, and thought I'd quickly fix this, so that it's easier to scan. It's a big diff, but we'll appreciate it long-term.

I've created a `scripts` sub-foler too, as we have a `flows` sub-folder. Along with this, I've improved/fixed some type hints.

TOWARDS PLA-481


scripts/update_classifier_spec.py
```diff
@@ -86,15 +85,23 @@ def get_relevant_model_version(
...
     """Check to see if this model already has a version found for the env"""
-    env_models = [m.split(":")[0] for m in classifier_specs]
+    env_models = [m.name for m in classifier_specs]
     return model_name not in env_models
 
 
+def sort_specs(specs: list[ClassifierSpec]) -> list[ClassifierSpec]:
+    # Have stable ordering. First, the name is gotten from
+    # `Q000:v0`, and then the number part is gotten from `Q000`.
+    return sorted(specs, key=lambda x: int(x.alias[1:]))
```
harrisonpim, line 99: could we achieve the same effect for `WikibaseID`s everywhere by adding a couple of extra dunder methods to the class? 

eg cursor suggests

```py

class WikibaseID(str):
    """A Wikibase ID, which is a string that starts with a 'Q' followed by a number."""

    regex = r"^Q[1-9]\d*$"

    @property
    def numeric_value(self) -> int:
        """Get the numeric portion of the Wikibase ID"""
        return int(self[1:])

    def __lt__(self, other) -> bool:
        if not isinstance(other, WikibaseID):
            return NotImplemented
        return self.numeric_value < other.numeric_value

    def __gt__(self, other) -> bool:
        if not isinstance(other, WikibaseID):
            return NotImplemented
        return self.numeric_value > other.numeric_value

    def __le__(self, other) -> bool:
        if not isinstance(other, WikibaseID):
            return NotImplemented
        return self.numeric_value <= other.numeric_value

    def __ge__(self, other) -> bool:
        if not isinstance(other, WikibaseID):
            return NotImplemented
        return self.numeric_value >= other.numeric_value

    @classmethod
    def _validate(cls, value: str, field=None) -> str:
        """Validate that the Wikibase ID is in the correct format"""
        if not re.match(cls.regex, value):
            raise ValueError(f"{value} is not a valid Wikibase ID")
        return value

    @classmethod
    def __get_validators__(cls):
        """Return a generator of validators for the WikibaseID class"""
        yield cls._validate

    def __new__(cls, value: str) -> "WikibaseID":
        """Create a new instance of WikibaseID after validation"""
        validated_value = cls._validate(value)
        return str.__new__(cls, validated_value)
```

jesse-c: We could, and I'd also like to add a new type for `Alias` as well to get comparable benefits. I started on both but ran out of time, for this PR!

tests/scripts/test_update_classifier_spec.py
```diff
@@ -147,3 +151,20 @@ def test_parse_spec_file_invalid_format(invalid_contents, tmp_path):
...
             parse_spec_file(test_env)
+
+
+def test_sort_specs():
+    unsorted_specs = [
+        ClassifierSpec(name="Q123", alias="Q456"),
+        ClassifierSpec(name="Q789", alias="Q123"),
+        ClassifierSpec(name="Q456", alias="Q789"),
+        ClassifierSpec(name="Q999", alias="Q10"),
+        ClassifierSpec(name="Q111", alias="Q1000"),
```
harrisonpim, line 162: we expect these aliases to be something like `v456` rather than `Q456`, right?

tests/scripts/test_update_classifier_spec.py
```diff
@@ -147,3 +151,20 @@ def test_parse_spec_file_invalid_format(invalid_contents, tmp_path):
...
+    ]
+
+    sorted_specs = sort_specs(unsorted_specs)
+
+    expected_order = ["Q10", "Q123", "Q456", "Q789", "Q1000"]
+    actual_order = [spec.alias for spec in sorted_specs]
```
harrisonpim, line 168: only just realised that you're sorting by the highest _version_, rather than the QID... is that the most meaningful order?

tests/scripts/test_update_classifier_spec.py
jesse-c: **done:**

tests/scripts/test_update_classifier_spec.py
jesse-c: **done:** Yes, this was me being distracted!

scripts/update_classifier_spec.py
jesse-c: Nice work! 🎉

flows/classifier_specs/staging.yaml
```diff
@@ -74,3 +46,31 @@
 - Q973:v5
 - Q983:v5
 - Q986:v6
+- Q1016:v0
```
THOR300, line 49: question: Do we need to do sandbox.yaml and labs.yaml as well?

THOR300 (approved):
Looks good, approving subject to tests passing.

flows/classifier_specs/staging.yaml
jesse-c: There were no changes to commit for those environments

PR #324, opened 2025-04-01
make wikibase ids sort numerically, rather than lexicographically

harrisonpim:
following a conversation with @jesse-c in #323, i thought i'd add this little feature to our `WikibaseID` class anyway!

also added a small test which fails on the previous version, and now passes


kdutia (approved):
whoah, `total_ordering` is clever! ✨

PR #325, opened 2025-04-01
fix: Check for Prefect state failures

jesse-c:
In https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/fd984a38-16fe-48bf-b246-237c39c4c632, the `updates-by-s3` pipeline passed, but if you look in it, there was a failed task, so overall it should've been reported as failed.

The approach is the same used elsewhere[^1][^2].

TOWARDS PLA-481

[^1]: https://github.com/climatepolicyradar/knowledge-graph/blob/fix%2Freport-flow-state-type-failures/flows/deindex.py#L318-L340
[^2]: 
CleanShot 2025-04-01 at 14 37 17


THOR300 (approved):
LGTM

PR #327, opened 2025-04-01
fix: Deserialise per item

jesse-c:
Nothing notable to say, really. This is the way the data is serialised at the present[^2]. I got a ton of failures in Prefect[^1].

The fixtures are fresh from S3 from one of the flow run failures.

Ideally, we're not saving them in 2 different ways, especially one that's broken. This PR at least gives as a nice migration when fixing the other way.

TOWARDS PLA-481

[^1]: https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/1210c142-de95-4ae2-9201-b531244a8b69?entity_id=5cae68f4-8db7-4142-a34a-4a67ea7bd63b
[^2]: I _think_ the culprit is https://github.com/climatepolicyradar/knowledge-graph/blob/fix%2Fload-serialised-lists-of-strings/flows/inference.py#L300-L304, where `model_dump_json` isn't used.


flows/boundary.py
```diff
@@ -305,6 +305,21 @@ def load_labelled_passages_by_uri(
...
     """Load and transforms the S3 object's body into LabelledPassages objects."""
     object_json = json.loads(_s3_object_read_text(s3_path=document_object_uri))
+    if len(object_json) == 0:
+        return []
+
+    # Currently we _sometimes_ serialise the labelled passages as a
+    # JSON list, but the items of the list are a funny serialisation
+    # of them, so they're still raw strings, when loaded in.
+    if isinstance(object_json[0], str):
+        object_json = list(
```
THOR300, line 315: **suggestion**: `object_json = [json.loads(labelled_passage) for labelled_passage in object_json]`

THOR300 (approved):
LGTM

flows/boundary.py
jesse-c: **done:**

PR #328, opened 2025-04-02
fix: Surface tasks failures in inference

jesse-c:
I did an inference run[^1] and everything succeeded! jk, none that I sampled did. Similar to changes yesterday, this change surfaces failures for tasks.

TOWARDS PLA-481

[^1]: https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/9c17ae1c-a9b6-4fa6-b1c7-c6d7e6a61994?entity_id=f3afae10-f141-44ae-b969-ad695df5b74c

### DIFF
--- a/flows/inference.py
+++ b/flows/inference.py
@@ -1,11 +1,13 @@
 import asyncio
 import json
 import os
+from collections import defaultdict
 from collections.abc import Generator
 from dataclasses import dataclass
 from io import BytesIO
 from pathlib import Path
 from typing import Final, Optional, TypeAlias
+from uuid import UUID
 
 import boto3
 import prefect.artifacts as artifacts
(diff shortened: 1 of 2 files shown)


flows/inference.py
```diff
@@ -399,15 +402,15 @@ async def run_classifier_inference_on_document(
     classifier_name: str,
     classifier_alias: str,
     classifier: Classifier,
-) -> DocumentRunIdentifier:
+) -> None:
```
jesse-c, line 405: **note:** The return value was being used for nothing, really, and `None` is easier to check for, 'cause typing doesn't allow you to use `isinstance` and the like on `TypeAlias` or `NewType`.

olaughter (approved):

PR #329, opened 2025-04-02
Add a WikidataSession class

harrisonpim:
This PR adds a sister class to our `WikibaseSession`, which fetches concepts from wikidata instead. 

This was motivated by some experimental work that @kdutia and I are doing on geographical concepts and classifiers. 

As an example, this class allows you to run eg

```py
from src.wikidata import WikidataSession
from src.classifier import KeywordClassifier
from rich import print

wikidata = WikidataSession()

entity_id_for_country = "Q6256"
concept_of_country = wikidata.get_concept(entity_id_for_country)
instances_of_country = wikidata.get_instances_of(entity_id_for_country)
print(f"found {len(instances_of_country)} instances of {concept_of_country.preferred_label}")

concept_of_a_specific_country = wikidata.get_concept(instances_of_country[0])
print(f"Pulled the concept for {concept_of_a_specific_country.preferred_label} from wikidata")

classifier = KeywordClassifier(concept_of_a_specific_country)
sentence_about_our_country = f"This sentence is about {concept_of_a_specific_country.preferred_label}, aka {concept_of_a_specific_country.alternative_labels[0]}"
print(classifier.predict(sentence_about_our_country))
```

```
found 196 instances of country
Pulled the concept for Canada from wikidata
[
    Span(
        text='This sentence is about Canada, aka CA',
        start_index=23,
        end_index=29,
        concept_id='Q16',
        labellers=['KeywordClassifier("Canada")'],
        timestamps=[datetime.datetime(2025, 4, 2, 11, 7, 12, 872431)],
        id='cg38vj6e',
        labelled_text='Canada'
    ),
    Span(
        text='This sentence is about Canada, aka CA',
        start_index=35,
        end_index=37,
        concept_id='Q16',
        labellers=['KeywordClassifier("Canada")'],
        timestamps=[datetime.datetime(2025, 4, 2, 11, 7, 12, 872461)],
        id='xguy8742',
        labelled_text='CA'
    )
]
```

We probably won't end up using KeywordClassifiers to identify geography mentions (we might not even use this `WikidataSession` class), but thought it might be worth sharing anyway!

N.B. If we _do_ end up using this class more widely, I reckon we should add a `source` attribute to our `Concept` objects which differs depending on where the data was pulled from.


src/wikidata.py
```diff
@@ -0,0 +1,224 @@
...
+            """
+        else:
+            query = f"""
+            SELECT ?item ?itemLabel WHERE {{
+              ?item wdt:{property_id} wd:{entity_id}.
+              SERVICE wikibase:label {{ bd:serviceParam wikibase:language "en". }}
+            }}
+            LIMIT {current_page_size}
+            OFFSET {offset}
+            """
```
kdutia, line 85: **polish:** for readability, it might just be worth parametrising the `s p o` bit here and then assembling the query after the if/else?

src/wikidata.py
```diff
@@ -0,0 +1,224 @@
...
+        return self.get_property_values(
+            self.INSTANCE_OF_PROPERTY, entity_id, limit=limit
+        )
+
+    def get_parent_entities(
+        self, entity_id: WikibaseID, limit: Optional[int] = None
+    ) -> list[WikibaseID]:
+        """
+        Get all parent entities of a given Wikidata entity.
+
```
kdutia, line 160: it might be worth specifying that this is using the 'instance of' property here. When i initially read this i thought it could be related to 'subclass of' (which is probably my misunderstanding of ontologies/hierarchies, but also then likely others'!)

kdutia (approved):
looks great! two nonblocking comments

src/wikidata.py
harrisonpim: good point!

src/wikidata.py
harrisonpim: yeah, i really struggled with the naming of this method! 

i usually love giving painfully long, unambiguous names to methods/variables/tests, but `get_items_of_which_the_specified_item_is_said_to_be_an_instance()` felt like a stretch, even for me 😅 

docstring definitely has room for improvement though 🙂

harrisonpim: renamed to `get_instance_of_values()`. even if it's a bit harder to understand what the method does at a glance, i think it's also harder to _misunderstand_ at a glance, hopefully nudging users to read the docstring 🤞

THOR300 (approved):
No qualms from a platform perspective as there's no usage of this in the production code base yet (considering there's no tests) 👍

PR #330, opened 2025-04-02
build: Separate CI/CD

jesse-c:
I've found it slightly distasteful to have checks permanently skipped for PRs, as it makes it slightly less clear on PRs when looking at checks, and I think it's neater to explicitly have the dependency as "iff CI passes" checked as few times as possible, and at the highest level possible.

It's now possible to have cross-workflow dependencies[^1]!

[^1]: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,60 +1,17 @@
-name: CI/CD
+name: CD
+
 on:
   workflow_dispatch:
-  pull_request:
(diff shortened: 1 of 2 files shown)


THOR300 (approved):
A nice improvement!

PR #332, opened 2025-04-02
Removing legacy argilla v1

mpjuhasz:
## What this PR does
After sucessfully migrating to Argilla v2, and not seeing any issues the past weeks, removing all the legacy argilla code from the repo now.


jesse-c:
@mpjuhasz: I see that the [dependency itself](https://github.com/climatepolicyradar/knowledge-graph/blob/main/pyproject.toml#L39) wasn't removed. Can I do that?

PR #333, opened 2025-04-02
refactor: Don't allow missing documents for inference

jesse-c:
I found it odd that we'd allow this. If, at this level, of the pipeline we're trying to run inference we'd surely be expecting the document to exist.

Now we do!

TOWARDS PLA-481


THOR300 (approved):
Sounds good, checked and we have return_exceptions=True for the async running of the tasks so no concern with causing a parent process to fail 👍

PR #334, opened 2025-04-02
fix: Use Vespa-friendly family document ID

jesse-c:
We don't store the translated prefeix for the family document ID. Thus, strip it out.

I came across this when investigating a de-index pipeline run[^1].

Some time, we should add tests for the translation scenario.

[^1]: https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/226ea582-2bcc-465a-99be-e6a85005e1dc?entity_id=a2ec38ff-35eb-4fcf-9443-0cc3c3fc805d


THOR300 (approved):
LGTM!

PR #335, opened 2025-04-02
fix: Surface more errors

jesse-c:
As per normal!

TOWARDS PLA-481

### DIFF
--- a/flows/boundary.py
+++ b/flows/boundary.py
@@ -1126,16 +1126,18 @@ async def run_partial_updates_of_concepts_for_document_passages__remove(
         for labelled_passage in document_labelled_passages
     }
 
+    logger.info(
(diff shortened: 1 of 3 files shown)


flows/boundary.py
```diff
@@ -1126,16 +1126,18 @@ async def run_partial_updates_of_concepts_for_document_passages__remove(
         for labelled_passage in document_labelled_passages
     }
 
-    with cm:
-        logger.info(
-            f"starting partial updates for {len(grouped_concepts)} grouped concepts"
-        )
+    logger.info(
```
jesse-c, line 1129: **note:** Moving as much as possible out of the `ContextManager`

THOR300 (approved):
LGTM

PR #337, opened 2025-04-03
fix: Properly timeout on Prefect

jesse-c:
We fundamentally misunderstood what the timeout parameter was for.

> By default, this function blocks until the flow run finishes executing. Specify a timeout (in seconds) to wait for the flow run to execute before returning flow run metadata. To return immediately,
without waiting for the flow run to execute, set timeout=0.[^1]

It’s a timeout for returning metadata, not for execution to finish. E.g. by having 3600, we’re saying “no matter if the `FlowRun` is still executing, return the metadata, and let the `FlowRun` continue on until it finishes, with no actual timeout”.

For an incident[^2], we were confused as to why they hadn't timed out! We actually had been confused before too, and we did set a timeout, or increased it, I don't remember which. We didn't realise this then due to our general sporadic work across many projects.

Now, we properly wait![^3]

An alternative approach is to annotate the decorator:

```python
from prefect import flow
import time

@flow(timeout_seconds=1, log_prints=True)
def show_timeouts():
    print("I will execute")
    time.sleep(5)
    print("I will not execute")
```

The change I'm proposing felt more suitable with the current code.

What if a task actually goes over? I'm not sure. I already know that this timeout will be too short for really big documents.[^4]

[^1]: https://docs-2.prefect.io/latest/api-ref/prefect/deployments/deployments/#prefect.deployments.deployments.run_deployment
[^2]: https://climate-policy-radar.slack.com/archives/C08LN4VA2D8/p1743668576959819
[^3]: https://docs-2.prefect.io/latest/api-ref/prefect/flow_runs/#prefect.flow_runs.wait_for_flow_run
[^4]: If you look at
https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/3a6dbe77-bebb-4788-89d2-f581ce2e948e?entity_id=5bedd38b-8432-4f1b-b792-3a71bbcb4eb7,
you'll see most are relatively quick, but a few are >3, or >6 hours.

### DIFF
--- a/flows/boundary.py
+++ b/flows/boundary.py
@@ -25,6 +25,7 @@ from cpr_sdk.ssm import get_aws_ssm_param
 from prefect import flow, get_run_logger
 from prefect.client.schemas.objects import FlowRun, StateType
 from prefect.deployments.deployments import run_deployment
+from prefect.flow_runs import wait_for_flow_run
 from prefect.logging import get_logger
 from pydantic import BaseModel
 from vespa.io import VespaQueryResponse, VespaResponse
@@ -815,19 +816,14 @@ async def run_partial_updates_of_concepts_for_batch_flow_or_deployment(
     aws_env: AwsEnv,
     as_deployment: bool,
     partial_update_flow: Operation,
+    ttl_per_task_s: int,
 ) -> FlowRun | None:
     """Run partial updates for a batch of documents as a sub-flow or deployment."""
-    logger = get_run_logger()
-    logger.info(
-        "Running partial updates of concepts for batch as sub-flow or deployment: "
-        f"batch length {len(documents_batch)}, as_deployment: {as_deployment}"
-    )
-
     if as_deployment:
         flow_name = function_to_flow_name(run_partial_updates_of_concepts_for_batch)
         deployment_name = generate_deployment_name(flow_name=flow_name, aws_env=aws_env)
 
-        return await run_deployment(
+        flow_run: FlowRun = await run_deployment(
             name=f"{flow_name}/{deployment_name}",
             parameters={
                 "documents_batch": documents_batch,
@@ -836,7 +832,14 @@ async def run_partial_updates_of_concepts_for_batch_flow_or_deployment(
                 "concepts_counts_prefix": concepts_counts_prefix,
                 "partial_update_flow": partial_update_flow,
             },
-            timeout=3600,
+            # Return the metadata immediately
+            timeout=0,
+        )
+
+        # Now, do the actual waiting, since we have the metadata
+        return await wait_for_flow_run(
+            flow_run_id=flow_run.id,
+            timeout=ttl_per_task_s,  # Seconds
         )
 
     return await run_partial_updates_of_concepts_for_batch(
@@ -895,15 +898,27 @@ async def updates_by_s3(
     for i, updates_task_batch in enumerate(updates_task_batches, start=1):
         logger.info(f"Processing updates task batch #{i}")
 
+        ttl_per_task_s = 30 * 60
+
         updates_tasks = [
-            run_partial_updates_of_concepts_for_batch_flow_or_deployment(
-                documents_batch=documents_batch,
-                documents_batch_num=documents_batch_num,
-                cache_bucket=cache_bucket,
-                concepts_counts_prefix=concepts_counts_prefix,
-                aws_env=aws_env,
-                as_deployment=as_deployment,
-                partial_update_flow=partial_update_flow,
+            asyncio.wait_for(
+                run_partial_updates_of_concepts_for_batch_flow_or_deployment(
+                    documents_batch=documents_batch,
+                    documents_batch_num=documents_batch_num,
+                    cache_bucket=cache_bucket,
+                    concepts_counts_prefix=concepts_counts_prefix,
+                    aws_env=aws_env,
+                    as_deployment=as_deployment,
+                    partial_update_flow=partial_update_flow,
+                    ttl_per_task_s=ttl_per_task_s,
+                ),
+                # We could just rely on the timeout via Prefect, but,
+                # this may also not be running on Prefect.
+                #
+                # Realistically, there is some spin-up time on
+                # Prefect, for the task to actually be executing our
+                # code, so be aware of that.
+                timeout=ttl_per_task_s,  # Seconds
             )
             for documents_batch_num, documents_batch in enumerate(
                 updates_task_batch, start=1


flows/boundary.py
```diff
@@ -815,19 +816,14 @@ async def run_partial_updates_of_concepts_for_batch_flow_or_deployment(
     aws_env: AwsEnv,
     as_deployment: bool,
     partial_update_flow: Operation,
+    ttl_per_task_s: int,
 ) -> FlowRun | None:
     """Run partial updates for a batch of documents as a sub-flow or deployment."""
-    logger = get_run_logger()
```
THOR300, line 820: **question**: What's the reason for the logging removal, is it just as we have the prefect "flow" entity in the UI for understanding whether the task is a flow or deployment?

flows/boundary.py
```diff
@@ -895,15 +898,27 @@ async def updates_by_s3(
...
         updates_tasks = [
-            run_partial_updates_of_concepts_for_batch_flow_or_deployment(
-                documents_batch=documents_batch,
-                documents_batch_num=documents_batch_num,
-                cache_bucket=cache_bucket,
-                concepts_counts_prefix=concepts_counts_prefix,
-                aws_env=aws_env,
-                as_deployment=as_deployment,
-                partial_update_flow=partial_update_flow,
+            asyncio.wait_for(
```
THOR300, line 904: **comment**: Reading the documentation [here](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait_for) this does look like the correct method to use for what we want 👍 

I'll run a little local test as well to confirm.

THOR300: Ran the local verification to confirm to myself that this looks good: 

```python 
import asyncio


async def some_function() -> None:
    print("Starting task...")
    await asyncio.sleep(10000)
    print("Task completed!")


async def main():
    ttl_per_task_s = 1
    
    updates_tasks = [
        asyncio.wait_for(
            some_function(),
            timeout=ttl_per_task_s,
        )
        for _ in range(10)
    ]

    results = await asyncio.gather(
        *updates_tasks, return_exceptions=True
    )
    
    for result in results:
        if isinstance(result, Exception):
            print(f"Task raised an exception!")


asyncio.run(main())
```

Results confirming that this is what we want: 

```shell
Starting task...
Starting task...
Starting task...
Starting task...
Starting task...
Starting task...
Starting task...
Starting task...
Starting task...
Starting task...
Task raised an exception!
Task raised an exception!
Task raised an exception!
Task raised an exception!
Task raised an exception!
Task raised an exception!
Task raised an exception!
Task raised an exception!
Task raised an exception!
Task raised an exception!
``

File: flows/boundary.py (hunk 1 above)
Comment (jesse-c) [reply flows/boundary.py:820]: When you see the logs in Prefect, it basically is just unhelpful spam. It looks like one line repeated 20 times.

File: flows/boundary.py [hunk 3]
```diff
@@ -25,6 +25,7 @@
 from prefect import flow, get_run_logger
 from prefect.client.schemas.objects import FlowRun, StateType
 from prefect.deployments.deployments import run_deployment
+from prefect.flow_runs import wait_for_flow_run
```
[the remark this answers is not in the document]
jesse-c: We're still on Prefect v2, so it's fine! It's there in [v3 anyway](https://reference.prefect.io/prefect/flow_runs/#prefect.flow_runs.wait_for_flow_run) too. ;)

I've also included an alternative solution in the PR description, that's [v3 based](https://docs.prefect.io/v3/develop/cancel#timeouts), that could be used later.

flows/boundary.py
```diff
@@ -836,7 +832,14 @@ async def run_partial_updates_of_concepts_for_batch_flow_or_deployment(
                 "concepts_counts_prefix": concepts_counts_prefix,
                 "partial_update_flow": partial_update_flow,
             },
-            timeout=3600,
+            # Return the metadata immediately
+            timeout=0,
+        )
+
+        # Now, do the actual waiting, since we have the metadata
+        return await wait_for_flow_run(
```
THOR300, line 840: **thought**: I reckon here we use the same `asyncio.wait_for` syntax with a time out that makes sense for the deployment in question.

flows/boundary.py
jesse-c: Thank you for confirming!

flows/boundary.py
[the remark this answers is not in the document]
THOR300: Ahh thanks prefect rag interface haha! Nice!

flows/boundary.py
jesse-c: I .. admittedly don't understand what you're suggesting. 😬

flows/boundary.py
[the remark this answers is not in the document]
jesse-c: It's gotten me before, which is why I was doubly sure to check the API reference itself 😅

flows/boundary.py
THOR300: Can we just use the same `asyncio.wait_for(...)`  syntax here?

flows/count_family_document_concepts.py
```diff
@@ -344,14 +352,26 @@ async def count_family_document_concepts(
     ) in enumerate(documents_batches, start=1):
         logger.info(f"processing batch documents #{documents_batch_num}")
 
+        ttl_per_task_s = 20 * 60
```
THOR300, line 355: **thought**: When we ran the indexing pipeline last we got tasks that weren't completing inside of an hour so are we expecting this to throw errors if we run after merging? Assuming what's going to happen is we hit run, the time out of 20 mins is hit and we raise a time out? 

- Not saying this is wrong as it would be highlighting an issue with how long the tasks are taking, just confirming.

flows/deindex.py
```diff
@@ -283,14 +290,26 @@ async def cleanups_by_s3(
     for i, cleanups_task_batch in enumerate(cleanups_task_batches, start=1):
         logger.info(f"Processing clean-ups task batch #{i}")
 
+        ttl_per_task_s = 30 * 60
```
THOR300, line 293: thought: lift this up to the function params with `30*60` as the default value?

flows/inference.py
```diff
@@ -570,18 +571,45 @@ async def classifier_inference(
...
+            flow_run: FlowRun = await run_deployment(
                 name=f"{flow_name}/{deployment_name}",
                 parameters={
                     "batch": batch,
                     "config_json": config.to_json(),
                     "classifier_name": classifier_spec.name,
                     "classifier_alias": classifier_spec.alias,
                 },
-                timeout=1200,
+                timeout=0,
```
THOR300, line 587: suggestion: When I saw this I thought ahh won't we just raise immediately. Checked the prefect package and it says: 

> Setting `timeout` to 0 will return the flow run metadata immediately.

Wondering whether it's worth chucking a comment above, no dramas though.

THOR300 (approved):
Great to get this fix in! 🚀

flows/boundary.py
jesse-c: Ah, no, that wouldn't work at all? It's not the same _syntax_, it's a fundamentally different function and runtime. One is native Python, and the other is Prefect. Prefect needs to be told timeout, so this function has to be used. This is what I went into in the PR description.

At most, the native Python could be removed, and the Prefect one could be relied upon. I went with both, to be doubly sure. It can easily be removed if it's unnecessary, or problematic.

flows/count_family_document_concepts.py
jesse-c: Yup! I covered this in the PR description with:

> What if a task actually goes over? I'm not sure. I already know that this timeout will be too short for really big documents.[4](https://github.com/climatepolicyradar/knowledge-graph/pull/337/files#user-content-fn-4-74cf1b9f6e08dc6bb608b7e4da5e6f97)

So, work TBD. I have some ideas but I'm not sure yet. It may just simply need to be a lot longer, which will be a lot better, since I've fixed the resource leak.

flows/inference.py
jesse-c: Haha yup, I covered this in the beginning of the PR description, with a quote and link to the docs. I've got it as a comment in 3 other places, so I probably should add it here!

flows/count_family_document_concepts.py
jesse-c: This is for the counting pipeline anyway, where that issue is present. The (de-)indexing one is [elsewhere](https://github.com/climatepolicyradar/knowledge-graph/blob/fix%2Fproperly-timeout-on-prefect/flows/boundary.py#L901). I'll at least bump it to 60 minutes for now.

flows/deindex.py
jesse-c: I'm not sure yet what the best values are, and there already are different values, so I'm going to not be too pre-emptive. We saw this happen with batching values, and we now have different values, and also multiple values ([example](https://github.com/climatepolicyradar/knowledge-graph/blob/fix%2Fproperly-timeout-on-prefect/flows/boundary.py#L54-L55)).

PR #339, opened 2025-04-03
remove argilla v1 dependency

mpjuhasz:
## What this pr does
Finalising the previous removal of legacy code, by removing the `argilla-v1` dependency and all references to it in the codebase.


jesse-c (approved):
LGTM, though I imagine users of the now broken script may want to chime in?

harrisonpim:
i think that script was a one-off which won't need to be run again - no need to worry about it being broken!

PR #343, opened 2025-04-07
feat: Reduce CPU load on Vespa by reducing batch sizes

jesse-c:
Again, there's high CPU load[^1]. This halves the batches sizes. In practice we have nested batches. There's at least 3. That means we can end up with a lot of happening with `batchA × batchB × batchC`. In `batchC`, that's where partial updates are being performed on Vespa.

[^1]: https://climate-policy-radar.slack.com/archives/C078S46DJFP/p1744041826736619?thread_ts=1744029313.223879&cid=C078S46DJFP

TOWARDS PLA-481


THOR300 (approved):
Looking good! 

**thought**: I believe we need to change the batch size constants in `deindex.py` as well as that will result in read operations?

jesse-c:
> **thought**: I believe we need to change the batch size constants in `deindex.py` as well as that will result in read operations?

Originally I didn't think it would be a worry since it's done at a much smaller scale, and less frequently, but I've changed it now too!

PR #344, opened 2025-04-08
fix: Use canceller timeout on flows directly

jesse-c:
This approach[^1] was listed as an alternative, on the previous attempt.

The previous attempt didn't solve the resource leak, it simply moved where we incorrectly waited for a possible timeout[^2].

TOWARDS PLA-481

[^1]: https://docs.prefect.io/v3/develop/cancel#timeouts
[^2]: https://climate-policy-radar.slack.com/archives/C078S46DJFP/p1744043419589339?thread_ts=1744029313.223879&cid=C078S46DJFP


flows/boundary.py
```diff
@@ -767,7 +770,12 @@ def op_to_fn(operation: Operation):
             return run_partial_updates_of_concepts_for_document_passages__remove
 
 
-@flow
+@flow(
+    # This is the next place, after the top-level (de-)index pipeline
+    # where we want to give a timeout, that's smaller than that
+    # top-level.
+    timeout_seconds=PARENT_TIMEOUT_S,
```
THOR300, line 777: **confirmation**: Just confirming that this is meant to be the parent timeout as in `run_classifier_inference_on_batch_of_documents` we assign a task timeout level?

flows/boundary.py
```diff
@@ -851,6 +852,8 @@ async def run_partial_updates_of_concepts_for_batch_flow_or_deployment(
     )
 
 
+# No timeout it set here as it's called directly from one of the
```
THOR300, line 855: **quibble**: typo in the comment

> No timeout is set...

THOR300 (approved):
Great to get this in!

flows/boundary.py
jesse-c: **done:**

flows/boundary.py
jesse-c: Yup, they're structured differently, so it's fine, unless I've missed something.

jesse-c: Going to merge in, so I can get running of some production needs. This can be addressed after, if needed.

PR #345, opened 2025-04-08
Distil BERT-based models based on passages labelled by an LLM

harrisonpim:
## what's in this PR

- adds a `BertBasedClassifier`, which can be fine-tuned by passing a `list[LabelledPassage]` to its `fit()` method
- almost entirely rewrites the existing `LLMClassifier` to use [pydantic ai](https://ai.pydantic.dev/)
  - I'm totally sold on this framework tbh - it's allowed me to cut out a lot of the mental overhead associated with this model and i can see lots of interesting future work being enabled by it (eg see note on retry loops below)
  - `LLMClassifier.predict_batch()` now runs batches of single requests asynchronously, before gathering and returning the results. Despite running async, a regular user should never need to think about async code if they don't want to! very sexy ✨
- adds a script for distilling BERT-based models based on passages labelled by an LLM

## a few indicative numbers

I've tried this with a few concepts so far but haven't done any comprehensive evaluation / comparison yet. Running the script with a typical concept does the following:

- evenly samples 50,000 passages from the dataset
- labels ~2,500 of those 50,000 as plausible positive matches using the embedding classifier with a relatively low threshold.  this step takes ~2.5 mins to run on my laptop
- takes those candidate passages and labels them using the LLMClassifier. this usually results in around 50% of passages being labelled as positive, with the rest containing no spans. this step takes ~2 mins to run, and gives us ~2,500 positive/negative labelled passages. 
  - 2,500 is roughly the same volume of labelled passages as we used to train the targets classifier back in 2023. For what it's worth, while this process produces a similar _quantity_ of data to the targets classifier inputs, i expect the _quality_ is worse, though sufficient for the time being.
- finally, fine-tunes `climatebert/distilroberta-base-climate-f` using those ~2,500 positive/negative labelled passages, across three epochs. this step takes ~6 mins to run. i haven't though much about the fine tuning approach/config here tbh, so i expect we can improve performance with a couple of careful changes

I'll let folks know when a more complete set of predictions have been made available in the vibe checker, and/or which concepts produce classifiers which are candidates for promotion 👨‍🔬

## stuff that i'd still like to do

These will probably be added in smaller, subsequent PRs

- add some clever retry logic to the `LLMClassifier` in cases where eg the annotated output text doesn't match the input. pydantic ai (should) make this super simple with their [reflection and self correction patterns](https://ai.pydantic.dev/agents/#reflection-and-self-correction).
- think a bit more carefully about the fine tuning config for BERT-based models. I expect there are a bunch of under-used best practices which we can adopt here for some extra performance gain.
- figure out a way of getting the `reasoning` from each LLM prediction into the `metadata` of the resulting labelled passage


kdutia:
without looking at this fully tonight, it looks like there are quite a few little driveby changes to other parts of the code (including those we're using in production). are these necessary? I'm a little bit uncomfortable merging those now with little further context as to what they helped with, and whether they'll break any of our production efforts

harrisonpim:
Fair! I've tried to justify the changes below and make it a bit easier to navigate the PR. Just shout if there's anything else you're uncertain about.

As far as I can tell, the majority of files here are only affected by linter changes which CI was complaining about, mainly to do with import statements. As far as I'm concerned, the changes from [35476c6](https://github.com/climatepolicyradar/knowledge-graph/pull/345/commits/35476c6a7ec5aa210ced7c2f41b6761a7ea93ec8) onwards can be skimmed as they're only motivated by satisfying the linter/test gods. No idea why they chose to rain blood on this particular PR 🤷 

A few files are related to the substantive changes described above, and are definitely worth reviewing:
- `src/classifier/llm.py` (big change, but not used in any prod stuff yet)
- `src/classifier/bert_based.py` (big change, but not used in any prod stuff yet)
- `src/classifier/__init__.py` (allows the above classifiers to be imported)
- `src/classifier/classifier.py` (added `**kwargs` to the `fit()` method to allow `BertBasedClassifier` to take training args, shouldn't affect the behaviour of child classes)
- other files in `src/classifier` (also adapted to allow for those `kwargs`)
- `src/labelled_passage.py` (added a `sanitise` staticmethod and a `sanitised_text` property, which are used by the llm classifier)

I think the only other driveby change which isn't described by the above is related to the removal of the `DateTimeEncoder` in `src/span.py`. When I tried to load/save some labelled passages during the course of the experimentation related to this PR, I realised that the `DateTimeEncoder` was only there because we had been serialising objects to save as json using pydantic's `model_dump()` (which alone creates a dict, potentially containing non-json-serialisable objects like datetimes). Instead, we should have been using pydantic's `model_dump_json()` which serialises the data in a json-compatible format, and handles a lot of the gnarly datetime stuff automatically. Switching to `model_dump_json()` and removing the `DateTimeEncoder` should leave the behaviour unchanged while hopefully reducing the complexity of the code. The tests related to this functionality and the affected flow still pass, so i'm reasonably confident that this won't cause issues.

Are there any other files/changes that you're concerned about?

src/classifier/llm.py
```diff
@@ -1,312 +1,183 @@
...
-            for text in texts
-        }
+    def predict(self, text: str) -> list[Span]:
+        """Predict whether the supplied text contains an instance of the concept."""
+        response: AgentRunResult[LLMResponse] = self.agent.run_sync(text)
+        self._validate_response(input_text=text, response=response)
+        return Span.from_xml(
+            xml=response.data.marked_up_text,
+            concept_id=self.concept.wikibase_id,
+            labellers=[str(self)],
```
kdutia, line 141: it looks like this classifier inherits its `__repr__` method from parent `Classifier`. Given this classifier's behaviour is also dependent on model and system prompt choice, is it worth giving `LLMClassifier` a separate `__repr__` which is sensitive to changes in these?

src/classifier/llm.py
```diff
@@ -1,312 +1,183 @@
...
+        model_name: Annotated[
+            str,
+            Field(
+                description=(
+                    "The name of the model to use. See https://ai.pydantic.dev/models/ "
+                    "for a list of available models and the necessary environment "
+                    "variables needed to run each."
+                ),
+            ),
+        ] = "gemini-1.5-flash",
```
kdutia, line 75: There's a `gemini-2.0-flash` out – do we want to use that?

Also, it might be worth pointing to the specific stable version of the model here ([see docs](https://ai.google.dev/gemini-api/docs/models#gemini-1.5-flash:~:text=123-,Versions,-Read%20the%20model)) to make sure we're using the same model every time we run this.

kdutia (commented):
sorry for the inconclusive comments and thanks for the detailed comment – I got half way through a review and then got a migraine 😢 

overall looks great, but there are some minor things I think worth changing/discussing before this is merged!

src/classifier/bert_based.py
```diff
@@ -0,0 +1,171 @@
...
+    def predict_batch(self, texts: list[str]) -> list[list[Span]]:
+        """Predict whether the supplied texts contain instances of the concept."""
+        self.model.eval()
+
+        predictions = self.pipeline(texts, padding=True, truncation=True)
+
+        results = []
+        for text, prediction in zip(texts, predictions):
+            text_results = []
+            if prediction["label"] == "LABEL_1":
```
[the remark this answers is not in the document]
harrisonpim: good point, i'll clear this up!

src/classifier/classifier.py
```diff
@@ -41,11 +41,19 @@ def _validate_concept_id(self, expected_ids: Sequence[WikibaseID]) -> None:
                 f"{expected}, not {self.concept.wikibase_id}"
             )
 
-    def fit(self) -> "Classifier":
+    def fit(self, **kwargs) -> "Classifier":
```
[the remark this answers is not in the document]
harrisonpim: ### should this be an `@abstractmethod`?

Maybe, but I'm leaning towards no. IIRC making it an `@abstractmethod` would force all subclasses to implement a `fit()` implementation, even when they don't need training. The current version is a template method which can be ignored by subclasses unless it needs to be overridden

### should it be deleted?

Again, I'm leaning towards keeping it here, if only to nudge future classifier implementers towards using the same basic set of methods. The `fit()` template is a no-op, but it gives people a skeleton to build around. If we got rid of it, i'd worry that `future-contributor-A` might create a `fit()` method on their classifier while `future-contributor-B` might create a `train()` method, both with the same intended behaviour 😬 Keeping this set of basic methods around (even as no-ops) allows other bits of the codebase to treat all classifiers the same way, regardless of the underlying implementation

src/classifier/llm.py
harrisonpim: v good point! will add a `__repr__`

src/classifier/llm.py
harrisonpim: In practice, I've actually found `gemini-2.0-flash` to be quite a lot slower than `1.5`, and I don't have good enough evaluations to justify using the more recent, more expensive model yet! Figured we might as well stick to a super speedy lightweight model while working out the kinks 🤷

tests/local_vespa/test_documents/document_passage.json
```diff
@@ -721565,11 +721565,11 @@
...
             },
             {
-              "id": "Q857",
-              "name": "Q857-name"
+              "id": "Q856",
+              "name": "Q856-name"
             }
           ],
-          "parent_concept_ids_flat": "Q856,Q857,"
+          "parent_concept_ids_flat": "Q856,Q856,"
```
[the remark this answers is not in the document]
harrisonpim: hopefully explained by https://github.com/climatepolicyradar/knowledge-graph/pull/295 ! This was a tiny driveby tweak, unrelated to the rest. Thanks for calling it out though!

src/classifier/classifier.py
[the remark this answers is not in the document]
kdutia: makes sense! thanks for clarifying. the first response might be worth including in the docstring if you think so too ☺️

src/classifier/llm.py
kdutia: ok, cool! then i'd suggest making this `gemini-1.5-flash-002`? I doubt they'll release a new one, but this will act as a subtle prompt to whoever goes and experiments with model versions later to add in the specific one

harrisonpim: good idea 👌

tests/local_vespa/test_documents/document_passage.json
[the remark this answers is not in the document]
jesse-c: I'll be pleasantly surprised if this doesn't break other tests, not because it's a bad change, but because the way use fixtures is .. fun.

jesse-c (commented):
How big is the model when downloaded? I'm just wondering about the total size for space on our Prefect workers, and also the time to download.

The explicit CPU usage is good, and required!

Otherwise, I've given the most cursory of reviews. If as Platform I can go deeper anywhere, lmk.

tests/local_vespa/test_documents/document_passage.json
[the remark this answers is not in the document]
jesse-c: Haha yes, not me questioning it being done, I should've used past tense, sorry.

jesse-c:
> > How big is the model when downloaded?
> 
> The `BertBasedClassifier` models take the same backbone as the targets classifier by default, so they should be the same size if not marginally smaller.
> 
> re: practicalities like space on prefect workers, I wouldn't worry - I'm not expecting a sudden proliferation of these models in our stack any time soon. We still have a tonne of experimentation and evaluation to do so platform should have plenty of warning before we start chucking these down the pipeline 🙂

Sounds good! I was also wondering more just if it would even fit into a worker. I think they have 50gb volume atm. I imagine this will fit haha

src/classifier/llm.py
harrisonpim: Just sticking another comment here for the record to say that I've just run a small test of `gemini-2.0-flash-lite` against `gemini-1.5-flash-002` with the eval passages from argilla for forestry sector (Q787):
                                                                                             
|                                   | Precision | Recall | F1 Score | Support |
|-|-|-|-|-|
| Passage-level metrics for `gemini-2.0-flash-lite` |     1.000 |  0.789 |    0.882 |     130 |
| Passage-level metrics for `gemini-1.5-flash-002`  |     0.980 |  0.773 |    0.865 |     130 |
| Span-level (0.9) metrics for `gemini-2.0-flash-lite`    |     0.661 |  0.433 |    0.523 |     211 |
| Span-level (0.9) metrics for `gemini-1.5-flash-002`     |     0.579 |  0.415 |    0.483 |     229 |

flows/inference.py
```diff
@@ -306,10 +306,8 @@ def document_passages(
 
 
 def serialise_labels(labels: list[LabelledPassage]) -> BytesIO:
-    data = [label.model_dump() for label in labels]
-
-    # Use the datetime encoder from the span module when dumping to JSON
-    return BytesIO(json.dumps(data, cls=DateTimeEncoder).encode("utf-8"))
+    data = [label.model_dump_json() for label in labels]
+    return BytesIO(json.dumps(data).encode("utf-8"))
```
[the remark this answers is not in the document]
harrisonpim: sweet, thank you!

kdutia (approved):

src/classifier/bert_based.py
```diff
@@ -0,0 +1,177 @@
...
+        """Return a string representation of the classifier."""
+        return f'{self.name}({self.concept.preferred_label}, base_model="{self.base_model}")'
+
+    def predict(self, text: str) -> list[Span]:
+        """Predict whether the supplied text contains an instance of the concept."""
+        return self.predict_batch([text])[0]
+
+    def predict_batch(self, texts: list[str]) -> list[list[Span]]:
+        """Predict whether the supplied texts contain instances of the concept."""
+        self.model.eval()
```
mpjuhasz, line 78: We might want to use `no_grad` as a context rather than explicitly setting it here. My thinking is that in the (unlikely) case of someone 1. running `predict`, then 2. running `fit`, the model would still be set to `no_grad`, potentially creating a hard-to-debug training issue.

src/classifier/bert_based.py
```diff
@@ -0,0 +1,177 @@
...
+
+        tokenized_inputs = self.tokenizer(
+            texts, padding="max_length", truncation=True, return_tensors="pt"
+        )
+
+        return Dataset.from_dict({**tokenized_inputs, "labels": labels})
+
+    def fit(
+        self,
+        *,
```
mpjuhasz, line 130: Are we expecting any undeclared args? I don't see any used below.

mpjuhasz (commented):
🚀

PR #347, opened 2025-04-09
fix: Don't allow misleading `latest` version

jesse-c:
Another attempt at this, that's considerably simpler. I've reverted #342.

This takes it back to being Prefect parameter friendly, as the previous approach didn't work:

CleanShot 2025-04-09 at 11 14 05

FIXES PLA-515

### DIFF
--- a/flows/inference.py
+++ b/flows/inference.py
@@ -20,6 +20,7 @@ from prefect.client.schemas.objects import FlowRun, StateType
 from prefect.concurrency.asyncio import concurrency
 from prefect.deployments import run_deployment
 from prefect.logging import get_run_logger
+from prefect.task_runners import ConcurrentTaskRunner
 from pydantic import SecretStr
 from wandb.sdk.wandb_run import Run
 
@@ -27,7 +28,6 @@ from flows.utils import SlackNotify, get_file_stems_for_document_id
 from scripts.cloud import (
     AwsEnv,
     ClassifierSpec,
-    disallow_latest_alias,
     function_to_flow_name,
     generate_deployment_name,
     get_prefect_job_variable,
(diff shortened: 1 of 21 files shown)


flows/inference.py
```diff
@@ -220,10 +221,7 @@ def determine_file_stems(
 
 
 def download_classifier_from_wandb_to_local(
-    run: Run,
-    config: Config,
-    classifier_name: str,
-    alias: str,
+    run: Run, config: Config, classifier_name: str, alias: str = "latest"
```
THOR300, line 224: **question**: Do we not just want to enforce providing an alias as opposed to defaulting to downloading latest here?

flows/inference.py
```diff
@@ -522,6 +520,7 @@ async def run_classifier_inference_on_batch_of_documents(
 
 @flow(
     log_prints=True,
+    task_runner=ConcurrentTaskRunner(),
```
THOR300, line 523: **question**: Is this needed for this PR, I'm not sure I fully understand the implications? 

>     """
>     A concurrent task runner that allows tasks to switch when blocking on IO.
>     Synchronous tasks will be submitted to a thread pool maintained by `anyio`.
>     ...
>     """

THOR300 (approved):
LGTM

flows/inference.py
jesse-c: That's effectively what's been done, as before #342, `ClassifierSpec`'s `alias` field had a default value of `latest`. The 2nd commit, after reverting, explicitly shows me removing that.

This function is buried within, and is fine as it is, but I'll remove the default value to be more complete in the solution.

PR #356, opened 2025-04-15
Feature/pla 528 update pipeline to remove value error raised on non english

THOR300:
This Pull Request: 
---
- Updates the inference pipeline to no longer run on non-english documents when the run configuration that's provided is all/some classifier specs. 
- This is as we are raising a ValueError when inference is attempted on non-english documents and thus in the prefect console the runs look failed which is confusing. 
- An example of this can be seen [here](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/5f8502ac-6ecc-4979-8650-b4d60c749425). 
- Testing on staging: [run](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/b80e96f6-a582-4c25-a2bc-236cfb0234ef?entity_id=b80e96f6-a582-4c25-a2bc-236cfb0234ef)

### DIFF
--- a/flows/inference.py
+++ b/flows/inference.py
@@ -419,11 +419,24 @@ async def run_classifier_inference_on_document(
     document = load_document(config, file_stem)
     print(f"Loaded document with file stem {file_stem}")
 
-    if document.languages != ["en"] and (
-        document.pdf_data is not None or document.html_data is not None
+    # Handle documents with no text or language
+    if (
+        not document.languages
+        and document.pdf_data is None
+        and document.html_data is None
     ):
+        store_labels(
+            config=config,
+            labels=[],
+            file_stem=file_stem,
+            classifier_name=classifier_name,
+            classifier_alias=classifier_alias,
+        )
+
+    # Raise on non-english documents
+    if document.languages != ["en"]:
         raise ValueError(
-            f"Cannot run inference on {file_stem}, it has text and non-english language: "
+            f"Cannot run inference on {file_stem}, it has non-english language: "
             f"{document.languages}"
         )
 
--- a/flows/inference.py
+++ b/flows/inference.py
@@ -436,7 +436,7 @@ async def run_classifier_inference_on_document(
     # Raise on non-english documents
     if document.languages != ["en"]:
         raise ValueError(
-            f"Cannot run inference on {file_stem}, it has non-english language: "
+            f"Cannot run inference on {file_stem} as it has non-english language: "
             f"{document.languages}"
         )
 
--- a/flows/inference.py
+++ b/flows/inference.py
@@ -12,6 +12,7 @@ from uuid import UUID
 
 import boto3
 import prefect.artifacts as artifacts
+import wandb
 from cpr_sdk.parser_models import BaseParserOutput, BlockType
 from cpr_sdk.ssm import get_aws_ssm_param
 from prefect import flow
@@ -21,8 +22,8 @@ from prefect.deployments import run_deployment
 from prefect.logging import get_run_logger
 from prefect.task_runners import ConcurrentTaskRunner
 from pydantic import SecretStr
+from wandb.sdk.wandb_run import Run
 
-import wandb
 from flows.utils import (
     SlackNotify,
     filter_non_english_language_file_stems,
@@ -40,7 +41,6 @@ from scripts.update_classifier_spec import parse_spec_file
 from src.classifier import Classifier
 from src.labelled_passage import LabelledPassage
 from src.span import Span
-from wandb.sdk.wandb_run import Run
 
 # The "parent" AKA the higher level flows that do multiple things
 PARENT_TIMEOUT_S: int = int(timedelta(hours=5).total_seconds())
--- a/flows/inference.py
+++ b/flows/inference.py
@@ -419,11 +419,11 @@ async def run_classifier_inference_on_document(
     document = load_document(config, file_stem)
     print(f"Loaded document with file stem {file_stem}")
 
-    # FIXME: This makes the logs look like the run has failed and can happen like 300
-    # times per batch.
-    if document.languages != ["en"]:
+    if document.languages != ["en"] and (
+        document.pdf_data is not None or document.html_data is not None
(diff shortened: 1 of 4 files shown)


flows/utils.py
```diff
@@ -16,6 +17,8 @@
 DocumentImportId: TypeAlias = str
 DocumentStem: TypeAlias = str
 
+DOCUMENT_ID_PATTERN = re.compile(r"^((?:[^.]+\.){3}[^._]+)")
```
mpjuhasz, line 20: From memory, there can be underscores in the document ids -- would this not fail to match those?

flows/utils.py
```diff
@@ -235,3 +238,34 @@ def _s3_object_write_bytes(s3_uri: str, bytes: BytesIO) -> None:
...
+            f"File stems that do not match the pattern: {grouped['no_match']}"
+        )
+
+    filtered_file_stems = []
+    for group in grouped.values():
+        translated_file_stems = [f for f in group if "_translated" in f]
+        if translated_file_stems:
+            filtered_file_stems.extend(translated_file_stems)
+        else:
+            filtered_file_stems.extend(group)
```
mpjuhasz, line 269: I remember reviewing this same logic elsewhere 🤔 is this duplicated, or am I remembering wrong?

flows/utils.py
mpjuhasz: Never mind me -- saw the tests, and wrapped my head around this, looks good 👌

mpjuhasz (approved):
🚀

flows/utils.py
THOR300: There is a `get_file_stems_for_document_id` function as well but that was where we had a list of document ids we want to process (so we're adding translated file stems). 

In this instance we have all the file stems for the bucket and want to filter out the file stems related to non-english documents.

THOR300:
Tested on staging, unfortunately we have some documents that are still being presented from inference, this is as they have no related `${doc_id}_translated_en` file as they have no source url and thus are not presented for translated: 
image

Parser Code: 
- https://github.com/climatepolicyradar/navigator-document-parser/blob/main/cli/translate_outputs.py#L24

jesse-c:
Did not see that you were using staging! I just deployed my branch, sorry. Feel free to re-deploy your branch. LMK once you're done with staging 🙇🏻

flows/utils.py
```diff
@@ -235,3 +238,34 @@ def _s3_object_write_bytes(s3_uri: str, bytes: BytesIO) -> None:
     _ = s3.put_object(
         Bucket=bucket, Key=key, Body=bytes, ContentType="application/json"
     )
+
+
+def filter_non_english_language_file_stems(
```
jesse-c, line 243: **suggestion:** A FP approach would be to have a function that returns a boolean if it's English or not, and then use Python's builtin `filter`.

flows/utils.py
```diff
@@ -235,3 +238,34 @@ def _s3_object_write_bytes(s3_uri: str, bytes: BytesIO) -> None:
...
+        if match:
+            document_id = match.group(1)
+            grouped[document_id].append(f)
+        else:
+            grouped["no_match"].append(f)
+
+    if grouped["no_match"]:
+        raise ValueError(
+            f"File stems that do not match the pattern: {grouped['no_match']}"
+        )
```
jesse-c, line 261: **polish:** This function does not just filtering, but validation as well. I'd move validation to a separate function, or update the function name to reflect its dual purposes.

flows/utils.py
jesse-c, line 269: **suggestion:** I find the data structure choice of a list confusing, at least without it being recorded that it should only ever have length of `[1,2]`, where there's either just the original or the original + translated. I'd either use a different data structure, or add a comment.

Or, can it be length of > 2?

jesse-c (commented):
Looks good so far!

flows/inference.py
```diff
@@ -412,8 +419,21 @@ async def run_classifier_inference_on_document(
...
     print(f"Loaded document with file stem {file_stem}")
 
-    # FIXME: This makes the logs look like the run has failed and can happen like 300
-    # times per batch.
+    # Handle documents with no text or language
+    if (
+        not document.languages
+        and document.pdf_data is None
+        and document.html_data is None
+    ):
```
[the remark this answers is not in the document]
THOR300: Ahh yes good post, it's actually the comment that's the wrong definition!

mpjuhasz (approved):
Looks good 🕺

flows/utils.py
THOR300: Updated to using a functional approach, let me know what you think! I like the steer, looking much better now.

jesse-c: Agreed! It makes the data and control flows so direct. To continue the FP path, I'd remove `filter_non_english_language_file_stems` and move its code inline. Since `filter` itself doesn't need to be tested, that function doesn't need to be tested (or exist!).

Sadly vanilla Python means you need to have the `list()` wrapper, so it can look verbose.

The Clean Code approach is to make functions for everything, but the FP is to use as many built-ins as possible, and have a direct flow.

Feel free to ignore the rest of this comment, but here's some expanded things about this. An example:

```
stems
|> map(parse_str_to_stem_id)
|> filter(is_file_stem_for_english_language_document)
|> ...
```

Where `str_to_stem_id` is the previous validation about if it's a `DOCUMENT_ID` or not, and also the benefit of parsing[^1].

A tidbit with this is that this is also the Prefect transaction (with optional rollbacks) way of viewing it. https://fsharpforfunandprofit.com/rop/ is a fun way to view it.

[^1]: https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

jesse-c (approved):
LGTM

Filtering over removing the exception is a good approach.

PR #357, opened 2025-04-15
Adding in a fix for the slack notification call.

THOR300:
This Pull Request: 
---
- Adds a fix for the slack notification. 
- Essentially the load method can either return the Slack webhook object or a coroutine and thus we need to handle the different scenarios. 
- A coroutine is returned if the async_load function is called which occurs when the object is instantiated within an async function like in the KG pipeline. 
- An example of the traceback that was seen can be seen in the related linear ticket.


mpjuhasz (approved):
LGRM

PR #358, opened 2025-04-15
Removing llm group from dockerfile.

THOR300:
This Pull Request: 
---
- Is a bug fix for the docker build that was caused by an llm dependency group being present when it is not in the `pyproject.toml`. 
- Failed CI can be seen [here](https://github.com/climatepolicyradar/knowledge-graph/actions/runs/14469560735/job/40579978114).


olaughter (approved):
Good spot

PR #359, opened 2025-04-16
Adding a helper script for identifying whether classifiers have outputs.

THOR300:
This Pull Request: 
---
- Adds a script for identifying whether classifiers have outputs in s3 or not. 

_Note: This was a 2 minute write and I have no complaint if this isn't merged. Just putting here in case others find it useful as well._

### DIFF
--- a/scripts/do_classifier_specs_have_results.py
+++ b/scripts/do_classifier_specs_have_results.py
@@ -39,11 +39,20 @@ def check_classifier_specs(
 
         try:
             # List objects with the given prefix to see if path exists
-            response = s3.list_objects_v2(Bucket=bucket_name, Prefix=prefix, MaxKeys=1)
+            # Check if path exists and count objects with pagination
+            paginator = s3.get_paginator("list_objects_v2")
+            total_objects = 0
+            for page in paginator.paginate(Bucket=bucket_name, Prefix=prefix):
+                if "Contents" in page:
+                    total_objects += len(page["Contents"])
 
-            if "Contents" in response:
-                print(f"✅ S3 path exists: {s3_path}")
-            else:
+            response = {"Contents": []} if total_objects > 0 else {}
+            if total_objects > 0:
+                print(
+                    f"✅ S3 path exists: {s3_path} (contains {total_objects} objects)"
+                )
+
+            if "Contents" not in response:
                 print(f"❌ S3 path does not exist: {s3_path}")
                 to_process.append(f"{classifier_model}:{classifier_alias}")
         except ClientError as e:
--- a/scripts/do_classifier_specs_have_results.py
+++ b/scripts/do_classifier_specs_have_results.py
@@ -20,11 +20,11 @@ def check_classifier_specs(
     """
     Check if the classifier specs have s3 outputs and there respective counts.
 
-    Once inference has been run we want to validate whether the classifiers that we 
-    have listed in the classifier spec yaml files have labelled passage outputs 
-    as well as the respective counts of the labelled passages for the classifiers. 
+    Once inference has been run we want to validate whether the classifiers that we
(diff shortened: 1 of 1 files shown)


olaughter (commented):
Would love to see this set up to be run by the cli and with a docstring as we do with the other scripts, ([example](https://github.com/climatepolicyradar/knowledge-graph/blob/main/scripts/train.py#L174-L212)).

scripts/do_classifier_specs_have_results.py
```diff
@@ -0,0 +1,38 @@
...
+
+s3 = boto3.client("s3")
+
+to_process = []
+for classifier_spec in data:
+    classifier_model = classifier_spec.split(":")[0]
+    classifier_alias = classifier_spec.split(":")[1]
+
+    labelled_passages_s3_path = (
+        PROD_LABELLED_PASSAGES_S3_PATH + classifier_model + "/" + classifier_alias + "/"
```
olaughter, line 19: Safer to use os.path.join, here?

scripts/do_classifier_specs_have_results.py
```diff
@@ -0,0 +1,38 @@
+import boto3
+import yaml
+from botocore.exceptions import ClientError
+
+PROD_YAML = "flows/classifier_specs/prod.yaml"
+PROD_LABELLED_PASSAGES_S3_PATH = "s3://cpr-prod-data-pipeline-cache/labelled_passages/"
+
+with open(PROD_YAML, "r") as file:
+    data = yaml.safe_load(file)
```
olaughter, line 9: Nice to parameterise these so we can just select "prod" or "staging" and see the results

scripts/do_classifier_specs_have_results.py
```diff
@@ -0,0 +1,38 @@
...
+    # Parse the S3 URI to get bucket and prefix
+    bucket_name = labelled_passages_s3_path.split("/")[2]
+    prefix = "/".join(labelled_passages_s3_path.split("/")[3:])
+
+    try:
+        # List objects with the given prefix to see if path exists
+        response = s3.list_objects_v2(Bucket=bucket_name, Prefix=prefix, MaxKeys=1)
+
+        if "Contents" in response:
+            print(f"✅ S3 path exists: {labelled_passages_s3_path}")
```
olaughter, line 31: Length of the contents object could be useful detail too

scripts/do_classifier_specs_have_results.py
```diff
@@ -0,0 +1,65 @@
...
+        classifier_alias = classifier_spec.split(":")[1]
+
+        # Use os.path.join for path construction
+        s3_path = os.path.join(
+            labelled_passages_s3_path, classifier_model, classifier_alias
+        )
+
+        # Parse the S3 URI to get bucket and prefix
+        bucket_name = s3_path.split("/")[2]
+        prefix = "/".join(s3_path.split("/")[3:])
```
olaughter, line 38: This feels a bit too flakey, any thoughts on what might keep us confident its looking at the right things?

scripts/do_classifier_specs_have_results.py
```diff
@@ -0,0 +1,65 @@
...
+@app.command()
+def check_classifier_specs(
+    yaml_path: str = typer.Argument(
+        help="Path to the YAML file containing classifier specifications"
+    ),
+    labelled_passages_s3_path: str = typer.Argument(
+        help="S3 path where labelled passages should be stored"
+    ),
+) -> None:
+    """Check if the classifier specs in the YAML file have corresponding S3 paths."""
```
olaughter, line 20: Be good to flesh this out a bit more and highlight when and why its useful

olaughter (approved):
Looks good, couple of small improvements below

PR #361, opened 2025-04-16
Make sure LLM classifier outputs are identical to the inputs

harrisonpim:
- a tiny, subtle bugfix for the `LLMClassifier` that @mpjuhasz and i discussed earlier today
- squashes two other little type-ish bugs 🐛🔨


mpjuhasz (approved):
Looks good 🚀

PR #363, opened 2025-04-17
refactor: Use Vespa connection pool

jesse-c:
Instead of creating new connections every time we interact with Vespa, create a pool instead. This is intended to reduce the overhead of creating a new network connection and the Python objects. This would've had a larger effect when we were doing per-document passage read requests, but is still helpful.

**Testing**

Successful run: https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/92adf6f0-8547-4010-becf-3d6f695ec39f?entity_id=92adf6f0-8547-4010-becf-3d6f695ec39f

FIXES PLA-522

### DIFF
--- a/flows/boundary.py
+++ b/flows/boundary.py
@@ -3,7 +3,6 @@
 import asyncio
 import base64
 import json
-import math
 import os
 import re
 import tempfile
@@ -970,25 +969,6 @@ async def updates_by_s3(
         raise ValueError("there was at least 1 task that failed")
 
 
-def connections_from_batch_size(
-    items: list[T],
-    batch_size: int,
-) -> PositiveInt:
-    """
-    Get an appropriate number of connections for items in batches.
-
-    Calculates a minimum of 1 connection.
-    """
-    return max(
-        # Guarantee: 1 if the list is empty and `ceil(0/batch size)` is none.
-        1,
-        # Guarantee: 1 if the quotient is a fraction. E.g. 2 items
-        # and a batch size of 10 would be 0.2 at 2/10. Using the
-        # ceiling gives 1.
-        math.ceil(len(items) / batch_size),
-    )
-
-
 # Index -------------------------------------------------------------------------
 
 
@@ -1037,11 +1017,6 @@ async def run_partial_updates_of_concepts_for_document_passages__update(
         batch_size=DEFAULT_DOCUMENTS_BATCH_SIZE,  # How many tasks to have running at once
     )
 
-    # Get the ceiling so the minimum is "above" e.g. if the result is < 0.1 it becomes 1.
-    vespa_connections_n = connections_from_batch_size(
-        list(grouped_concepts.items()), DEFAULT_DOCUMENTS_BATCH_SIZE
-    )
-
     document_import_id = remove_translated_suffix(document_importer[0])
 
     text_blocks_ids: list[TextBlockId] = list(grouped_concepts.keys())
@@ -1055,9 +1030,11 @@ async def run_partial_updates_of_concepts_for_document_passages__update(
             vespa_public_cert_param_name="VESPA_PUBLIC_CERT_FULL_ACCESS",
         )
 
-    async with vespa_search_adapter.client.asyncio(  # pyright: ignore[reportOptionalMemberAccess]
-        connections=vespa_connections_n
-    ) as vespa_connection_pool:
+    async with (
+        vespa_search_adapter.client.asyncio(  # pyright: ignore[reportOptionalMemberAccess]
+            connections=DEFAULT_DOCUMENTS_BATCH_SIZE,  # How many tasks to have running at once
+        ) as vespa_connection_pool
+    ):
         # Read all the document passages from Vespa in as fewer reads as possible
         text_blocks: dict[TextBlockId, tuple[VespaHitId, VespaPassage]] = {}
 
@@ -1253,13 +1230,11 @@ async def run_partial_updates_of_concepts_for_document_passages__remove(
 
     has_failures = False
 
-    vespa_connections_n = connections_from_batch_size(
-        list(grouped_concepts.items()), DEFAULT_DOCUMENTS_BATCH_SIZE
-    )
-
-    async with vespa_search_adapter.client.asyncio(  # pyright: ignore[reportOptionalMemberAccess]
-        connections=vespa_connections_n
-    ) as vespa_connection_pool:
+    async with (
+        vespa_search_adapter.client.asyncio(  # pyright: ignore[reportOptionalMemberAccess]
+            connections=DEFAULT_DOCUMENTS_BATCH_SIZE,  # How many tasks to have running at once
+        ) as vespa_connection_pool
+    ):
         # Read all the document passages from Vespa in as fewer reads as possible
         text_blocks: dict[TextBlockId, tuple[VespaHitId, VespaPassage]] = {}
 
--- a/flows/boundary.py
+++ b/flows/boundary.py
@@ -3,6 +3,7 @@
 import asyncio
 import base64
 import json
+import math
 import os
 import re
 import tempfile
@@ -15,6 +16,7 @@ from pathlib import Path
 from typing import Any, TypeAlias, TypeVar, Union
 
 import boto3
+import httpx
 import vespa.application
 import vespa.querybuilder as qb
 from cpr_sdk.models.search import Concept as VespaConcept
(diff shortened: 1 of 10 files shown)


flows/boundary.py
```diff
@@ -443,39 +443,29 @@ def convert_labelled_passage_to_concepts(
     return concepts
 
 
-def get_vespa_search_adapter(
-    vespa_search_adapter: VespaSearchAdapter | None,
-) -> tuple[
-    ContextManager[str] | ContextManager[None],
-    VespaSearchAdapter,
-]:
+def get_vespa_search_adapter() -> VespaSearchAdapter:
```
jesse-c, line 446: **note:** Refactored this to not require a context manager, so I don't need to create an asynchronous friendly context manager—I did try, and it wasn't fun.

flows/boundary.py
```diff
@@ -563,10 +553,10 @@ def get_document_passage_from_vespa(
     return passage_id, passage
 
 
-def get_document_passages_from_vespa(
+async def get_document_passages_from_vespa(
```
jesse-c, line 556: **note:** Using an asynchronous connection pool now, so made this asynchronous.

tests/flows/conftest.py
```diff
@@ -146,13 +146,19 @@ def vespa_app(
...
+    instance_url = "http://localhost:8080"
+    adapter = VespaSearchAdapter(
+        instance_url=instance_url,
     )
+
+    # We can't currently optionally use certs with our search adapter.
+    #
+    # Instead, overwrite it here.
+    adapter.client = Vespa(url=instance_url, cert=None)
+    print("adapter.client.cert", adapter.client.cert)
```
jesse-c, line 160: **note:** In the `pyvespa` docs, they don't use certs for local Vespa.

tests/flows/conftest.py
```diff
@@ -146,13 +146,18 @@ def vespa_app(
...
+    """VespaSearchAdapter instantiated from mocked SSM params."""
+    instance_url = "http://localhost:8080"
+    adapter = VespaSearchAdapter(
+        instance_url=instance_url,
     )
+
+    # We can't currently optionally use certs with our search adapter.
+    #
+    # Instead, overwrite it here.
+    adapter.client = Vespa(url=instance_url, cert=None)
```
jesse-c, line 159: **note:** In the `pyvespa` docs, they don't use certs for local Vespa.

Using the mocked certs, was giving me a headache with SSL and the new connection pool. It was throwing some invalid SSL error.

flows/boundary.py
```diff
@@ -1218,7 +1221,7 @@ async def run_partial_updates_of_concepts_for_document_passages__remove(
         f"starting partial updates for {len(grouped_concepts)} grouped concepts"
     )
 
-    batches = iterate_batch(
+    tasks_batches = iterate_batch(
```
jesse-c, line 1224: **note:** A more informative name, since "batch" is around a lot.

flows/boundary.py
```diff
@@ -1229,28 +1232,30 @@ async def run_partial_updates_of_concepts_for_document_passages__remove(
 
     has_failures = False
 
-    with cm:
+    # Get the ceiling so the minimum is "above" e.g. if the result is < 0.1 it becomes 1.
+    vespa_connections_n = math.ceil(
```
jesse-c, line 1236: **note:**  Similar anti-DRY motivation for the time being, as on my previous PR for batch reads.

flows/boundary.py
```diff
@@ -1229,28 +1232,30 @@ async def run_partial_updates_of_concepts_for_document_passages__remove(
...
+
+    async with vespa_search_adapter.client.asyncio(  # pyright: ignore[reportOptionalMemberAccess]
+        connections=vespa_connections_n
+    ) as vespa_connection_pool:
         # Read all the document passages from Vespa in as fewer reads as possible
-        def collect_text_blocks(
-            acc: dict[TextBlockId, tuple[VespaHitId, VespaPassage]],
-            text_blocks_ids_batch: list[TextBlockId],
-        ) -> dict[TextBlockId, tuple[VespaHitId, VespaPassage]]:
+        text_blocks: dict[TextBlockId, tuple[VespaHitId, VespaPassage]] = {}
```
jesse-c, line 1244: **note:** Sadly the now async method means that we lost the FP conciseness.

flows/count_family_document_concepts.py
```diff
@@ -203,7 +203,8 @@ async def load_update_document_concepts_counts(
 
             logger.info(f"processed batch document object URIs #{batch_num}")
 
-    cm, vespa_search_adapter = get_vespa_search_adapter(vespa_search_adapter)
+    if vespa_search_adapter is None:
+        vespa_search_adapter = get_vespa_search_adapter()
```
jesse-c, line 207: **note:** I find this more _direct_, and as you can see, I had to think about the Vespa search adapter a lot for this issue. 😅

tests/local_vespa/README.md
jesse-c, line 1: **note:** Just my own self-effacing quibbles.

flows/boundary.py
```diff
@@ -833,12 +824,16 @@ async def run_partial_updates_of_concepts_for_batch(
...
                 cache_bucket=cache_bucket,
                 concepts_counts_prefix=concepts_counts_prefix,
+                vespa_search_adapter=vespa_search_adapter,
             )
 
             logger.info(f"processed batch documents #{documents_batch_num}")
 
         except Exception as e:
             document_stem: DocumentStem = documents_batch[i][0]
+            import traceback
```
olaughter, line 834: **Question**: Is this here because it can cause issues importing it globally? Or is there some other reason to do this here?

flows/boundary.py
```diff
@@ -833,12 +824,16 @@ async def run_partial_updates_of_concepts_for_batch(
...
+
+            traceback.print_exc()  # This prints the full stacktrace
             logger.error(
                 f"failed to process document `{document_stem}`: {e.__str__()}",
             )
```
olaughter, line 839: Wonder if its worth combining these two?

flows/boundary.py
```diff
@@ -1005,8 +1004,6 @@ async def run_partial_updates_of_concepts_for_document_passages__update(
     """
     logger = get_run_logger()
 
-    cm, vespa_search_adapter = get_vespa_search_adapter(vespa_search_adapter)
-
```
olaughter, line 1009: **praise**: Nice to see how this approach leads to a cleaner abstraction

flows/boundary.py
```diff
@@ -1031,32 +1028,37 @@ async def run_partial_updates_of_concepts_for_document_passages__update(
         batch_size=DEFAULT_DOCUMENTS_BATCH_SIZE,  # How many tasks to have running at once
     )
 
+    # Get the ceiling so the minimum is "above" e.g. if the result is < 0.1 it becomes 1.
+    vespa_connections_n = math.ceil(
+        len(list(grouped_concepts.items())) / DEFAULT_DOCUMENTS_BATCH_SIZE
+    )
+
```
olaughter, line 1035: **question**: Not sure I understand what is happening here, is this basically ensuring we always have at least one connection?

tests/flows/conftest.py
olaughter: Makes sense, this might have been an overhang from when the sdk expected credentials to work (which Mark has since fixed)

tests/local_vespa/README.md
olaughter: Good grammar is overrated 😛

olaughter (approved):
Looks great! A nice upgrade!

flows/boundary.py
```diff
@@ -1031,32 +1028,37 @@ async def run_partial_updates_of_concepts_for_document_passages__update(
...
     document_import_id = remove_translated_suffix(document_importer[0])
 
     text_blocks_ids: list[TextBlockId] = list(grouped_concepts.keys())
 
-    with cm:
+    if vespa_search_adapter is None:
+        vespa_search_adapter = get_vespa_search_adapter()
+
+    async with vespa_search_adapter.client.asyncio(  # pyright: ignore[reportOptionalMemberAccess]
```
THOR300, line 1043: **thought**: Is this pyright ignore needed now that we have the conditional check above as to whether the vespa_search_adapter is none?

flows/boundary.py
jesse-c: I was silly about the temp. directory though, since it was being cleaned-up after the function returned. I've made it a little less DRY again.

flows/boundary.py
jesse-c: Yes! I'll move this to a function.

flows/boundary.py
jesse-c: **update:** Ended up removing this in-between function all together[^1].

[^1]: https://github.com/climatepolicyradar/knowledge-graph/pull/363#discussion_r2053842416

flows/boundary.py
jesse-c: Good point! It doesn't appear to. I've removed it, and similar ones elsewhere.

flows/boundary.py
```diff
@@ -973,7 +1001,7 @@ async def updates_by_s3(
 
 
 # No timeout set since the caller of this has one.
-@flow
+@flow(log_prints=True)
```
jesse-c, line 1004: **note:** Since `get_document_passages_from_vespa` isn't a flow or task, the logger messages weren't bring printed in Prefect. Now they are, as normal print statements.

flows/boundary.py
```diff
@@ -568,13 +580,29 @@ async def get_document_passages_from_vespa(
 
         conditions &= text_block_id.contains(text_block_id_contains)
 
+    text_blocks_ids_n: PositiveInt = (
+        VESPA_MAX_LIMIT if text_blocks_ids is None else len(text_blocks_ids)
```
THOR300, line 584: thought: Is there a potential issue here where `len(text_block_ids)` is greater than `VESPA_MAX_LIMIT`?

THOR300 (approved):
LGTM!

flows/boundary.py
jesse-c: There's a check for that [earlier](https://github.com/climatepolicyradar/knowledge-graph/blob/refactor%2Fuse-vespa-connection-pool/flows/boundary.py#L568-L571),

PR #368, opened 2025-04-23
feat: Batch partial updates

jesse-c:
Under-the-hood, `feed_iterable` uses the same approach we do in that we use a connection pool and tasks, albeit they differ in the tasks approach.

There may be performance improvements since they presumably know Vespa better. It also let's us defer the implementation to them for maintenance and improvement.

I don't think we need fine-grained control beyond what's provided.

There doesn't appear to be another way to do batch writes that would perform better[^1].

**Testing**

[Deploy](https://github.com/climatepolicyradar/knowledge-graph/actions/runs/14706134271/job/41266946412) and [successful run](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/84e40723-f221-48a8-bdef-0fc4dcadfe4b?entity_id=84e40723-f221-48a8-bdef-0fc4dcadfe4b).

FIXES PLA-521

[^1]: https://docs.vespa.ai/en/reads-and-writes.html


flows/boundary.py
```diff
@@ -1089,87 +1101,90 @@ async def run_partial_updates_of_concepts_for_document_passages__update(
...
-            logger.info(
-                f"gathered partial {len(results)} updates tasks for batch {batch_num}"
-            )
+            concepts_counts.update(concepts_models)
 
-            for i, result in enumerate(results):
-                text_block_id, concepts = batch[i]
+    class DataPoint(TypedDict):
+        id: VespaDataId
+        fields: dict[str, Any]
```
olaughter, line 1134: 👌 nice

flows/boundary.py
```diff
@@ -1089,87 +1101,90 @@ async def run_partial_updates_of_concepts_for_document_passages__update(
...
+    # The previously established connection pool isn't used since
+    # `feed_iterable` creates its own.
+    vespa_search_adapter.client.feed_iterable(  # pyright: ignore[reportOptionalMemberAccess]
+        iter=data,
+        schema="document_passage",
+        namespace="doc_search",
+        operation_type="update",
+        max_connections=DEFAULT_DOCUMENTS_BATCH_SIZE,  # How many tasks to have running at once
+        callback=_feed_result_callback,
+    )
```
olaughter, line 1162: Cool! Nice to see we can set the operation_type to update and use this for updates. Does it cause a failure if the passage is missing, or will it try and create it?

flows/boundary.py
```diff
@@ -1089,87 +1101,90 @@ async def run_partial_updates_of_concepts_for_document_passages__update(
...
+    # We need to count the concepts across all document passages
+    concepts_counts: Counter[ConceptModel] = Counter()
 
-        # For each document passage, update (AKA merge) its concepts.
-        #
-        # We use several tasks at once to speed this up.
-        for batch_num, batch in enumerate(tasks_batches, start=1):
-            logger.info(f"processing partial updates batch {batch_num}")
+    # Batch updates (writes) and let the Vespa SDK take care of the complexities
+    def _feed_result_callback(response: VespaResponse, data_id: VespaDataId) -> None:
```
olaughter, line 1108: The callback is a nice way of organising this 👍

flows/boundary.py
```diff
@@ -1089,87 +1101,90 @@ async def run_partial_updates_of_concepts_for_document_passages__update(
...
+        if not response.is_successful():
+            raise PartialUpdateError(data_id, response.get_status_code())
```
olaughter, line 1110: Related to my point below, I guess we might get a unsuccssful response because a passage is missing, or credentials are mis configured, or a temporary network issue. Do we want to try and handle some of these more specifically? Or perhaps add some other context from the response to the error?.

flows/boundary.py
```diff
@@ -1089,87 +1101,90 @@ async def run_partial_updates_of_concepts_for_document_passages__update(
...
-            )
+    data: Iterable[dict[str, Any]] = list(
+        map(lambda x: dict(_to_data(*x)), grouped_concepts.items())
+    )
 
-            serialised_concepts_counts = json.dumps(
-                {str(k): v for k, v in concepts_counts.items()}
-            )
+    # The previously established connection pool isn't used since
+    # `feed_iterable` creates its own.
```
olaughter, line 1154: Did we work out in the end wether `feed_async_iterable` was useful for this?

flows/boundary.py
```diff
@@ -1089,87 +1101,90 @@ async def run_partial_updates_of_concepts_for_document_passages__update(
...
+    vespa_search_adapter.client.feed_iterable(  # pyright: ignore[reportOptionalMemberAccess]
+        iter=data,
+        schema="document_passage",
+        namespace="doc_search",
+        operation_type="update",
+        max_connections=DEFAULT_DOCUMENTS_BATCH_SIZE,  # How many tasks to have running at once
```
olaughter, line 1160: There are some other arguments that we could also consider customising here, not sure if you already looked into them:
- max_queue_size: The maximum size of the blocking queue and max in-flight operations.
- max_workers: The maximum number of workers in the threadpool executor

olaughter (commented):
Neat, left some thoughts on the draft

flows/boundary.py
THOR300: From the docs: 
_If the document to update is not found, it is not considered a failure._

olaughter: Nice, thanks! And good!

flows/boundary.py
jesse-c: I was looking at this earlier, and it says:

> Prefer using this method over feed_iterable when the operation is I/O bound from the client side.

There are network requests happening around Vespa (e.g. getting the document passages) _before_ this point. I wasn't sure if it's useful at _this_ point though.\

It might be though since the big parent function is being run in a batch[^1] on one host/container.

[^1]: https://github.com/climatepolicyradar/knowledge-graph/blob/feat%2Fbatch-updates/flows/boundary.py#L825-L839

flows/boundary.py
jesse-c: I was thinking of trusting the defaults for now as I've been guessing a lot of numbers lately for things like this 😅

flows/boundary.py
jesse-c: I think that is a good idea in general. I'd say it could be a good future improvement. It could, for example, batch failures and do another feed after this feed finishes.

flows/boundary.py
jesse-c: I'll change it to be the async version.

flows/boundary.py
THOR300: **thought**: Also just lifting out the following:

Our `DEFAULT_DOUMENTS_BATCH_SIZE` is 50 whereas the default for max connections is 16.
-  Are we happy with a ~300% increase on the default? 
- I'm just wondering whether when we're running 20 ecs tasks in parallel whether 1000 connections becomes problematic.

jesse-c: > Are we happy with a ~300% increase on the default?

I have no idea about the motivation for the default, so a ~300% or 5% or -100% increase has no meaning 😅

> I'm just wondering whether when we're running 20 ecs tasks in parallel whether 1000 connections becomes problematic.

I'm not sure, without prior data and benchmarking. This PR doesn't change the number of connections, as `DEFAULT_DOCUMENTS_BATCH_SIZE` is already being used in the connection pool.

flows/boundary.py
jesse-c: May not be possible:

```
FAILED tests/flows/test_index.py::test_run_partial_updates_of_concepts_for_document_passages - RuntimeError: asyncio.run() cannot be called from a running event loop
```

flows/boundary.py
```diff
@@ -643,6 +643,23 @@ def get_data_id_from_vespa_hit_id(hit_id: VespaHitId) -> VespaDataId:
...
+    Extract just the text block ID from a fully qualified Vespa ID for it.
+
+    Example:
+    "1273" from Vespa data ID like "CCLW.executive.10014.4470.1273".
+    """
+    splits = data_id.split(".")
+    expected_splits = 2
+    if len(splits) < expected_splits:
+        raise ValueError(
+            f"received {len(splits)} splits, when expecting ≥ expected_splits: {splits}"
```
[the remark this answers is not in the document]
jesse-c: Do they always have 5 splits? I've never really thought about the dotted ID structure!

jesse-c:

- #368 ◀
    - #370

tests/flows/test_deindex.py
```diff
@@ -224,108 +221,6 @@ def test_remove_concepts_from_existing_vespa_concepts_non_matching():
     assert result_non_matching[2].get("id") == "Q789"
 
 
-@pytest.mark.asyncio
-@pytest.mark.vespa
-async def test_partial_update_text_block_with_removal(
```
THOR300, line 229: **suggestion**: Looks like we've removed all of the tests and invocations of the `partial_update_text_block` function so we could remove the function aswell for clarity.

THOR300: **suggestion**: Seeing as we're essentially replacing this function with the code block in the `run_partial_updates_of_concepts_for_document_passages__remove` function (`_feed_result_callback` function, `_to_data` function, `data` list comprehension and `feed_iterable`) we could lift that out to a separate function and add unit tests like we had for `partial_update_text_block`.

tests/flows/test_deindex.py
```diff
@@ -224,108 +221,6 @@ def test_remove_concepts_from_existing_vespa_concepts_non_matching():
...
-    local_vespa_search_adapter: VespaSearchAdapter,
```
THOR300, line 230: **suggestion**: Seeing as we're essentially replacing this function with the code block in the `run_partial_updates_of_concepts_for_document_passages__remove` function (`_feed_result_callback` function, `_to_data` function, `data` list comprehension and `feed_iterable`) we could lift that out to a separate function and add unit tests like we had for `partial_update_text_block`.

THOR300 (approved):
Huge to get this in, two non-blocking comments from me.

flows/boundary.py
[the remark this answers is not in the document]
olaughter: Yeah the id situation is weird to be honest, feels like a lot of unnecessary complexity is squeezed into it, but I believe its always consistently the four part document id appended by the passage id. Its something like: source, type, document numbering x2, then the passage count. I'm not sure why four parts where chosen for the doc id over one, but here we are! 😄

flows/boundary.py
olaughter: Aha that makes sense

olaughter (approved):

flows/boundary.py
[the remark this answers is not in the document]
jesse-c: Haha thanks for the info. Updated it and added tests.

tests/flows/test_deindex.py
jesse-c: It's still being used in setting up a test. I'll move it to the test module.

tests/flows/test_deindex.py
jesse-c: In https://github.com/climatepolicyradar/knowledge-graph/pull/370 I've already moved it out. Just need to add tests!

PR #369, opened 2025-04-24
Active learning 

mpjuhasz:
> [!NOTE]
> This PR was done in May, but due to the sensitive nature of the KG launch, we've decided against merging it, due to its risk. As I'm leaving on the 5th of June, @harrisonpim will be picking up the merging of this.

## What this PR does
This PR started off as adding Active Learning, but to do that properly, I've had to make various other changes.

### Passage and Source
Introducing a few new concepts here, and extending them to LabelledPassage:
- The new `src/models` folder now contains base models, separated out into 4 scripts
- `Passage`: a base-model object with text and a `Source`. The `LabelledPassage` now inherits from this, extending it with the `Source` object to track where it originates from.
> [!WARNING] 
> Because I moved the `LabelledPassage` definition, now all the pickles are broken. I've worked around it by creating a custom unpickling that updates the hard-coded object definition path to the new one. It feels super hacky, and future generations will decide whether it is to be kept 🤷‍♂️ 

- Further extensions of the `Passage` class include: `SyntheticPassageWithClassifierConfidence`, `SyntheticPassageWithConfidence` and `PassageWithClassifierConfidence`. These are used for synthetic data creation and confidence tracking. 

Also note, that I've had some troubles with syrupy. Getting it working involved adding an `id` field to the `Passage`, otherwise the order of attributes in the `__repr__` (which apparently syrupy uses) is wrong, resulting in all tests failing.

### Data
To create some base abstractions, we now have the `SyntheticData` and `ActiveLearningData` classes within the `models` folder. 
- `ActiveLearningData` is a base class with core functionality such as passage filtering using a classifier
- It has two extensions: `ActiveLearningCorpusData` and `ActiveLearningSyntheticData`, which use the above methods on passages samples from a CPR corpus, or generated synthetically respectively

### Confidences
Confidences can now be tracked on the passage, thanks to the extensions. There are 2 types of confidences introducted in this PR:
- expected confidence: for synthetic data this is a field for the data generator to provide an expectation for how the passage will perform when passed to the classifier. It can be used to feedback and tune the generator
- actual confidence: this is what we usually mean by "confidence score", i.e. the probability attached to the passage by some classifier. 
> Disclaimer: I think these concepts could be refined, and I'd imagine when we start using them, it'll become clear where the limitations are. One immediate thought: we might want to track, which classifier a specific confidence refers to.

Building on top of these confidences, the pooler has also got an upgrade:
- the `_aggregate_two_overlapping_span_confidence` method introduces a new way for aggregating confidence, breaking away from the Jaccard-index
- using this, the `Pooler` can aggregate the predictions of multiple (types) of classifiers, opening the door to using multiple synthetic data creators in parallel
- I've left a TODO, as I'm still not happy with the implementation, but again after using this metric, it'll be clearer what needs to be improved

### Minor changes:
- refactors the targets classifiers in a way that they now attach confidence scores to spans
- adds streaming to `HuggingfaceSession` (so that our whole dataset does not need to be pulled for active learning)
- refactors the existing `ClassifierPooler` to use span confidences rather than new constructs, and to take into account the already existing confidences on the spans for the aggregation

#### Synthetic data for active learning
Here's an example illustrating the combination of synthetic and active learning run for the targets classifier:

```python
sd = ActiveLearningSyntheticData(
    concept=concept,
    human_labelled_passages=labelled_passages_from_argilla,
    classifier=TargetClassifier(concept=concept),
    model_name="gemini-2.0-flash",
)

batch = sd.generate(num_samples=10, max_iterations=30)

console.log(batch)

>>> 

[
    SyntheticPassageWithClassifierConfidence(
        text="The country is committed to planting one million trees annually over the next decade to offset carbon emissions.",
        source=Source(
            type="Synthetic",
            model="gemini-2.0-flash",
            prompt="system_prompt_v1",
            document_id=None,
        ),
        expected_confidence=0.5,
        actual_confidence=0.41033363342285156,
    ),
    SyntheticPassageWithClassifierConfidence(
        text="The project envisions developing a carbon capture and storage facility capable of sequestering 5 million tons of CO2 annually by 2035, contingent on securing international funding and technological advancements.",
        source=Source(
            type="Synthetic",
            model="gemini-2.0-flash",
            prompt="system_prompt_v1",
            document_id=None,
        ),
        expected_confidence=0.5,
        actual_confidence=0.41398000717163086,
    ),
    SyntheticPassageWithClassifierConfidence(
        text="The framework proposes establishing a national green fund with a target capitalization of $5 billion by 2028 to support climate mitigation and adaptation projects, pending parliamentary approval.",
        source=Source(
            type="Synthetic",
            model="gemini-2.0-flash",
            prompt="system_prompt_v1",
            document_id=None,
        ),
        expected_confidence=0.5,
        actual_confidence=0.44369399547576904,
    ),
    SyntheticPassageWithClassifierConfidence(
        text="The authority anticipates a 20% decrease in water consumption in the agricultural sector by 2030, achieved through the adoption of efficient irrigation technologies and drought-resistant crops, subject to farmers' willingness   to adopt new methods and the availability of financial incentives.",
        source=Source(
            type="Synthetic",
            model="gemini-2.0-flash",
            prompt="system_prompt_v1",
            document_id=None,
        ),
        expected_confidence=0.5,
        actual_confidence=0.6530038714408875,
    ),
    SyntheticPassageWithClassifierConfidence(
        text="The country is committed to enhancing carbon sequestration in agricultural soils, with a goal to sequester an additional 5 million tons of CO2 equivalent annually by 2035, assessed through soil sampling and modeling techniques.",
        source=Source(
            type="Synthetic",
            model="gemini-2.0-flash",
            prompt="system_prompt_v1",
            document_id=None,
        ),
        expected_confidence=0.5,
        actual_confidence=0.5559934377670288,
    ),
    SyntheticPassageWithClassifierConfidence(
        text="The project targets to create 5000 green jobs by 2028, through investments in renewable energy and sustainable agriculture, contingent on attracting sufficient private sector investment and international aid .",
        source=Source(
            type="Synthetic",
            model="gemini-2.0-flash",
            prompt="system_prompt_v1",
            document_id=None,
        ),
        expected_confidence=0.5,
        actual_confidence=0.5115469694137573,
    ),
    SyntheticPassageWithClassifierConfidence(
        text="The council aims to enhance the city's green infrastructure by planting 5,000 trees annually for the next five years, conditional on securing community involvement, addressing potential conflicts with existing infrastructure, and ensuring long-term maintenance and watering of the newly planted trees .",
        source=Source(
            type="Synthetic",
            model="gemini-2.0-flash",
            prompt="system_prompt_v1",
            document_id=None,
        ),
        expected_confidence=0.5,
        actual_confidence=0.5378943681716919,
    ),
]


```

#### Corpus data for active learning
Similarly, we can do this with sampling from our corpus as well:

```python
hf = HuggingfaceSession()
wikibase = WikibaseSession()
concept = wikibase.get_concept("Q1651")

al = ActiveLearningCorpusData(
    classifier=TargetClassifier(concept=concept),
    upper_bound=0.7,
    lower_bound=0.3,
)

examples = al.find_examples(2, hf.stream("all-document-text-data-weekly"), skip=20)

>>>
[
    PassageWithClassifierConfidence(
        text="Update of the water code to ensure explicit recognition of on climate impacts on water resource availability by end of 2013 At least 2 sets of sub- regulations developed under the Water Code to implement a) progressive and differentiated tariffs, b) support for water delivery services under communal management",
        source=Source(
            type="Natural",
            model=None,
            prompt=None,
            document_id="AF.document.003MTMWR.n0000",
        ),
        confidence=0.3268326222896576,
    ),
    PassageWithClassifierConfidence(
        text="By 2018 at least five lessons learned materials produced and disseminated.",
        source=Source(
            type="Natural",
            model=None,
            prompt=None,
            document_id="AF.document.017MERRR.n0006",
        ),
        confidence=0.5626274943351746,
    ),
]
```


src/passage.py
```diff
@@ -0,0 +1,70 @@
...
+            if not self.document_id:
+                raise ValueError("document_id must be provided for natural passages")
+        return self
+
+
+class Passage(BaseModel):
+    """Raw passage object"""
+
+    text: str
+    source: Optional[Source] = Field(default=None)
```
harrisonpim, line 32: would love to blend this with the existing `LabelledPassage` class!

src/data.py
```diff
@@ -0,0 +1,255 @@
...
+                    continue
+
+                synth_passage = SyntheticPassageWithClassifierConfidence(
+                    text=example.text,
+                    expected_confidence=example.expected_confidence,
+                    actual_confidence=actual_confidence,
+                    source=Source(
+                        type="Synthetic",
+                        model=self.model_name,
+                        prompt="system_prompt_v1",  # TODO: need to decide whether we want to retain the whole prompt
```
harrisonpim, line 173: maybe we could use an ID here, rather than the prompt itself? ie wrap the text of the prompt in a `deterministic_hash()` and just retain the hash, rather than a name. This would allow users to verify that a given prompt is equivalent to the one stored on the Passage.

If you wanted to go all the way down that rabbit hole, maybe we could create a new prompt class, eg:

```py
class Prompt(BaseModel):
    """A prompt for an LLM classifier."""

    text: str = Field(
        description="The text of the prompt",
    )

    @property
    def id(self) -> str:
        """A unique identifier"""
        return deterministic_hash(self.text)
```

going _even further_, maybe a `Prompt` is just another type of `Passage`?

```py

class Passage(BaseModel):
    "It's basically just text, but with some magic helper methods"
    text: str = Field(
        description="Just text :)",
    )

    @property
    def id(self) -> str:
        """A unique identifier"""
        return deterministic_hash(self.text)

    ... etc

class Prompt(Passage):
    """A prompt for an LLM classifier."""
    pass
```

src/data.py
harrisonpim, line 1: i feel like this `data.py` might be too generic a filename for this stuff, and there's a chance it'll get overloaded with weirdness in the near future if we're not careful! maybe it could be `active_learning.py`? or `synthetic_data.py`? or maybe you disagree! what do you reckon?

src/data.py
```diff
@@ -0,0 +1,255 @@
...
+        ActiveLearningData.__init__(self, classifier, upper_bound, lower_bound)
+
+    def find_examples(
+        self, num_samples: int, dataset: Iterable, skip: int = 100
+    ) -> list[PassageWithClassifierConfidence]:
+        """
+        Finds examples on the decision boundary of the classifier
+
+        Takes a dataset (e.g. a stream from `HuggingfaceSession`) and finds example in it, using the
+        `filter_text` method. The `skip` argumante is used to skip every nth passage in the dataset for
```
harrisonpim, line 220:
```suggestion
        `filter_text` method. The `skip` argument is used to skip every nth passage in the dataset for
```

src/data.py
```diff
@@ -0,0 +1,255 @@
...
+        """
+
+        passages = []
+
+        counter = 0
+        for i in dataset:
+            counter += 1
+            if counter % skip == 0:
+                text = i["text_block.text"]
+                if text is not None and len(text) > 50:
```
harrisonpim, line 234: might be worth spelling out in the docstring that we're excluding short passages, and/or making this sort of thing configurable so that the heuristics can be easily tweaked after eg kalyan's chunking work comes through. maybe allow the user to pass in a function to filter passages?

harrisonpim (commented):
i want to spend a bit more time thinking about / discussing / playing with this before it gets merged, but i'm loving this so far! 🕺  awesome work

src/data.py
mpjuhasz: Yes, I think I need some more time to mull over this, and understand what is it really that we want to capture under `prompt`, if anything. I think there are two main options:
- _validation_: i.e. we just want to check that e.g. two passages were created by the same prompt, or validate that a prompt I have to hand was used for this passage. In this case your solution fits perfectly.
- _reproducibility_: this is where I was trying to go, but couldn't get all the way there, as it seemed an overkill. This is where we're saying "I want this passage text to be reproducibly based on the metadata I have attached to it". Obviously this might be unfeasible with the growing number of arguments we use to control generation. 

Now, the hash wouldn't give us number 2, but I'm uncertain whether we do want number 2 at all...

Slightly related note: a few months back I've looked into prompt management services for side-project. It's a gaping hole in the market. There's nothing remotely usable that'd give you easy iteration of prompts (both manually and programmatically), version controlled storage, meta-tags to organise them and multiple prompt versions. I think this is, what's at the bottom of my above rumination.

src/passage.py
mpjuhasz: Yes, that's what I thought when writing this, but it seemed to be quite a large change 🤷

src/data.py
mpjuhasz: I agree! I was wondering whether all this deserves a `data/` folder which could then have `LabelledPassage`, `Passage`, `SyntheticData`, and all other things we'll come up with in the future. What do you reckon, is this a logical arrangement of concepts for you?

src/data.py
mpjuhasz: Good shout, this is a naughty hidden condition here. Tbh, I might as well just remove it (I added it for targets, as my testing was taking too long 😅 )

src/data.py
harrisonpim: maybe they live under `/models`? i would guess that `/data` would be a directory of _actual_ datasets, or a set of utils for loading/saving data... I think we've used `/models` in other projects at cpr for the core structures that a project is built around (?)

src/data.py
harrisonpim: v interesting side note!! wonder whether that service will spring up now that you've mentioned it 😅

src/passage.py
mpjuhasz: I've actually done it now, with `LabelledPassage` inheriting from `Passage`. It involved some wrangling of the validation logic, but I think it's for the better. Will be keen on your input though.

PR #372, opened 2025-04-30
Adding functionality to query vespa passages using continuation tokens

THOR300:
This Pull Request: 
---
- Adds functionality to query vespa passages using continuation tokens. 
- Tested on a large doc via script locally, see [comment on the ticket](https://linear.app/climate-policy-radar/issue/PLA-567/investigate-use-continuation-tokens-to-handle-large-documents#comment-0c110bb0). 

Thoughts: 
- This pull request presents some functions for getting passages and continuation tokens from the vespa query response. 
- The structure / format of the query response a dependent upon the query that is used. 
- Thus these functions are strongly tied to the query so I'm considering whether it would make more sense to tie them together by a class or update the naming so that the relationship can be better inferred. 

Unit test for yielding more results than the max hits: 
- Captured on [this](https://github.com/climatepolicyradar/knowledge-graph/pull/402) stacked PR.


flows/boundary.py
```diff
@@ -545,6 +547,86 @@ def get_document_passage_from_vespa(
...
+
+    continuations = []
+
+    vespa_query_response_root = vespa_query_response.json["root"]
+    group_hits = dig(vespa_query_response_root, "children", 0, "children")
+    for hit in group_hits:
+        hit_continuation_token = dig(hit, "continuation", "next")
+        if hit_continuation_token:
+            continuations.append(hit_continuation_token)
+    return continuations
```
THOR300, line 563: make clear that this is at the child level

flows/boundary.py
```diff
@@ -545,6 +547,86 @@ def get_document_passage_from_vespa(
     return passage_id, passage
 
 
+def get_continuation_tokens_from_query_response(
+    vespa_query_response: VespaQueryResponse,
+) -> list[str]:
+    """Retrieve a continuation token from the response if it exists."""
+
+    continuations = []
```
THOR300, line 555: rename

tests/flows/test_boundary.py
```diff
@@ -1025,3 +1029,59 @@ def test_load_labelled_passages_by_uri_raw(mock_bucket, mock_s3_client):
...
+
+    assert len(response) > 1  # Validate that we did paginate
+    for vespa_passages in vespa_passage_generator:
+        assert isinstance(vespa_passages, list)
+        assert len(vespa_passages) == grouping_max
+        for passage in vespa_passages:
+            assert isinstance(passage[0], str)
+            assert isinstance(passage[1], VespaPassage)
+
+    # TODO: Confirm that we yield all the passages for a family_document_ref
```
THOR300, line 1087: **thought:** Shouldn't the last yielded array of passages most probably not have exactly 10 but less than 10 matches 🤔

flows/boundary.py
```diff
@@ -545,6 +547,112 @@ def get_document_passage_from_vespa(
...
+        #         "coverage": {
+        #             "coverage": 100,
+        #             "documents": 0,
+        #             "full": true,
+        #             "nodes": 1,
+        #             "results": 1,
+        #             "resultsFull": 1
+        #         }
+        #     }
+        # }
```
THOR300, line 644: In this instance we would yield no passages due to the conditional below so maybe we're fine.

flows/boundary.py
```diff
@@ -545,6 +547,112 @@ def get_document_passage_from_vespa(
...
+        for passage in passage_hits
+    ]
+
+    return vespa_passages
+
+
+def get_document_passages_from_vespa__generator(
+    document_import_id: DocumentImportId,
+    vespa_search_adapter: VespaSearchAdapter,
+    continuation_tokens: list[str],
```
THOR300, line 596: Test whether this works when we don't provide an initial continuation token.

flows/boundary.py
```diff
@@ -545,6 +547,93 @@ def get_document_passage_from_vespa(
...
+    grouping_max: int = 10,
+) -> Generator[list[tuple[str, VespaPassage]], None, None]:
+    conditions = qb.QueryField("family_document_ref").contains(
```
olaughter, line 599: I believe you could use `document_import_id` instead and then you'd just need the id value

flows/boundary.py
```diff
@@ -545,6 +547,93 @@ def get_document_passage_from_vespa(
...
+    ]
+
+    return vespa_passages
+
+
+def get_document_passages_from_vespa__generator(
+    document_import_id: DocumentImportId,
+    vespa_search_adapter: VespaSearchAdapter,
+    continuation_tokens: list[str],
+    grouping_max: int = 10,
```
olaughter, line 597: Not sure this needs parameterising?

olaughter: Ah or is it effectively, a page size? Maybe it would be simpler to keep it as a constant within the function, as I can't see it needing to be changed at runtime?

olaughter: Also (assuming I am right that it is page size), probably want to use a much higher value like 500+ or something

flows/boundary.py
```diff
@@ -545,6 +547,93 @@ def get_document_passage_from_vespa(
...
+            raise QueryError(vespa_query_response.get_status_code())
+
+        vespa_passages = get_vespa_passages_from_query_response(vespa_query_response)
+
+        if vespa_passages:
+            yield vespa_passages
+
+        continuation_tokens = get_continuation_tokens_from_query_response(
+            vespa_query_response
+        )
```
olaughter, line 634: I see what you meant in our call about considering a refactor to encapsulate the logic. I wonder if some form of Client object would be useful that encapsulates the query and its response and abstracts the implementation details...

tests/flows/test_boundary.py
```diff
@@ -1025,3 +1029,77 @@ def test_load_labelled_passages_by_uri_raw(mock_bucket, mock_s3_client):
...
+@pytest.mark.vespa
+@pytest.mark.asyncio
+async def test_get_document_passages_from_vespa__generator(
+    document_passages_test_data_file_path: str,
+    local_vespa_search_adapter: VespaSearchAdapter,
+    vespa_app,
+):
+    """Test that we can successfully utilise pagination with continuation tokens."""
+
+    grouping_max = 10
```
olaughter, line 1071: Realising, following on from my comment about this above, that this parameter is probably here to make testing easier?

flows/boundary.py
olaughter: Actually, that is also kind of what the sdk adapter is supposed to be for searches. We are only using the adapter here to access the pyvespa client 🤔

olaughter (approved):
General direction looks good to me! Definitely see your point about a neater version of this functionality. Happy to talk through a bit more how we could make this work. Also happy for that to be ticketed and followed up later

tests/flows/test_boundary.py
```diff
@@ -1025,3 +1029,77 @@ def test_load_labelled_passages_by_uri_raw(mock_bucket, mock_s3_client):
...
+        document_passage_test_data = json.load(f)
+
+    document_passages_count = len(
+        [
+            i["fields"]["family_document_ref"]
+            for i in document_passage_test_data
+            if document_import_id in i["fields"]["family_document_ref"]
+        ]
+    )
+
```
jesse-c, line 1084: **polish:** I'd add a check like this, to ensure this doesn't become a regression:
```suggestion

    assert document_passages_count > boundary.VESPA_MAX_LIMIT, "the fixture has insufficient document passages to validate the test case"

```

flows/boundary.py
```diff
@@ -25,13 +25,15 @@
 from cpr_sdk.s3 import _get_s3_keys_with_prefix, _s3_object_read_text
 from cpr_sdk.search_adaptors import VespaSearchAdapter
 from cpr_sdk.ssm import get_aws_ssm_param
+from cpr_sdk.utils import dig
```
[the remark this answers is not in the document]
olaughter: Oh neat, for the sdk implementation the original inspiration was ruby: https://docs.ruby-lang.org/en/3.1/dig_methods_rdoc.html

flows/boundary.py
THOR300: Yes it's effectively the number of passages to return from the grouping statement! As per your comment below it did make testing easier and is updated to 1000 here on this stacked PR: 
- https://github.com/climatepolicyradar/knowledge-graph/pull/374/files

flows/boundary.py
THOR300: Yeh this is one of those nuanced questions with many solutions, due to the time constraints I'd be happy to ticket and look to chew over it once this is in.

THOR300:
> Seems reasonable so far. I'm unsure how the initial query response is gotten (the one that has a fixture in this PR).
> 
> From the chat this morning, how can we add a test that demonstrates it is indeed working? Fixture(s) for a family document and document passages where there's > 50,000 document passages?
> 
> `CCLW.executive.10014.4470` doesn't have enough documents, according to this:
> 
> ```shell
> $ rg 'family_document_ref": "id:doc_search:family_document::CCLW.executive.10014.4470' tests/local_vespa/test_documents/document_passage.json | wc -l
> 1830
> ```

Good point we should definitely add that to the tests. 

The max hits is a configured value from the vespa application package. One thing we could do is to edit the max hits for the local vespa instance and that way we ensure we're definitely exceeding this! 

https://github.com/climatepolicyradar/knowledge-graph/blob/main/tests/local_vespa/test_app/search/query-profiles/default.xml#L3

tests/flows/test_boundary.py
```diff
@@ -1035,3 +1039,82 @@ def test_load_labelled_passages_by_uri_raw(mock_bucket, mock_s3_client):
...
+        continuation_tokens=["BKAAAAABKBGA"],
+        grouping_max=grouping_max,
+    )
+
+    response = list(vespa_passage_generator)
+
+    assert len(response) > 1  # Validate that we did paginate
+    for vespa_passages in vespa_passage_generator:
+        assert isinstance(vespa_passages, list)
+        assert len(vespa_passages) == grouping_max
```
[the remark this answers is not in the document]
THOR300: Great shout, I'll update as this is brittle! I believe the test is passing because there are `3660` passages for this specific document import id. 

image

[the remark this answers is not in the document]
jesse-c: That explains the multiple of 10, with my count based on a grep search being wrong 🙏🏻

tests/flows/test_boundary.py
THOR300: ### Are we yielding more than the max hits 
---

### Adding a unit test

Currently in development
- Started on [this](https://github.com/climatepolicyradar/knowledge-graph/pull/382/files) PR but loading the local vespa took to long with over 50,000 passages. 
- Now considering [this](https://github.com/climatepolicyradar/knowledge-graph/pull/402/files) approach on this PR (lowering max hits). 

### Proof of over 50k docs yielded
- I have also validated that we are yielding over 50,000 docs from the staging runs. 
- ECS [logs](https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/prefect-mvp-0b5e2a5/log-events$3FfilterPattern$3D$2522No.+of+text+blocks+in+text+blocks+dict$253A+159417$2522$26start$3D-86400000). 
￼
Pasted Graphic 4

￼
### Clarity on the run times of jobs in prefect cloud looking similar:
I think yielding the documents from vespa with the continuation token method is just very fast so it’s a tiny time in relation to the time spinning up an ecs task (when running one document per host), as we can see below 2000 text blocks are yielded from the generator in under a second. 27,000 took 13s which would not be noticeable from the ui 
- ECS [Logs](https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/prefect-mvp-0b5e2a5/log-events/prefect-mvp$252Fprefect$252Fbfb90e5308da470dad36016b8d17f61d$3Fstart$3D1747055709909$26refEventId$3D38960644236548715478176951144223898007899930808436195329). 
￼
Pasted Graphic 6
Pasted Graphic 7

￼

olaughter (approved):

PR #374, opened 2025-05-01
Feature/pla 570 adopt the use of continuation tokens to handle large

THOR300:
This Pull Request: 
---
- Integrates the use of continuation tokens in to our indexing pipeline. 
- Prefect [run](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/811fa339-58b4-4a93-8c3f-ccc100c0f6cc?entity_id=811fa339-58b4-4a93-8c3f-ccc100c0f6cc&state=cancelled&state=cancelling&state=completed&state=crashed&state=failed&state=paused&state=pending&state=running&state=scheduled&entity=flowRuns&entity=taskRuns&entity=events&entity=logs&entity=artifacts).


flows/boundary.py
```diff
@@ -582,20 +582,33 @@ def get_vespa_passages_from_query_response(
...
+            VespaPassage.model_validate(passage["fields"]),
+        )
         for passage in passage_hits
-    ]
+    }
 
     return vespa_passages
 
 
-def get_document_passages_from_vespa__generator(
```
THOR300, line 593: Updated to async to accommodate the use of the vespa connection pool.

flows/boundary.py
```diff
@@ -571,7 +571,7 @@ def get_continuation_tokens_from_query_response(
 
 def get_vespa_passages_from_query_response(
     vespa_query_response: VespaQueryResponse,
-) -> list[tuple[str, VespaPassage]]:
+) -> dict[TextBlockId, tuple[VespaHitId, VespaPassage]]:
```
THOR300, line 574: Updated to match the response type required by the `to_data` function.

flows/boundary.py
```diff
@@ -1182,38 +1195,24 @@ async def run_partial_updates_of_concepts_for_document_passages(
...
+            continuation_tokens=["BKAAAAABKBGA"],
+        )
 
-        grouped_concepts_n = len(grouped_concepts)
-        text_blocks_n = len(text_blocks)
-        if grouped_concepts_n != text_blocks_n:
-            raise ValueError(
-                f"there were {grouped_concepts_n} labelled passages and only "
-                f"{text_blocks_n} document passages were read from Vespa"
-            )
```
THOR300, line 1216: Conditional check removed as we no longer iterate over the `text_blocks` variable which came from grouped concepts on line `1185`. Now we could yield 1000 passages from vespa and only have 5 labelled passages and from my understanding that should be fine 🤔

flows/boundary.py
```diff
@@ -1289,9 +1280,17 @@ class DataPoint(TypedDict):
...
+        text_block_id: TextBlockId, concepts: list[VespaConcept]
+    ) -> DataPoint | None:
+        try:
+            document_passage_id = text_blocks[text_block_id][0]
+            document_passage = text_blocks[text_block_id][1]
+        except KeyError:
+            logger.error(
+                f"document passage `{text_block_id}` not found in Vespa for document `{document_import_id}`"
+            )
+            return None
```
THOR300, line 1293: Catching key errors after seeing the issue in staging runs. 

I believe this is when we have text blocks from the labelled passage array that aren't yielded from vespa.

flows/boundary.py
```diff
@@ -1302,9 +1301,13 @@ def _to_data(text_block_id: TextBlockId, concepts: list[VespaConcept]) -> DataPo
...
-    data: Iterable[dict[str, Any]] = list(
-        map(lambda x: dict(_to_data(*x)), grouped_concepts.items())
-    )
+    logger.info("Beginning creation of DataPoints.")
+    data: Iterable[dict[str, Any]] = []
+    for text_block_id, concepts in grouped_concepts.items():
+        data_point = _to_data(text_block_id, concepts)
+        if data_point:
+            data.append(dict(data_point))
+    logger.info("Finished creation of DataPoints.")
```
THOR300, line 1310: Update required to handle the potential `null` response from `_to_data`

flows/boundary.py
olaughter: Nice! Thanks for doing that

flows/boundary.py
```diff
@@ -1240,40 +1243,28 @@ async def run_partial_updates_of_concepts_for_document_passages(
...
-            # Update the accumulated dictionary with new results
-            next_text_blocks = {
-                text_block.text_block_id: (vespa_hit_id, text_block)
-                for vespa_hit_id, text_block in results
-            }
-            text_blocks = text_blocks | next_text_blocks
+        passages_generator = get_document_passages_from_vespa__generator(
+            document_import_id=document_import_id,
+            vespa_connection_pool=vespa_connection_pool,
+            continuation_tokens=["BKAAAAABKBGA"],
```
olaughter, line 1255: This shouldnt be hardcoded, should it?

tests/flows/test_boundary.py
```diff
@@ -1101,26 +1104,34 @@ async def test_get_document_passages_from_vespa__generator(
...
+
+        assert len(responses) > 1  # Validate that we did paginate
+        for vespa_passages in responses:
+            assert isinstance(vespa_passages, dict)
+            assert 0 <= len(vespa_passages) <= grouping_max
+            for passage in vespa_passages.items():
+                assert isinstance(passage[1][0], str)
+                assert isinstance(passage[1][1], VespaPassage)
+
+        assert (
```
THOR300, line 1128: Remove?

tests/flows/test_boundary.py
```diff
@@ -1101,26 +1104,34 @@ async def test_get_document_passages_from_vespa__generator(
...
+            sum(len(vespa_passages) for vespa_passages in responses)
+            == document_passages_count
+        )
```
THOR300, line 1131: Remove?

flows/boundary.py
olaughter: Probably makes sense to consider this separately to not block the continuation tokens, but I can only think this would happen when we have an issue we want to know about so would prefer to keep the check (and investigate further on possible causes)

flows/boundary.py
THOR300: Ahh yeah I don't know why I thought vespa made you pass in a primer 🙃

flows/boundary.py
THOR300: I've reverted it for this PR, once we get all the changes in and do a run we'll all be on the same page about what the issues are so can design a more appropriate solution 👍

flows/boundary.py
```diff
@@ -597,12 +597,12 @@ def get_continuation_tokens_from_query_response(
         hit_continuation_token = dig(hit, "continuation", "next", default=None)
         if hit_continuation_token:
             continuation_tokens.append(hit_continuation_token)
-    return continuation_tokens
+    return continuation_tokens if continuation_tokens else None
```
kdutia, line 600: **suggestion:** I think these are functionally equivalent?

```suggestion
    return continuation_tokens or None
```

kdutia: Even better, could this just return an empty list instead of `None` when there aren't any continuation tokens? Might mean downstream code is clearer. I can't find a part of the code where you need to know the difference between `None` and `[]`

kdutia (commented):
i don't have enough context for this PR but left a small comment anyway!

THOR300:
So I've added back in the conditional check that confirms that vespa text blocks match labelled passages and we have test failures. Upon investigation I'm unclear on how to solve this! 

From what I can see the issue is we yield 1830 passages from vespa for CCLW.executive.10014.4470 but there is only one grouped concept yielded from the labelled passages file at: 
`('CCLW.executive.10014.4470', 's3://test_bucket/labelled_passages/Q760/v4/CCLW.executive.10014.4470.json')`

image

flows/boundary.py
THOR300: Thanks! 

In the `get_document_passages_from_vespa__generator` function I'm using the conditional `while tokens is not None` to break the loop, so if we returned [] it would just continue.

olaughter (approved):

PR #379, opened 2025-05-06
Bugfix/fix logger instantiation

THOR300:
This Pull Request: 
---
- Is in response to failures seen during an indexing run.
- Issue is from trying to create a prefect logger from a vanilla python function with no flow or task decorator. 
- Errors are not exposed to the prefect console but can be seen in the ecs [logs](https://eu-west-1.console.aws.amazon.com/ecs/v2/clusters/prefect-fargate-cluster-prefect-mvp-staging-e0738bb/tasks/89e5a149250f46248c12d6dfd33a56fc/logs?region=eu-west-1).  
- The fix aims to solve the issue by passing in the logger object.
- This was deemed better than printing the errors (we'd lose the logging level functionality), converting the function to a flow or task (we may hit up against prefect api limits due to this being called many times).

### DIFF
--- a/flows/boundary.py
+++ b/flows/boundary.py
@@ -12,6 +12,7 @@ from collections.abc import Generator
 from datetime import timedelta
 from enum import Enum
 from io import BytesIO
+from logging import Logger
 from pathlib import Path
 from typing import Any, Iterable, Protocol, TypeAlias, TypedDict, TypeVar, Union
 
@@ -29,6 +30,7 @@ from prefect import flow, get_run_logger
 from prefect.client.schemas.objects import FlowRun, StateType
 from prefect.deployments import run_deployment
 from prefect.logging import get_logger
+from prefect.logging.loggers import LoggingAdapter
 from pydantic import BaseModel, NonNegativeInt, PositiveInt
 from vespa.io import VespaQueryResponse, VespaResponse
 from vespa.package import Document, Schema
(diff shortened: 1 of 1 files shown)


flows/boundary.py
```diff
@@ -1308,7 +1311,9 @@ def remove_feed_result_callback(
     grouped_concepts: dict[TextBlockId, list[VespaConcept]],
     response: VespaResponse,
     data_id: VespaDataId,
+    logger: Union[Logger, LoggingAdapter],
 ) -> None:
+    # FIXME: logger
```
kdutia: question: is this FIXME still needed?

flows/boundary.py
```diff
@@ -806,6 +808,7 @@ def __call__(
         grouped_concepts: dict[TextBlockId, list[VespaConcept]],
         response: VespaResponse,
         data_id: VespaDataId,
+        logger: Union[Logger, LoggingAdapter],
```
kdutia, line 811: nitpick (ignore this if you like): is it a good idea to have a function that can _only_ run if you pass in a logger? Could this ever hinder testing or reuse of this function, if reuse was likely?

I think reuse out of this context is unlikely but an alternative here would be to pass an optional logger. This depends on whether you think logging is a core part of the method's reason for being (ie it should **never** not log) I guess...

kdutia (approved):
lgtm! a couple of little, ignorable comments

flows/boundary.py
THOR300: Thanks for the comment! I would rather enforce the logging parameter to avoid conditional paths within the function!

PR #380, opened 2025-05-07
Benchmarking the targets classifiers

mpjuhasz:
## What this PR does
In [PLA-550](https://linear.app/climate-policy-radar/issue/PLA-550/inference-of-classifier-q1653-takes-a-very-long-time) it's been noted, that targets classifiers are 2 magnitudes slower than the rest. This is expected to a degree: the former are neural networks of 350M parameters, whilst the rest mostly keyword based classifiers, which **are** expected to be much faster. 

Regardless, we wanted to do some benchmarking and understanding of whether the Roberta based classifiers can be sped up. To do so, I've run some tests on CPU and GPU, with- and without batching.

|Setup|Time for 1000 paragraphs|
|----|----|
|CPU for loop| 17.80s|
|CPU batches of 10|18.73s|
|CPU batches of 100|19.80s|
|GPU for loop| 4.80s|
|GPU batches of 10|4.46s|
|GPU batches of 100|4.43s|

For reference, the keyword classifier on CPU and GPU for the same dataset took 0.12s and 0.19s respectively.

For the GPU benchmarking I've used a `g5.xlarge` instance, with a single Nvidia A10G GPU and 24GiB GPU memory. Note, that for this to work I've had to load the model to CUDA _after_ it has been initialised (because of how model pickling works), which is non-trivial. It also requires wrangling the dependencies, so that torch compiled with CUDA is available on the system -- I have not included this in the pr, as it'd risk breaking everything else.

> Future work: if we wanted to go down this path, rather than using my batching function, turning the text batches into `Dataset` objects and using built in batching might also speed things up.


src/benchmark_classifiers.py
harrisonpim, line 1: I think scripts like this should probably live in `/scripts` rather than `/src`

harrisonpim (commented):
v cool stuff! just leaving a tiny comment for now, will come back with proper comments when i find time for a more thorough read/think

flows/inference.py
```diff
@@ -345,12 +345,58 @@ def store_labels(
...
+    for batch_idx in range(0, len(all_text), batch_size):
+        text_batch = all_text[batch_idx : batch_idx + batch_size]
+        block_ids = all_block_ids[batch_idx : batch_idx + batch_size]
+
+        outputs.extend(
+            _text_block_inference_for_single_batch(
+                classifier=classifier, text_batch=text_batch, block_ids=block_ids
+            )
+        )
+    return outputs
```
jesse-c, line 366: **polish:** Mostly for my own amusement, I wrote this:
```suggestion
    return [
        _text_block_inference_for_single_batch(classifier=classifier, text_batch=text_batch, block_ids=block_ids)
        for text_batch, block_ids in zip(
            iterate_batch(all_text, batch_size),
            iterate_batch(all_block_ids, batch_size)
        ),
    ]
```

src/benchmark_classifiers.py
```diff
@@ -0,0 +1,101 @@
...
+    end_time = time.time()
+
+    time_taken_Q1653 = end_time - start_time
+    print(f"Time taken for Q1653: {time_taken_Q1653} seconds")
+
+    start_time = time.time()
+    all_text, all_block_ids = [], []
+    for i in list(range(1000)):
+        all_block_ids.append(f"block_{i}")
+        all_text.append("This is a test text for the classifier." * 5)
```
jesse-c, line 48: **question:** I think this setup could be moved before the start time initialisation, so it's not counted towards how long batch prediction took.

jesse-c (approved):

PR #384, opened 2025-05-07
fix: Use non-memory leaking aiohttp

jesse-c:
When building the image, I noticed this warning:

```
12.48 Warning: The file chosen for install of multidict 6.3.2 (multidict-6.3.2-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl) is yanked. Reason for being yanked: Memory leak: https://github.com/aio-libs/multidict/issues/1131#issuecomment-2787514071
```

Poetry also warned about it:

```
Warning: The locked version 6.3.2 for multidict is a yanked version. Reason for being yanked: Memory leak: https://github.com/aio-libs/multidict/issues/1131#issuecomment-2787514071
```

It now is a version that uses a fixed[1][2] version of multidict.

[^1]: https://github.com/aio-libs/multidict/issues/1131#issuecomment-2787514071
[^2]: https://github.com/aio-libs/aiohttp/blob/1acf8c94e7b5ad7d1a2ce6cfec0937abbe43a6eb/requirements/multidict.txt#L5

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -44,92 +44,92 @@ files = [
 
 [[package]]
 name = "aiohttp"
-version = "3.11.16"
+version = "3.11.18"
 description = "Async http client/server framework (asyncio)"
 optional = false
 python-versions = ">=3.9"
 files = [
-    {file = "aiohttp-3.11.16-cp310-cp310-macosx_10_9_universal2.whl", hash = "sha256:fb46bb0f24813e6cede6cc07b1961d4b04f331f7112a23b5e21f567da4ee50aa"},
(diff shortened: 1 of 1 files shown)


poetry.lock
olaughter, line 1: Is this pr missing the pyproject.toml file?

jesse-c: I forgot to update the description, since I removed the change from `pyproject.toml`l and just updated the lock file.

olaughter: Oh I see 👍

olaughter (approved):

PR #393, opened 2025-05-12
Make it easier to audit gaps in inference results

olaughter:
This builds on the existing script that checks the output of inference flows to confirm classifiers have actually run. The original checks for files under the classifier prefix and runs sequentially. This sets up the script to run in parallel, and adds the option to write out the file ids (these can then be used for finding gaps in the results).

As a bonus I also added tests and reworked the input aiming to require less typing.


THOR300 (approved):
LGTM and great that there's even tests for this!

PR #395, opened 2025-05-12
Fix sorting classifier specs

olaughter:
Makes sure that Q123 comes before Q1234 by sorting on the number numerically


scripts/update_classifier_spec.py
```diff
@@ -98,7 +98,8 @@ def model_artifact_is_primary(aliases: list[str], aws_env: AwsEnv) -> bool:
 def sort_specs(specs: list[ClassifierSpec]) -> list[ClassifierSpec]:
     # Have stable ordering. First, the name is gotten from
     # `Q000:v0`, and then the number part is gotten from `Q000`.
-    return sorted(specs, key=lambda x: x.name)
+    # The end result is to sort on 000, so that Q100 is before Q1000
+    return sorted(specs, key=lambda x: int(x.name.removeprefix("Q")))
```
jesse-c, line 102: **fix:** The `name` is just a string. I did once begin to make it a `WikibaseID`, but it wasn't a quick one.
```suggestion
    return sorted(specs, key=lambda x: int(WikibaseID(x.name).numeric()))
```

jesse-c (approved):
I called this a regression, but it isn't! My sorting implementation was better but still wrong. Thanks for fixing this.

tests/scripts/test_update_classifier_spec.py
```diff
@@ -16,6 +14,8 @@
     read_spec_file,
     sort_specs,
 )
+from wandb.apis.public import ArtifactType
```
jesse-c, line 17: **thought:** This annoying thing strikes again!

scripts/update_classifier_spec.py
olaughter: Ah nice thats much neater

harrisonpim: if I remember correctly, `.numeric` is a property, so we don't need those `()` brackets at the end!

olaughter: You are correct, thanks Harrison!

PR #397, opened 2025-05-12
Increase the timeout for inference

olaughter:
This follows hitting a number of timeouts when running inference. I doubt this will be the final value for these, but will make it easier to run larger classifiers in the immediate future.


mpjuhasz (approved):
looks good 🚀

PR #400, opened 2025-05-12
fix: Safely access shared data across Vespa feed callbacks

jesse-c:
Whilst we don't have reproducible steps for the memory issue, and thus I'm not certain if this will fix the issue, it has at least led me to identify this issue.

While list operations are atomic[^1], I have some doubts about relying on the bytecode ordering, and they do suggest to use a mutex if in doubt, so here's a mutex in the form of a lock.

There will be _a_ performance hit, though around that area, the slowest code would be the network call, and updating a list and a counter should be quick.

TOWARS PLA-573

[^1]: https://docs.python.org/3/faq/library.html#what-kinds-of-global-value-mutation-are-thread-safe

### DIFF
--- a/flows/boundary.py
+++ b/flows/boundary.py
@@ -7,6 +7,7 @@ import math
 import os
 import re
 import tempfile
+import threading
 from collections import Counter
 from collections.abc import Generator
 from datetime import timedelta
@@ -1306,18 +1307,29 @@ async def run_partial_updates_of_concepts_for_document_passages(
         map(lambda x: dict(_to_data(*x)), grouped_concepts.items())
     )
 
+    response_cb_lock: threading.Lock = threading.Lock()
+
     # Wrap the callback with the appropriate state and make it match
     # the expected signature.
     def _vespa_response_handler_cb_with_state(
         response: VespaResponse, data_id: VespaDataId
     ):
-        vespa_response_handler_cb(
-            failures,
-            concepts_counts,
-            grouped_concepts,
-            response,
-            data_id,
-        )
+        try:
+            response_cb_lock.acquire(
+                blocking=True,
+                timeout=timedelta(seconds=60).total_seconds(),
+            )
+
+            vespa_response_handler_cb(
+                failures,
+                concepts_counts,
+                grouped_concepts,
+                response,
+                data_id,
+            )
+        finally:
+            if response_cb_lock.locked():
+                response_cb_lock.release()
 
     @vespa_retry()
     def _feed_updates(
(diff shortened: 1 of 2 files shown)


jesse-c:
[Deploy](https://github.com/climatepolicyradar/knowledge-graph/actions/runs/14978098376/job/42075415749) / [Index run](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/0682229d-292b-796a-8000-44450f4b8682?entity_id=068222a3-84bd-7850-8000-39a68cc58044)

flows/boundary.py
```diff
@@ -1156,18 +1157,28 @@ def _to_data(text_block_id: TextBlockId, concepts: list[VespaConcept]) -> DataPo
...
-            failures,
-            concepts_counts,
-            grouped_concepts,
-            response,
-            data_id,
-        )
+        try:
+            response_cb_lock.acquire(
+                blocking=True,
+                timeout=timedelta(seconds=30).total_seconds(),
```
olaughter, line 1170: Thought: I wonder if this is high enough? Maybe 60 is more reasonable?

olaughter (approved):

flows/boundary.py
jesse-c: ✓

flows/boundary.py
```diff
@@ -1195,18 +1196,28 @@ def _to_data(text_block_id: TextBlockId, concepts: list[VespaConcept]) -> DataPo
         map(lambda x: dict(_to_data(*x)), grouped_concepts.items())
     )
 
+    response_cb_lock: threading.Lock = threading.Lock()
+
     # Wrap the callback with the appropriate state and make it match
     # the expected signature.
     def _vespa_response_handler_cb_with_state(
```
THOR300, line 1203: **Thought**: Is it worth adding a test for this function and asserting that we don't throw a run time error should we fail to acquire and then run `release`

https://docs.python.org/3/library/threading.html#lock-objects

> The [release()](https://docs.python.org/3/library/threading.html#threading.Lock.release) method should only be called in the locked state

THOR300 (commented):
LGTM just one comment and suggesting we run on staging as it looks like sandbox didn't yield many if any passages 🤔

flows/boundary.py
jesse-c: I missed that! I'll add a conditional check for `lock.locked()`[^1], and only do it then.

[^1]: Return `True` if the lock is acquired.

PR #402, opened 2025-05-13
Feature: Add a test to ensure we yield more results than vespa max hits  

THOR300:
This Pull Request:
---
- Adds a query profile with lower max hits than the default. 
- This is then used in the query with the vespa generator to ensure we yield beyond the max value. 
- Some time out increases to fix flakey tests.


tests/flows/conftest.py
```diff
@@ -144,6 +144,72 @@ def vespa_app(
...
+
+    subprocess.run(
+        ["just", "vespa_dev_down"],
+        capture_output=True,
+        text=True,
+        check=True,
+        timeout=60,  # Seconds
+    )
+
+    # TODO: Update the default.xml file in the local vespa
```
olaughter, line 172: I agree on the logic of lowering the limit, but think its worth looking for another approach for doing it then running shell commands and editing files as part of tests. Feels flakey and like it could leave the xml in a bad state... might be worth looking at the pyvespa deployments to see if theres at least some python code that could be used for it. If you need to handle files I'd imagine copying the existing directory to a temporary directory, making the change and then deploying from there would be safer then changing it in place

THOR300: Or 

Wondering whether we could append a query profile with lower max hits to the defaut.xml and use this in the query for that one test.  

https://docs.vespa.ai/en/query-profiles.html?mode=cloud#using-a-query-profile

jesse-c (approved):

PR #403, opened 2025-05-13
Feature/pla 592 add retrys

olaughter:
Add two kinds of retries for our indexing flow:
1. Vespa interaction: When communicating with vespa for a total of 3 attempts, 2 second wait between each, and only on specific errors that are likely to be network related (QueryError, VespaError) — note this means not catching `ValueError`, which are less likely to be transient but could also be added in the future if that assumption proves wrong.
2. Prefect flow, document level: Retries the whole document if it meets a failure, currently only tries again two times, and will alert in slack each failure/crash to make sure the issue is surfaced

### DIFF
--- a/flows/boundary.py
+++ b/flows/boundary.py
@@ -45,6 +45,7 @@ from vespa.io import VespaQueryResponse, VespaResponse
 from vespa.package import Document, Schema
 
 from flows.utils import (
+    SlackNotify,
     get_labelled_passage_paths,
     iterate_batch,
     remove_translated_suffix,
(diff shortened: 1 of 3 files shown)


THOR300 (dismissed):
It looks like prefect flows have retry params already! 

image

THOR300 (approved):
This looks great! Do we want to add any test coverage to assert this or just get it in quick, if no test coverage then I would recommend we test on staging! Running on the 100 docs should yield some errors and thus retries.

olaughter:
FWIW, I suspect we will fine tune this with practice too. The specific values are a bit finger in the air at the moment

jesse-c:
Nothing against this solution per se, but curious if you considered https://docs.prefect.io/v3/develop/write-tasks#retries?

PR #406, opened 2025-05-13
Disable timeouts for inference

olaughter:
Hopefully temporarily because i'd like to have a better solution here, but these are currently blocking target inference, and prefect didnt seem to change its config when I increased the value, so they don't seem to behave very well. Even if they did, the documentation states that: "Flow execution may continue until the next task is called" https://docs.prefect.io/v3/develop/write-flows#:~:text=ThreadPoolTaskRunner%20is%20used.-,timeout_seconds,-An%20optional%20number

Meaning they don't do a hard stop currently anyway

### DIFF
--- a/flows/inference.py
+++ b/flows/inference.py
@@ -510,7 +510,7 @@ def iterate_batch(data: list[str], batch_size: int = 400) -> Generator:
         yield data[i : i + batch_size]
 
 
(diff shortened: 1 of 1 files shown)


jesse-c (approved):
Very sad, but makes sense.

I do think we need to abandon relying on Prefect's timeouts mechanisms. 🫠

PR #407, opened 2025-05-13
test: Add very large test for indexing a document

jesse-c:
Huge documents simply are a thing, and thus we should have better testing around them. This also can help with benchmarking. This test whilst not calling the function directly that uses continuation tokens, calls it as part of a higher function.

`tests/local_vespa/test_documents/document_passage_UNFCCC.party.309.0.json` has 50,024 document passages in it. We wanted at least > 50,000. This data was pulled from Vespa (staging).

Doing the document passages feed took 18s once. The test took 48s once. I think this is fine to have run by default on CI.

FIXES PLA-590

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+tests/flows/fixtures/labelled_passages/Q218/v4/UNFCCC.party.309.0_translated_en.json filter=lfs diff=lfs merge=lfs -text
--- a/flows/boundary.py
+++ b/flows/boundary.py
@@ -1362,7 +1362,6 @@ async def run_partial_updates_of_concepts_for_document_passages(
         response: VespaResponse, data_id: VespaDataId
     ):
         try:
-            print("acquiring lock")
             acquired_lock = response_cb_lock.acquire(
                 blocking=True,
                 timeout=timedelta(minutes=2).total_seconds(),
@@ -1372,7 +1371,6 @@ async def run_partial_updates_of_concepts_for_document_passages(
                 print("failed to acquire lock")
                 return
 
-            print("handling response")
             vespa_response_handler_cb(
                 failures,
                 concepts_counts,
@@ -1382,7 +1380,6 @@ async def run_partial_updates_of_concepts_for_document_passages(
             )
         finally:
             locked = response_cb_lock.locked()
-            print(f"releasing lock? {locked}")
             if locked:
                 response_cb_lock.release()
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -91,7 +91,7 @@ env = [
   "WIKIBASE_PASSWORD=test_password",
   "WIKIBASE_URL=https://test.test.test"
 ]
-markers = ["vespa", "flaky_on_ci", "transformers"]
+markers = ["vespa", "flaky_on_ci", "transformers", "slow"]
 asyncio_default_fixture_loop_scope = "function"
 
 [build-system]
--- a/tests/flows/test_index.py
+++ b/tests/flows/test_index.py
@@ -1,7 +1,10 @@
 import json
+import subprocess
+import time
 from collections import Counter
 from datetime import datetime
 from functools import partial
+from io import BytesIO
 from pathlib import Path
 from unittest.mock import patch
 
@@ -33,6 +36,7 @@ from flows.index import (
 from scripts.cloud import ClassifierSpec
 from src.identifiers import WikibaseID
 from src.labelled_passage import LabelledPassage
+from tests.flows.conftest import load_fixture
 
 
 @pytest.mark.asyncio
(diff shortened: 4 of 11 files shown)


tests/flows/test_index.py
```diff
@@ -321,6 +324,123 @@ async def test_run_partial_updates_of_concepts_for_document_passages(
     assert test_counts_serialised == json.loads(result["Body"].read().decode("utf-8"))
 
 
+@pytest.mark.asyncio
+@pytest.mark.vespa
+@pytest.mark.slow
+async def test_run_partial_updates_of_concepts_for_document_passages_with_paging(
```
THOR300, line 330: Question: what's paging referring to here?

tests/flows/test_index.py
```diff
@@ -321,6 +324,123 @@ async def test_run_partial_updates_of_concepts_for_document_passages(
...
+        Bucket=mock_bucket, Key=key, Body=body, ContentType="application/json"
+    )
+
+    # Setup Vespa
+
+    subprocess.run(
+        [
+            "vespa",
+            "feed",
+            "tests/local_vespa/test_documents/family_document_UNFCCC.party.309.0.json",
```
THOR300, line 355: Thought: some of this could go in fixtures for reuse ability.

THOR300 (commented):
Looking good getting this big document in! I'll fully review when in the office!

tests/flows/test_index.py
```diff
@@ -321,6 +324,123 @@ async def test_run_partial_updates_of_concepts_for_document_passages(
...
+            document_import_id=document_import_id,
+            text_blocks_ids=None,
+            vespa_connection_pool=vespa_connection_pool,
+        )
+    updated_concepts = [
+        concept
+        for _, passage in updated_passages
+        if passage.concepts
+        for concept in passage.concepts
+    ]
```
THOR300, line 430: thought: We could add in a comparison of the initial and updated concepts

tests/flows/test_index.py
jesse-c: It was the mixed term we were using about improving reading from Vespa.

tests/flows/test_index.py
jesse-c: Probably![^1]

[^1]: https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

tests/flows/test_index.py
jesse-c: ✓

flows/boundary.py
jesse-c, line 1: **note:** These lines were so noisy.

pyproject.toml
```diff
@@ -91,7 +91,7 @@ env = [
   "WIKIBASE_PASSWORD=test_password",
   "WIKIBASE_URL=https://test.test.test"
 ]
-markers = ["vespa", "flaky_on_ci", "transformers"]
+markers = ["vespa", "flaky_on_ci", "transformers", "slow"]
```
olaughter, line 94: Does the justfile need updating for this too? We have a `test-without-vespa` and it would probably be useful to sometimes run tests without this large one

olaughter (approved):
lgtm!.

tests/flows/test_index.py
```diff
@@ -321,6 +325,132 @@ async def test_run_partial_updates_of_concepts_for_document_passages(
...
+            "tests/local_vespa/test_documents/document_passage_UNFCCC.party.309.0.json",
+        ],
+        capture_output=True,
+        text=True,
+        check=True,
+        timeout=600,  # Seconds
+    )
+    print(
+        f"finished feeding document passages (seconds): {time.perf_counter() - start:.2f}"
+    )
```
THOR300, line 381: **suggestion**: In the `vespa_app` fixture we use for the other tests we run `vespa_delete_data` so that we wipe and reload the vespa db for each test. This might be useful here so we don't have 50k passages for any subsequent tests?

THOR300 (approved):
Very nice to get this large chap into tests!

tests/flows/test_index.py
```diff
@@ -321,6 +325,132 @@ async def test_run_partial_updates_of_concepts_for_document_passages(
...
+        Bucket=mock_bucket, Key=key, Body=body, ContentType="application/json"
+    )
+
+    # Setup Vespa
+    start = time.perf_counter()
+    subprocess.run(
+        [
+            "vespa",
+            "feed",
+            "tests/local_vespa/test_documents/family_document_UNFCCC.party.309.0.json",
```
olaughter, line 356: Just occurred to me that as these are calling shell commands they risk picking up on local config that we don't want to run with, something like this might add some peace of mind
```suggestion
    subprocess.run(
        [
            "vespa",
            "--target",
            "local",
            "feed",
            "tests/local_vespa/test_documents/family_document_UNFCCC.party.309.0.json",
```

.gitattributes
```diff
@@ -0,0 +1 @@
+tests/flows/fixtures/labelled_passages/Q218/v4/UNFCCC.party.309.0_translated_en.json filter=lfs diff=lfs merge=lfs -text
```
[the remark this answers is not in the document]
jesse-c: ✓

tests/flows/test_index.py
jesse-c: I've updated this, and also the ones in the Justfile. They also have the `vespa config set target local`, so it's doubly safe.

tests/flows/test_index.py
jesse-c: I've made it a fixture now though, as it was easier for explicit clean-up.

PR #408, opened 2025-05-14
Feature/handle query returning all passages

THOR300:
This Pull Request: 
---
- Currently the indexing pipeline on main looks to only yield passages from vespa that exist in the labelled passages data from s3 by explicitly passing in text block ids into the query before using the `equiv` operator. 
- Now with continuation tokens we look to yield all passages for a document import id from vespa and thus this conditional check no longer applies.
- We do however still log a warning as we would really hope they match (as we create labelled passages from parser outputs in the pipeline cache). 
- We then update the `_to_data` function to handle the scenario where we have a text block from the labelled passages but not in the vespa passage query results.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,12 +2,12 @@ name: CI
 
 on:
   pull_request:
-    # TEMP
-    # branches:
-    #   - main
+    branches:
+      - main
   push:
     branches:
       - main
+  workflow_dispatch:
 
 permissions:
   contents: write
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,8 +2,9 @@ name: CI
 
 on:
   pull_request:
-    branches:
-      - main
+    # TEMP
+    # branches:
+    #   - main
   push:
     branches:
       - main
--- a/flows/boundary.py
+++ b/flows/boundary.py
@@ -1271,9 +1271,9 @@ async def run_partial_updates_of_concepts_for_document_passages(
         text_blocks_n = len(text_blocks)
         if grouped_concepts_n != text_blocks_n:
             logger.warning(
-                f"There were {grouped_concepts_n} labelled passages and only "
-                f"{text_blocks_n} document passages were read from Vespa for "
-                f"{document_import_id}"
+                f"There were {grouped_concepts_n} text block ids from the labelled "
+                f"passages and only {text_blocks_n} document passages were read from "
+                f"Vespa for {document_import_id}"
             )
 
     # Batch updates (writes)
--- a/flows/boundary.py
+++ b/flows/boundary.py
@@ -1272,7 +1272,7 @@ async def run_partial_updates_of_concepts_for_document_passages(
         if grouped_concepts_n != text_blocks_n:
             logger.warning(
                 f"There were {grouped_concepts_n} text block ids from the labelled "
-                f"passages and only {text_blocks_n} document passages were read from "
+                f"passages but {text_blocks_n} document passages were read from "
                 f"Vespa for {document_import_id}"
             )
 
--- a/flows/boundary.py
+++ b/flows/boundary.py
@@ -1270,9 +1270,10 @@ async def run_partial_updates_of_concepts_for_document_passages(
         grouped_concepts_n = len(grouped_concepts)
         text_blocks_n = len(text_blocks)
         if grouped_concepts_n != text_blocks_n:
-            raise ValueError(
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -2,8 +2,9 @@ name: CI
 
 on:
   pull_request:
-    branches:
-      - main
+    # TEMP
+    # branches:
+    #   - main
```
THOR300, line 7: Do we have any concerns with making this permanent as we are using stacked PR's a lot now it's good to have tests run not only against main.

flows/boundary.py
```diff
@@ -1270,9 +1270,10 @@ async def run_partial_updates_of_concepts_for_document_passages(
         grouped_concepts_n = len(grouped_concepts)
         text_blocks_n = len(text_blocks)
         if grouped_concepts_n != text_blocks_n:
-            raise ValueError(
-                f"there were {grouped_concepts_n} labelled passages and only "
-                f"{text_blocks_n} document passages were read from Vespa"
+            logger.warning(
+                f"There were {grouped_concepts_n} labelled passages and only "
```
THOR300, line 1274: ids from....

.github/workflows/ci.yml
olaughter: We could add a manual "workflow_dispatch" option if we want something permanent

flows/boundary.py
```diff
@@ -1270,9 +1270,10 @@ async def run_partial_updates_of_concepts_for_document_passages(
         grouped_concepts_n = len(grouped_concepts)
         text_blocks_n = len(text_blocks)
         if grouped_concepts_n != text_blocks_n:
-            raise ValueError(
```
THOR300, line 1273: Add a. version bump in the pyproject.toml

flows/boundary.py
olaughter, line 1274:
```suggestion
                f"There were {grouped_concepts_n} labelled passages but "
```

olaughter (approved):

.github/workflows/ci.yml
THOR300: I've added in the workflow dispatch option.

flows/boundary.py
```diff
@@ -1304,9 +1313,11 @@ def _to_data(text_block_id: TextBlockId, concepts: list[VespaConcept]) -> DataPo
...
 
     logger.info("Beginning creation of DataPoints.")
-    data: Iterable[dict[str, Any]] = list(
-        map(lambda x: dict(_to_data(*x)), grouped_concepts.items())
-    )
+    data: Iterable[dict[str, Any]] = []
+    for text_block_id, concepts in grouped_concepts.items():
+        data_point = _to_data(text_block_id, concepts)
+        if data_point:
+            data.append(dict(data_point))
```
jesse-c, line 1320: **quibble:** 

```python
data: Iterable[dict[str, Any]] = [
    dict(result) for x in grouped_concepts.items() 
    if (result := _to_data(*x)) is not None
]
```

Or:

```python
data: Iterable[dict[str, Any]] = list(
    filter(None, map(lambda x: _to_data(*x)), grouped_concepts.items()))
)
```

I haven't tested these, just wrote them down.

jesse-c: This isn't needed, just a joke about losing the lambda. ;)

flows/boundary.py
```diff
@@ -1270,9 +1270,10 @@ async def run_partial_updates_of_concepts_for_document_passages(
         grouped_concepts_n = len(grouped_concepts)
         text_blocks_n = len(text_blocks)
         if grouped_concepts_n != text_blocks_n:
-            raise ValueError(
-                f"there were {grouped_concepts_n} labelled passages and only "
-                f"{text_blocks_n} document passages were read from Vespa"
+            logger.warning(
```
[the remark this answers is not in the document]
THOR300: Currently retention is one week: 

I can make an update to the orchestrator repo to up this to one month as a follow up? 

image

flows/boundary.py
THOR300: Yeah I was sad to see it go haha!

flows/boundary.py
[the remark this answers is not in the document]
THOR300: Going to merge, we can continue this convo!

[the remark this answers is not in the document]
THOR300: Pr to update retention period: 

https://github.com/climatepolicyradar/orchestrator/pull/198

PR #409, opened 2025-05-14
Fix logging issues in indexing

olaughter:
See individual commits about the ways this resolves logging issues. One of which led to an actual failure:

- Fix logging around the `s3_accessor`
    
    I was curious why "s3_prefixes" didnt appear in the logs, and after
    taking a closer look at what that log line was doing realised theres a
    very good chance its failing to log out a gigantic log object. This
    would also make sense of why the open telemetry failure occured here as
    well as the underlying state update error due to request size
    
    I also moved the logic for what to log out of the main function for
    neatness reasons

- Log print statements
    
    Prefect was missing log statements that where found in ecs. This is due
    to the function `s3_paths_or_s3_prefixes` having print statements rather
    then use logging. Elsewhere we have adopted print statements for prefect
    flows because they are cleaner and seem to behave better with concurrency
    so I have switched the main flow here to do the same. This will then
    allow the s3 functions print statements to be logged

### DIFF
--- a/flows/boundary.py
+++ b/flows/boundary.py
@@ -111,6 +111,16 @@ class S3Accessor(BaseModel):
     paths: list[str] | None = None
     prefixes: list[str] | None = None
 
+    def __str__(self) -> str:
+        """String representation of the S3Accessor for logging"""
+        prefix_count = len(self.prefixes) if self.prefixes else 0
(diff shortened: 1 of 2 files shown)


THOR300 (approved):
That's a great shout!

PR #413, opened 2025-05-14
Make check specific for actual issues

olaughter:
We found examples of our query returning too few passages. After some investigation it turned out to be normal for inference to have less, but the other way around is a genuine problem that we want to surface right away

### DIFF
--- a/flows/boundary.py
+++ b/flows/boundary.py
@@ -1296,8 +1296,11 @@ async def run_partial_updates_of_concepts_for_document_passages(
             f"Memory size (Mb): {text_blocks_size_in_mb}. "
         )
 
-        if grouped_concepts_n != text_blocks_n:
-            logger.warning(
+        # It's normal for there to be less text blocks from inference than vespa
+        # because inference filters out more types. However it should never be
+        # The other way around
+        if grouped_concepts_n > text_blocks_n:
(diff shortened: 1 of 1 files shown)


flows/boundary.py
```diff
@@ -1296,8 +1296,11 @@ async def run_partial_updates_of_concepts_for_document_passages(
             f"Memory size (Mb): {text_blocks_size_in_mb}. "
         )
 
-        if grouped_concepts_n != text_blocks_n:
-            logger.warning(
+        # It's normal for there to be less text blocks from inference than vespa
+        # because inference filters out more types. However it should never be
+        # The other way around
+        if grouped_concepts_n > text_blocks_n:
+            raise ValueError(
```
THOR300, line 1303: At this point in the project if the query returns 99,000 passages out of 100,000 do we not want to update concepts for the 99,000 we do have rather than fail here?

THOR300 (approved):
LGTM

flows/boundary.py
jesse-c: It probably doesn't matter, unless we're able to re-run for just the 1,000. If we can't, then we'd have to run it on all of them anyway.

jesse-c (approved):

PR #414, opened 2025-05-14
Add precision to grouped query

olaughter:
Because of the way vespa works with querying we where losing passages in the response. This is because by default the way vespa handles searches across nodes and pools the results is optimised for performance not accuracy.

In our case here, we kinda really need accuracy and this can be changed by overriding the default "precision" parameter in the grouping query.

See:
- https://docs.vespa.ai/en/reference/grouping-syntax.html#precision
- https://docs.vespa.ai/en/grouping.html#performance-and-correctness

### DIFF
--- a/flows/boundary.py
+++ b/flows/boundary.py
@@ -648,7 +648,8 @@ async def get_document_passages_from_vespa__generator(
     document_import_id: DocumentImportId,
     vespa_connection_pool: VespaAsync,
     continuation_tokens: list[ContinuationToken] | None = [],
-    grouping_max: int = 1000,
+    grouping_max: int = 5_000,
+    grouping_precision: int = 100_000,
     query_profile: str = "default",
 ) -> AsyncGenerator[dict[TextBlockId, tuple[VespaHitId, VespaPassage]], None]:
     """
@@ -676,6 +677,7 @@ async def get_document_passages_from_vespa__generator(
     grouping = G.all(
         G.group("text_block_id"),
         G.max(grouping_max),
+        G.precision(grouping_precision),
         G.each(G.each(G.output(G.summary()))),
     )
 


THOR300 (approved):
LGTM - fingers crossed this works!

flows/boundary.py
```diff
@@ -648,7 +648,8 @@ async def get_document_passages_from_vespa__generator(
     document_import_id: DocumentImportId,
     vespa_connection_pool: VespaAsync,
     continuation_tokens: list[ContinuationToken] | None = [],
-    grouping_max: int = 1000,
+    grouping_max: int = 5_000,
```
jesse-c, line 651: **suggestion:** Where did these numbers come from? I glanced over the 2 links to the Vespa docs and didn't immediately see anything.

jesse-c (approved):
> A higher number costs more bandwidth ..

I'm worried about the bandwidth comment here, so we may need to fine-tune these numbers.

> .. but leads to higher accuracy in some cases.

Is our case one of them?

> Perfect correctness may not be achievable; result of queries using [non-default ordering](https://docs.vespa.ai/en/grouping.html?mode=cloud#ordering-and-limiting-groups) can be approximate, and correctness can only be partially achieved by a larger precision value that sacrifices performance.

Does this not mean that we can't expect 100% document passages being returned?

This may be the final nail in the coffin for me that we have no choice but to not change our design to not read from Vespa.

>  Important: The grouping session cache is only useful if the grouping expression uses default ordering. The drawback is that when max is specified in the grouping expression, it might cause inaccuracies in aggregated values such as count. It is recommended testing whether this is an issue or not, and if so, adjust the precision parameter to still get correct counts.

.. 

> The [grouping session cache](https://docs.vespa.ai/en/reference/grouping-syntax.html#grouping-session-cache) is enabled by default. Disabling it will improve correctness, especially for queries using order and max. The cost of multi-level grouping expressions will increase, though.

Shall we disable the cache?

We'd hit "The cost of multi-level grouping expressions will increase, though" though, since aren't we doing that now?

It may be helpful to start putting more of these numbers as unable parameters in the Deployment config, so we don't need a whole PR-loop to change them.

olaughter:
> > A higher number costs more bandwidth ..
> 
> I'm worried about the bandwidth comment here, so we may need to fine-tune these numbers.

Me too! I think this adds to the case that we don't want to be relying on vespa for this long term and should try to move to something more of a push down from s3 kind of approach as we've raised previously


> > .. but leads to higher accuracy in some cases.
> 
> Is our case one of them?

It is. I did some tests running `get_document_passages_from_vespa__generator` with different values against `CCLW.executive.1317.2153`, which was one of the problem docs. I tried a whole bunch of combinations, and for that specific doc, saw changes in reliability from both adjusting the group as well as the precision.

Bumping the group to 1850 worked reliably, even without changing the precision, which was interesting in that the doc has 18,000 passages. and the documentation implies a relationship between these numbers (even if its super vague!). I also tried a group size of 10 and tweaking the precision up to 10_000 where it ran reliabely (and slow).

> ...
> Does this not mean that we can't expect 100% document passages being returned?
> 
> This may be the final nail in the coffin for me that we have no choice but to not change our design to not read from Vespa.

Yep yep yep yeeep! 

> ...
> Shall we disable the cache?

I actually found doing so didnt change anything! Might be worth a revisit at some point though
 
> It may be helpful to start putting more of these numbers as unable parameters in the Deployment config, so we don't need a whole PR-loop to change them.

Yeah, I was thinking a dedicated query profile dedicated to indexing would be a good way of avoiding this from changing the defaults for search behaviour too! Probably worth considering alongside the not-reading-from-vespa planning to decide if its still worth having

jesse-c:
All sounds good to me! Thanks for the thorough testing.

PR #416, opened 2025-05-15
Confirm write out path for concept counts

olaughter:
Adds logging that will print the path for the concept_counts summary so we can more easily go and confirm its being written correctly


THOR300 (approved):
LGTM

PR #417, opened 2025-05-15
Remove invalid classifier

olaughter:
Greenhouse gases is failing validations which is causing problems in the indexing pipeline:

```
Concept
  Value error, A negative label should not be the same as a positive label
```


jesse-c (approved):
Did the scripts work as expected?

olaughter:
> Did the scripts work as expected?

They did! 👍

PR #418, opened 2025-05-19
ci: Check if the `workflow_run` completion was a success

jesse-c:
Deploying should either be automatic (via the CI work flow completing) or via a manual trigger. For the former, we had the trigger, but we weren't checking if the completed workflow was a success completion, as opposed to something else like a failure completion.


olaughter (approved):
Nice catch

PR #412, opened 2025-05-14
refactor: Explicitly control updating document passages

jesse-c:
This takes us back to a time before we [tried to use](https://github.com/climatepolicyradar/knowledge-graph/pull/368) `pyvespa` for working through all our updates operations.

It was way too far to do a simple revert, so I've written this based on what was before that PR, and where we've gotten to.

Effectively, no tests were needed to be updated. This is a good thing! It indicates that the functionality/effects weren't changed.

FIXES PLA-603

PR #421, opened 2025-05-22
refactor: Allow flexible async coroutines

jesse-c:
A batch of tasks means that the batch has to wait until all tasks finish, before starting the next batch. That means the coroutines are _inflexible_.

Instead, use a queue AKA lock AKA semaphore, so that the concurrency is managed by there being _space_ in the semaphore. E.g. a semaphore of 8 means that there's 8 concurrent tasks. A task blocks until it acquires a lock, from the semaphore.

TOWARDS PLA-615

### DIFF
--- a/flows/boundary.py
+++ b/flows/boundary.py
@@ -75,7 +75,7 @@ T = TypeVar("T")
 CONCEPT_COUNT_SEPARATOR: Final[str] = ":"
 CONCEPTS_COUNTS_PREFIX_DEFAULT: Final[str] = "concepts_counts"
 DEFAULT_DOCUMENTS_BATCH_SIZE: Final[PositiveInt] = 50
-DEFAULT_TEXT_BLOCKS_BATCH_SIZE: Final[PositiveInt] = 50
+DEFAULT_TEXT_BLOCKS_BATCH_SIZE: Final[PositiveInt] = 20
 DEFAULT_UPDATES_TASK_BATCH_SIZE: Final[PositiveInt] = 5
 
 # Get more logs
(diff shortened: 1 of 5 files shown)


jesse-c:

- #412
    - #421 ◀

flows/boundary.py
```diff
@@ -1369,46 +1371,43 @@ async def run_partial_updates_of_concepts_for_document_passages(
...
                 )
+            )
 
-            logger.info(f"Gathering updates tasks for batch #{i}")
-            tasks_results = await asyncio.gather(*tasks, return_exceptions=True)
-            responses.extend(tasks_results)
-            logger.info(f"Gathered updates tasks for batch #{i}")
+        start = time.perf_counter()
+        logger.info("starting gathering updates tasks")
+        tasks_results = await asyncio.gather(*tasks, return_exceptions=True)
```
olaughter, line 1406: Ahhh, I think `return_exceptions` is what I probably need for the aggregator! Also the semaphore logic as well. Will pinch once the dust settles on this one

flows/boundary.py
```diff
@@ -1369,46 +1371,43 @@ async def run_partial_updates_of_concepts_for_document_passages(
...
+        responses.extend(tasks_results)
+        logger.info(
+            f"finished gathering updates tasks in {time.perf_counter() - start:.2f}s"
```
olaughter, line 1409: This is nice to have! I made [a simple context manager implementation](https://github.com/climatepolicyradar/knowledge-graph/blob/main/scripts/do_outputs_align_for_a_document.py#L41-L61) as part of an audit script for this simple kind of profiling too. If we want to do this more often, we could also make that more resuable

flows/boundary.py
jesse-c: I was going to comment on that, but saw the try/catch with a return, and wasn't sure if it was deliberate, haha.

PR #422, opened 2025-05-22
Filtering out sabin documents from inference.

THOR300:
This Pull Request: 
---
- Filters out sabin documents from inference only (if they aren't in inference outputs they won't be presented to indexing). 
- Using the file stem to filter out the documents and ensuring we aren't case sensitive so we're resilient to minor upstream naming changes. 
- The alternative option I could see to using the file stem was to use the `source` field in the `document_metadata` of the parser outputs. However this would require reading the file from s3 and the document metadata field is just typed as a dict so this method was not deemed ideal. See example below: 

image

### DIFF
--- a/flows/inference.py
+++ b/flows/inference.py
@@ -12,7 +12,6 @@ from uuid import UUID
 
 import boto3
 import prefect.artifacts as artifacts
-import wandb
 from cpr_sdk.parser_models import BaseParserOutput, BlockType
 from cpr_sdk.ssm import get_aws_ssm_param
 from prefect import flow
@@ -22,8 +21,8 @@ from prefect.deployments import run_deployment
 from prefect.logging import get_run_logger
 from prefect.task_runners import ConcurrentTaskRunner
 from pydantic import SecretStr
-from wandb.sdk.wandb_run import Run
 
+import wandb
 from flows.utils import (
     SlackNotify,
     filter_non_english_language_file_stems,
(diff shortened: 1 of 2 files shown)


tests/flows/test_inference.py
```diff
@@ -374,3 +376,35 @@ async def test_run_classifier_inference_on_document_missing(
...
+                "Sabin.document.16944.17490",
+                "UNFCCC.document.1234.5678",
+                "Sabin.document.16945.17491",
+            ],
+            ["CCLW.executive.12345.6789", "UNFCCC.document.1234.5678"],
+        ),
+        (["sabin.document.16944.17490", "SABIN.document.16945.17491"], []),
+        (
+            ["SabinIndustries.document.1234.5678", "DocumentSabin.12345.6789"],
+            ["SabinIndustries.document.1234.5678", "DocumentSabin.12345.6789"],
```
THOR300, line 402: Note this example that with the current implementation we would not filter out `DocumentSabin`; happy to update the conditional in the list comprehension.

flows/inference.py
```diff
@@ -227,6 +227,16 @@ def determine_file_stems(
...
 
 
+def remove_sabin_file_stems(file_stems: list[DocumentStem]) -> list[DocumentStem]:
+    """
+    Remove Sabin document file stems from the list of file stems.
+
+    File stems of the Sabin source follow the below naming convention:
+    - "Sabin.document.16944.17490"
+    """
+    return [stem for stem in file_stems if stem.split(".")[0].lower() != "sabin"]
```
olaughter, line 237: Simpler?
```suggestion
    return [stem for stem in file_stems if not stem.startswith("Sabin")]
```

olaughter (approved):
Looks good! Thanks for adding such a comprehensive test as well

jesse-c (commented):
**thought:** I wonder how long this would take if we ever have to run inference on every document.

PR #423, opened 2025-05-22
fetch concepts in batches, without cycles

harrisonpim:
This PR significantly speeds up the fetching of concepts from wikibase when recursive properties are invoked.

- The bulk of the logic is moved from `get_concept` to `get_concepts`, where concepts are now fetched in batches.
- `get_concept` is now a wrapper around `get_concepts` where only one concept is fetched, with the options to include extra labels and relationships
- When fetching properties recursively, a `visited` set is used to prevent cycles and reduce the number of redundant calls
- Brought in a bit more consistency about where errors are raised


olaughter (approved):
Nice

PR #424, opened 2025-05-22
update specs after retraining with include_recursive_has_subconcept

harrisonpim:
follows #420 and #423


flows/classifier_specs/sandbox.yaml
```diff
@@ -4,7 +4,6 @@
 - Q404:v8
 - Q412:v13
 - Q757:v8
-- Q760:v9
```
jesse-c, line 7: **suggestion:** You may want to revert this one. From memory, it's a special value for something!

harrisonpim: done!

PR #425, opened 2025-05-22
Add initial inference results aggregation flow 

olaughter:
This takes the raw inference output, aka "laballed_passages", it groups them by passage and writes the results as `VespaConcept` objects back to s3 under "inference_results". These are what the passage indexing uses, so we can also shift some code out of the indexing and now better interrogate the data before it goes into vespa.

There's more that can be done for speed on this flow, but I'd like to take it for some spins without concurrency before adding any.

We will also need to add a sensible default for not specifying document ids at some point, but again felt we could air it out a bit before figuring that out.


flows/aggregate_inference_results.py
```diff
@@ -0,0 +1,165 @@
...
+    return combined_passages
+
+
+@flow(
+    on_failure=[SlackNotify.message],
+    on_crashed=[SlackNotify.message],
+    timeout_seconds=None,
+    log_prints=True,
+)
+def aggregate_inference_results(document_ids: list[str], config: Config | None = None):
```
THOR300, line 142: **praise**: Nice to see this not being async for once

THOR300, line 142: **question**: This flow enforces invocation with a list of known ids are we doing this just because we're gonna be running on the staging sample set for now or because we will look to invoke this flow with doc ids from a parent flow?

flows/aggregate_inference_results.py
```diff
@@ -0,0 +1,165 @@
...
+)
+from scripts.cloud import (
+    AwsEnv,
+    ClassifierSpec,
+    get_prefect_job_variable,
+)
+from scripts.update_classifier_spec import parse_spec_file
+from src.labelled_passage import LabelledPassage
+
+INFERENCE_RESULTS_PREFIX = "inference_results"
```
THOR300, line 27: **thought**: Add to the config object?

flows/aggregate_inference_results.py
```diff
@@ -0,0 +1,165 @@
...
+def build_run_output_prefix() -> str:
+    """Get the name of the flow run."""
+    run_context = get_run_context()
+    if not run_context:
+        raise MissingContextError()
+    start_time = run_context.flow_run.start_time.replace(tzinfo=None).isoformat(
+        timespec="minutes"
+    )
+    run_name = run_context.flow_run.name
+    return f"{INFERENCE_RESULTS_PREFIX}/{start_time}-{run_name}"
```
THOR300, line 61: praise: A sign of getting closer to idempotency.

THOR300 (approved):
Great to turn this around so fast.

flows/aggregate_inference_results.py
```diff
@@ -0,0 +1,165 @@
...
+        f"{len(classifier_specs)} classifiers, outputting to {run_output_prefix}"
+    )
+    for document_id in document_ids:
+        all_labelled_passages = get_all_labelled_passages_for_one_document(
+            document_id, classifier_specs, config
+        )
+        vespa_concepts = combine_labelled_passages(all_labelled_passages)
+
+        # Write to s3
+        s3_uri = f"s3://{config.cache_bucket}/{run_output_prefix}/{document_id}.json"
```
THOR300, line 162: **thought**: os.path.join here?

flows/aggregate_inference_results.py
```diff
@@ -0,0 +1,165 @@
...
+            )
+
+
+def combine_labelled_passages(
+    labelled_passages: dict[str, list[LabelledPassage]],
+) -> dict[str, list[dict[str, str]]]:
+    """Combine the labelled passages across the different classifier specs."""
+    labelled_passages_lists = list(labelled_passages.values())
+    if not check_all_values_are_the_same([len(lpl) for lpl in labelled_passages_lists]):
+        raise ValueError(
```
THOR300, line 113: **thought**: Because this isn't caught in the top level for loop we would stop the entire process if one document had a mismatch, might be better to identify and continue?

flows/aggregate_inference_results.py
olaughter: Yep, just for now. I mentioned this in the description that something smarter is still outstanding for this. We'll still need it even if we have a parent flow because the max context size will block us from passing in documents as parameters after a certain amount

flows/aggregate_inference_results.py
olaughter: Good call

flows/aggregate_inference_results.py
olaughter: Lets gooo

flows/aggregate_inference_results.py
olaughter: See how long it lasts 😅

flows/aggregate_inference_results.py
olaughter: The util function being used to write expects the s3 protocol prefix making it hard to do clearly with `os.path.join`. At some point we could simplify the util function though...

flows/aggregate_inference_results.py
olaughter: Really good idea, I think because this flow is still self contained its safe to come back and add that as its own pr'd change so will leave it from here now and follow up with it afterwards

tests/flows/conftest.py
```diff
@@ -38,6 +41,13 @@
 FIXTURE_DIR = Path(__file__).resolve().parent / "fixtures"
 
 
+@pytest.fixture(autouse=True, scope="session")
+def prefect_test_fixture():
+    with prefect_test_harness(server_startup_timeout=120):
+        with disable_run_logger():
+            yield
+
+
```
jesse-c, line 50: **fix:** This is a duplicate of https://github.com/climatepolicyradar/knowledge-graph/blob/refactor%2Fdont-bulk-update/tests/conftest.py#L20-L24, and should be removed?

tests/flows/test_aggregate_inference_results.py
```diff
@@ -0,0 +1,244 @@
...
+        "UNFCCC.party.492.0",
+    ]
+
+    with tempfile.TemporaryDirectory() as spec_dir:
+        # Write the concept specs to a YAML file
+        temp_spec_dir = Path(spec_dir)
+        concept_specs = ["Q123:v4", "Q223:v3", "Q218:v5", "Q767:v3", "Q1286:v3"]
+        spec_file = temp_spec_dir / "sandbox.yaml"
+        with open(spec_file, "w") as f:
+            yaml.dump(concept_specs, f)
```
jesse-c, line 44: **fix:** This should use `write_spec_file` so that we can't drift in how we write these files.

jesse-c (commented):
LGTM though there are some things we should fix.

Having this as a separate flow/deployment does go against the monolithic-pipeline-first pattern.

PR #426, opened 2025-05-25
Refactor aggregator for speed, readability and robustness

olaughter:
- Adds basic concurrency
- Reduces the amount of strings being passed around
- Improves type hinting
- Simplifies some duplicate & overly verbose code
- Collects failures


flows/aggregate_inference_results.py
```diff
@@ -138,37 +179,93 @@ def combine_labelled_passages(
...
+            else:
+                failures.append(
+                    DocumentFailure(document_id=document_id, exception=error)
+                )
+
+    # Results
+    print(f"Successes: {len(successes)}/{len(document_ids)}, failures: {len(failures)}")
+
+    if failures:
+        print(f"Failures: {failures}")
```
olaughter, line 269: Would quite like to switch this to an artifact but for whatever reason I couldnt get it to actually create the artifact when I ran it so I've removed for now

flows/aggregate_inference_results.py
```diff
@@ -138,37 +179,93 @@ def combine_labelled_passages(
...
+            key=os.path.join(
+                config.aggregate_inference_results_prefix,
+                run_output_identifier,
+                f"{document_id}.json",
+            ),
+        )
+        s3_object_write_text(str(s3_uri), json.dumps(vespa_concepts))
+        return document_id, None
+    except Exception as e:
+        return document_id, e
```
olaughter, line 208: This probably needs a rethink as it leaves the task state as Succeeded. Potentially revisiting and borrowing from the approach in inference / using a Result object

flows/aggregate_inference_results.py
```diff
@@ -138,37 +179,98 @@ def combine_labelled_passages(
...
     if not config:
         print("no config provided, creating one")
-        config = Config()
+        config = await Config.create()
 
-    run_output_prefix = build_run_output_prefix(
-        config.aggregate_inference_results_prefix
-    )
+    if not document_ids:
+        raise ValueError("No document ids provided, fallback is not supported yet")
```
jesse-c, line 229: **quibble:** I think a more _direct_ exception could make it slightly clearer.
```suggestion
        raise NotImplementedError("No document ids provided, fallback is not supported yet")
```

flows/aggregate_inference_results.py
```diff
@@ -138,37 +179,93 @@ def combine_labelled_passages(
...
+
```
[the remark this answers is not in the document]
jesse-c: Which state size? For passing between flows/tasks/deployments in Prefect?

flows/aggregate_inference_results.py
```diff
@@ -138,37 +179,98 @@ def combine_labelled_passages(
...
-        vespa_concepts = combine_labelled_passages(all_labelled_passages)
+        for document_id in document_ids
+    ]
 
-        # Write to s3
-        s3_uri = f"s3://{config.cache_bucket}/{run_output_prefix}/{document_id}.json"
-        s3_object_write_text(s3_uri, json.dumps(vespa_concepts))
+    # Process documents in batches to control concurrency
+    failures: list[DocumentFailure] = []
+    successes: list[DocumentImportId] = []
```
jesse-c, line 252: **thought:** We're definitely passing the rule of three on this pattern. It'll be nice once we're confident enough in this and have it encoded in code as a reusable thing!

flows/aggregate_inference_results.py
jesse-c: A CSV-kind of artifact could be so good for easy reporting! Another possible orchestration pattern for s.

scripts/update_classifier_spec.py
```diff
@@ -50,7 +50,7 @@ def parse_spec_file(aws_env: AwsEnv) -> list[ClassifierSpec]:
 
 def write_spec_file(file_path: Path, data: list[ClassifierSpec]):
     """Save a classifier spec YAML"""
-    serialised_data = list(map(lambda spec: f"{spec.name}:{spec.alias}", data))
+    serialised_data = list(map(lambda spec: str(spec), data))
```
jesse-c, line 53: **praise:** Nice! When I made that suggestion on the other PR, I assumed this was already here 😅

jesse-c (approved):
LGTM, and I'm curious if we need the start time in the run name.

flows/aggregate_inference_results.py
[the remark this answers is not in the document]
olaughter: It's true for all of these I think, though in this case its the input to the flow itself thats the problem, there is a 512kb limit on what you can pass in, which is a lot but if we are passing in some 15k document ids it adds up

flows/aggregate_inference_results.py
olaughter: Thats a good idea, I had similarly aimed for a table artifact initially though I like the idea of a csv because it can then be picked up from other places which is nice... hoping to revisit this at some point

flows/aggregate_inference_results.py
olaughter: It's not far off, hoping to come back and make it better

flows/aggregate_inference_results.py
```diff
@@ -28,8 +30,42 @@
...
+RunOutputIdentifier: TypeAlias = str
+
+# A string representation of a classifier spec (i.e. Q123:v4)
+SpecStr: TypeAlias = str
+
+# A serialised vespa concept, see cpr_sdk.models.search.Concept
+SerialisedVespaConcept: TypeAlias = list[dict[str, str]]
+
+
+class S3Uri:
```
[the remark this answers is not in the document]
jesse-c: I haven't heard of that one, but now I recall seeing it in some of our dependency lists, haha.

flows/aggregate_inference_results.py
```diff
@@ -41,61 +77,66 @@ class Config:
...
+    """Builds an identifier from the start time and name of the flow run."""
     run_context = get_run_context()
     if not run_context:
         raise MissingContextError()
     start_time = run_context.flow_run.start_time.replace(tzinfo=None).isoformat(
         timespec="minutes"
     )
     run_name = run_context.flow_run.name
-    return f"{aggregate_inference_results_prefix}/{start_time}-{run_name}"
+    return f"{start_time}-{run_name}"
```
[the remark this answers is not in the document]
olaughter: Just that it felt like it made more sense to add the `inference_results` prefix later on to simplify what was being passed around. I think as that value will basically never change it should add much too. Also the thinking behind using start_time + run time was to make it both sortable and easily findable by a human, there is also the run uuid which could have been used and had better guarantee of uniqueness than the run_name, but figured it would end up as something of a faff to work with

PR #428, opened 2025-05-27
refactor: Use stricter type hinting for `str`s and constants

jesse-c:
This came out of me, in a code review, asking for more of these `TypeAlias`s to be added, and wondering if they're effective for static checks, and not just making function signatures more human readable.

Turns out, they weren't good for the former[^1]. There is a little boilerplate with having to instantiate the type, but alas, I feel the potential type checker warnings is worth it. We'll see though, and maybe there's a way to avoid this.

[^1]: https://climate-policy-radar.slack.com/archives/C078S46DJFP/p1748337885539729

### DIFF
--- a/flows/boundary.py
+++ b/flows/boundary.py
@@ -21,9 +21,10 @@ from pathlib import Path
 from typing import (
     Any,
     Callable,
+    Final,
     Iterable,
+    NewType,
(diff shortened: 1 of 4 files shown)


THOR300 (approved):
Really nice!

PR #433, opened 2025-05-28
test: Add profiler

jesse-c:
**work-in-progress**

```python
from flows.utils import Profiler


@Profiler(printer=print)
def foo():
    time.sleep(5)
    return 7 * 2


if __name__ == "__main__":
    with Profiler(printer=print):
        print(1)
    foo()
```

```shell
poetry run python profiler_test.py
1
done in: 0.00 seconds
foo done in: 5.00 seconds
```

### DIFF
--- a/flows/utils.py
+++ b/flows/utils.py
@@ -1,15 +1,19 @@
+import functools
 import inspect
 import os
 import re
+import time
 from collections.abc import Generator
+from dataclasses import dataclass, field
 from io import BytesIO
 from pathlib import Path
-from typing import TypeAlias, TypeVar
+from typing import Callable, TypeAlias, TypeVar
 
 import boto3
 from botocore.exceptions import ClientError
 from prefect.settings import PREFECT_UI_URL
 from prefect_slack.credentials import SlackWebhook
+from typing_extensions import Self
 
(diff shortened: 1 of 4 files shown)


flows/utils.py
```diff
@@ -278,3 +282,46 @@ def filter_non_english_language_file_stems(
             file_stems,
         )
     )
+
+
+@dataclass
+class Profiler:
+    """Context manager for profiling and printing the duration."""
+
+    printer: Callable[[str], None] | None = None
```
olaughter, line 291: Could make `print` the default here?

flows/utils.py
```diff
@@ -278,3 +282,46 @@ def filter_non_english_language_file_stems(
...
+
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            # Use function name as name, if none provided
+            profiler_name = self.name or func.__name__
+            with Profiler(printer=self.printer, name=profiler_name):
+                result = func(*args, **kwargs)
+            return result
+
+        return wrapper
```
olaughter, line 327: Nice adding the decorator option for this!

olaughter (approved):
Neat!

PR #435, opened 2025-05-28
Feature/pla 619 index from aggregated inference results 

THOR300:
This Pull Request: 
---
- Integrates functionality to index aggregated inference results in to vespa via a prefect flow that can be called by a parent process which handles concurrency etc.


flows/index_from_aggregate_results.py
```diff
@@ -0,0 +1,65 @@
...
+            serialised_concepts=concepts,
+            vespa_connection_pool=vespa_connection_pool,
+        )
+        if not response.is_successful():
+            # TODO: Collect error
+            raise ValueError(
+                f"Failed to update text block {text_block_id} in Vespa: {response}"
+            )
+
+    # TODO: Consider updating concept counts solution.
```
THOR300, line 65: Add logging and retries, time outs?

flows/index_from_aggregate_results.py
```diff
@@ -0,0 +1,65 @@
...
+    """Index aggregated inference results from S3 into Vespa for a document."""
+    s3 = boto3.client("s3")
+    response = s3.get_object(Bucket=s3_uri.bucket, Key=s3_uri.key)
+    body = response["Body"].read().decode("utf-8")
+    # Do we want to convert to a pydantic type here?
+    # Good for validation but seemes odd if we're just going to convert it back to json.
+    aggregated_inference_results = json.loads(body)
+
+    # TODO: Find a better solution for this
+    document_import_id = DocumentImportId(s3_uri.key.split("/")[-1].split(".")[0])
```
THOR300, line 50: Find a better solution for this!

flows/index_from_aggregate_results.py
```diff
@@ -0,0 +1,61 @@
...
+
+async def _update_text_block(
+    text_block_id: TextBlockId,
+    document_import_id: DocumentImportId,
+    serialised_concepts: list[dict[str, Any]],
+    vespa_connection_pool: VespaAsync,
+) -> VespaResponse:
+    """Update a text block in Vespa with the given concepts."""
+
+    data_id = f"id:doc_search:document_passage::{document_import_id}.{text_block_id}"
```
THOR300, line 21: Make a function

flows/index_from_aggregate_results.py
```diff
@@ -0,0 +1,61 @@
...
+            serialised_concepts=concepts,
+            vespa_connection_pool=vespa_connection_pool,
+        )
+        if not response.is_successful():
+            # TODO: Collect error
+            raise ValueError(
+                f"Failed to update text block {text_block_id} in Vespa: {response}"
+            )
+
+    # TODO: Consider updating concept counts solution.
```
THOR300, line 61: I think we should handle it in a separate flow.

flows/index_from_aggregate_results.py
```diff
@@ -0,0 +1,110 @@
...
+        raise ValueError(
+            f"Some text blocks were not found in Vespa: {text_blocks_not_in_vespa}"
+        )
+    elif update_errors:
+        raise ValueError(f"Some updates to Vespa failed: {update_errors}")
+
+    logger.info(
+        "Successfully indexed all aggregated inference results for document import ID: %s",
+        document_import_id,
+    )
```
THOR300, line 110: Reads still required: 
---
Having to put a read back in due to the problem listed below. I'm utilising the generator which was found to be very fast for documents when I investigated last. We will only require one read per document for a full run so maybe this is acceptable. All ears for a different solution. 

Problem: 
--- 
I found an example in the test data where the text block id is "p_0_b_1 and the
data id of the passage is:
"id:doc_search:document_passage::CCLW.executive.4934.1571.1"
Meaning the passage silently didn't get updated as I was providing the wrong data id.

Looking at the [indexer](https://github.com/climatepolicyradar/navigator-search-indexer/blob/main/src/index/vespa_.py#L324) we use the index of the text block in the for loop as the data suffix.
And thus for a text block with id "p_0_b_1" the data id does not match the text block id. Thus I'm adding in functionality to query vespa.

flows/index_from_aggregate_results.py
```diff
@@ -0,0 +1,110 @@
...
+        )
+    elif text_blocks_not_in_vespa:
+        raise ValueError(
+            f"Some text blocks were not found in Vespa: {text_blocks_not_in_vespa}"
+        )
+    elif update_errors:
+        raise ValueError(f"Some updates to Vespa failed: {update_errors}")
+
+    logger.info(
+        "Successfully indexed all aggregated inference results for document import ID: %s",
```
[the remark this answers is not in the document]
olaughter: We actually have the aggregated concepts in scope. It would be quite nice to log how many we have added to the document

flows/aggregate_inference_results.py
```diff
@@ -46,19 +47,20 @@
...
-    ):
+    def __init__(self, bucket: str, key: str, protocol: str = "s3"):
         self.protocol = protocol
         self.bucket = bucket
         self.key = key
 
     def __str__(self) -> str:
         """Return the string representation of the S3 URI."""
-        if not self.bucket or not self.key:
-            raise ValueError("Bucket and key must be set")
```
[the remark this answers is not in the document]
THOR300: As per above they can't be null now.

flows/aggregate_inference_results.py
```diff
@@ -193,6 +200,10 @@ async def process_single_document(
         )
         vespa_concepts = combine_labelled_passages(all_labelled_passages)
 
+        if config.cache_bucket is None:
+            raise ValueError(
+                "Cache bucket must be set in config before processing documents."
+            )
```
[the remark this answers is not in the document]
THOR300: Yeah would be a better solution, I'll take a look!

flows/index_from_aggregate_results.py
```diff
@@ -0,0 +1,110 @@
...
+        data_id=vespa_data_id,
+        fields={"concepts": serialised_concepts},
+    )
+
+    return response
+
+
+@flow
+async def index_aggregate_results_from_s3_to_vespa(
+    s3_uri: S3Uri,
```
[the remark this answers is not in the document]
THOR300: How so? 

This function does need the bucket and inference results prefix to load that aggregated inference results file for the document?

flows/index_from_aggregate_results.py
```diff
@@ -0,0 +1,110 @@
...
+        schema="document_passage",
+        namespace="doc_search",
+        data_id=vespa_data_id,
+        fields={"concepts": serialised_concepts},
+    )
+
+    return response
+
+
+@flow
```
jesse-c, line 46: **suggestion:** I think this is better suited as a task, as it's more of a discrete unit of work than a flow which is tying multiple units of work together. It also would lend itself towards https://docs.prefect.io/v3/develop/task-runners, or at least our own current approach of it.
```suggestion
@task
```

flows/index_from_aggregate_results.py
```diff
@@ -0,0 +1,110 @@
...
+
+    logger = get_run_logger()
+
+    document_import_id = DocumentImportId(s3_uri.stem)
+
+    logger.info(f"Loading aggregated inference results from S3: {s3_uri}")
+    aggregated_inference_results = load_json_data_from_s3(s3_uri=s3_uri)
+
+    logger.info(
+        "Loading passages from Vespa for document import ID: %s", document_import_id
```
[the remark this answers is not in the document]
THOR300: Yeh absolutely it's not "new information", I personally like having a reasonable level of logging in jobs so we can see how far the job has progressed and where the longest time is being spent.

PR #437, opened 2025-05-28
Feature/pla 617 additions to the s3 push down aggregator 2

olaughter:
Adds default document selection to the aggregator. To do this I've made the s3 file stem functionality, (that existed in inference) more reusable.

### DIFF
--- a/flows/aggregate_inference_results.py
+++ b/flows/aggregate_inference_results.py
@@ -12,13 +12,17 @@ from prefect.exceptions import MissingContextError
 from prefect.task_runners import ConcurrentTaskRunner
 
 from flows.boundary import (
-    DocumentImportId,
     TextBlockId,
     convert_labelled_passage_to_concepts,
     s3_object_write_text,
 )
-from flows.inference import DOCUMENT_TARGET_PREFIX_DEFAULT
+from flows.inference import (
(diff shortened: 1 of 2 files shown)


flows/aggregate_inference_results.py
```diff
@@ -72,7 +76,9 @@ class Config:
     """Configuration used across flow runs."""
 
     cache_bucket: str | None = None
-    document_source_prefix: str = DOCUMENT_TARGET_PREFIX_DEFAULT
+    raw_inference_results_prefix: str = DOCUMENT_TARGET_PREFIX_DEFAULT
+    parser_output_prefix: str = DOCUMENT_SOURCE_PREFIX_DEFAULT
+    pipeline_state_prefix: str = "input"
```
olaughter, line 81: This has made me realise we ought to standardise these config objects some more rather than having duplication across each of them, perhaps we could have them inherit from a common top level Config or something

olaughter:
Realised a flaw in this that means the functionality isnt yet re-purposable across the flows, so still working out a few creases, have set to draft

olaughter:
Closing in favour of: https://github.com/climatepolicyradar/knowledge-graph/pull/447

Would have been nice to get some reusability going in these interfaces but its proven to be a much bigger task to tie everything together that needs it

PR #439, opened 2025-05-29
refactor: Allow flexible concurrency for inference

jesse-c:
There was a for loop over per-classifier, which within it had batches of documents. Now, it runs multiple classifiers at once, with each receiving the same batch of documents as previously.

This makes it a `N`x speed-up, where `N` is the concurrency limit.

Same approach as #436.

FIXES PLA-620


jesse-c:

- #412
    - #421
        - #436
            - #439 ◀

jesse-c:
Artifact example:

CleanShot 2025-05-29 at 16 57 42

flows/inference.py
```diff
@@ -653,42 +626,147 @@ async def classifier_inference(
...
+        {"Classifier": spec.name, "Alias": spec.alias, "Status": "✗"}
+        for spec in failures_classifier_specs
+    ]
+
+    # Create a single artifact with overview in description and classifier details in table
+    await create_table_artifact(
+        key=f"classifier-inference-{config.aws_env.value}",
+        table=classifier_details,
+        description=overview_description,
+    )
```
olaughter, line 764: This is soooooo good! Just seen the output in prefect!

olaughter: Obviously out of scope for now, but if we where to write the same fields into the data lake we would get some real nice tracking over time too!

scripts/do_outputs_align_for_a_document.py
```diff
@@ -8,7 +8,7 @@
 
 import boto3
 import typer
-import yaml
+import yaml  # pyright: ignore[reportMissingModuleSource]
```
olaughter, line 11: Ohh weird, its [definitely installed](https://github.com/climatepolicyradar/knowledge-graph/blob/main/pyproject.toml#L30), interestingly seeing we've added pyright ignores in other places for this too...

flows/inference.py
```diff
@@ -653,42 +626,147 @@ async def classifier_inference(
...
+
+                    logger.error(
+                        f"flow run's state was not completed. Flow run name: `{result.name}`",
+                    )
+                elif result.state.type == StateType.COMPLETED:
+                    classifier_spec = ClassifierSpec(
+                        name=result.parameters["classifier_name"],
+                        alias=result.parameters["classifier_alias"],
+                    )
+                    successes[classifier_spec] = result
```
olaughter, line 699: **suggestion:** refactor results processing to have a dedicated function. It would simplify `classifier_inference` and let us unit test it

flows/inference.py
```diff
@@ -627,7 +598,9 @@ async def classifier_inference(
     if not config:
         config = await Config.create()
 
-    print(f"Running with config: {config}")
+    logger = get_run_logger()
+
+    logger.info(f"Running with config: {config}")
```
olaughter, line 603: Are we preferring the logger again? I thought we had turned against using it in favour of the `log_prints` functionality? 😅

olaughter (commented):
Loving seeing this, even in draft! Looking really good so far!

flows/inference.py
jesse-c: Agreed!

scripts/do_outputs_align_for_a_document.py
jesse-c: I noticed that too, and it's annoying. We'll have to sort it later.

flows/inference.py
jesse-c: ✓

flows/inference.py
jesse-c: ✓

jesse-c:
Previously reported:

> .. each flow within the parent flow took ~6h ea. Based on all 3 going at once, that means it should finish in ~4h more. So that’s a 3x speed-up on inference, but in practice we’ll probably run with a concurrency limit of much higher, so could be 20x faster.

flows/inference.py
```diff
@@ -56,6 +55,8 @@
 }
 DOCUMENT_TARGET_PREFIX_DEFAULT: str = "labelled_passages"
 
+CLASSIFIER_CONCURRENCY_LIMIT: Final[PositiveInt] = 20
```
olaughter, line 58: **praise:** nice idea use PositiveInt for this

flows/inference.py
```diff
@@ -596,9 +565,51 @@ async def run_classifier_inference_on_batch_of_documents(
...
+    printer=print,
+    name="processing results",
+)
+def group_results(
+    results: list[FlowRun | BaseException],
+) -> tuple[
+    list[FlowRun | BaseException],
+    dict[ClassifierSpec, FlowRun],
+]:
+    failures: list[FlowRun | BaseException] = []
```
olaughter, line 578: **suggestion:** add a docstring here

flows/inference.py
```diff
@@ -653,42 +665,115 @@ async def classifier_inference(
...
-        for result in results:
-            # We'd never expect the flow run name to be missing, but
-            # Prefect does account for that in their class.
-            name: str = result.name
-            if result.state is None:
-                failures[classifier_spec].append(name)
-            else:
-                if result.state.type != StateType.COMPLETED:
-                    failures[classifier_spec].append(name)
+    failures, successes = group_results(results)
```
olaughter, line 706: Nice, this is looking really clean!

flows/inference.py
```diff
@@ -653,42 +665,115 @@ async def classifier_inference(
...
+        successes=successes,
+        failures_classifier_specs=failures_classifier_specs,
+        batch_size=batch_size,
+        classifier_concurrency_limit=classifier_concurrency_limit,
+    )
 
     if failures:
         raise ValueError(
-            f"some classifier specs. had failures: {', '.join(map(str, failures.keys()))}"
+            f"some classifier specs. had failures:.: {','.join(map(str, failures_classifier_specs))}"
```
olaughter, line 722: **suggestion:** Was this a typo?
```suggestion
            f"some classifier specs. had failures: {','.join(map(str, failures_classifier_specs))}"
```

flows/inference.py
```diff
@@ -653,42 +665,115 @@ async def classifier_inference(
...
+
+## Overview
+- **Environment**: {config.aws_env.value}
+- **Total Documents Processed**: {total_documents}
+- **Total Classifiers**: {total_classifiers}
+- **Successful Classifiers**: {successful_classifiers}
+- **Failed Classifiers**: {failed_classifiers}
+- **Batch Size**: {batch_size}
+- **Classifier Concurrency Limit**: {classifier_concurrency_limit}
+- **Generated At**: {datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")}
```
olaughter, line 754: **quible:** Could also go with the following which I think is cleaner:
```suggestion
- **Generated At**: {datetime.datetime.isoformat(timespec="seconds", sep=" ")}
```

olaughter: Good to add a timezone too, not sure we're super consistent on that mind...

tests/flows/test_inference.py
```diff
@@ -408,3 +411,38 @@ def test_remove_sabin_file_stems(
 ):
     result = remove_sabin_file_stems(input_stems)
     assert result == expected_output
+
+
+def test_group_results(snapshot):
```
olaughter, line 416: Nice use of snapshots 👌

olaughter (approved):

flows/inference.py
jesse-c: ✓ The best time is to start now then. tbh I think this is unnecessary, since the artifact itself is a thing that is known when it was created (generated). So I'll remove it, and even some of the dupe params that are in the `FlowRun` could be removed.

PR #441, opened 2025-05-29
Feature: Add a parent flow and deployment for indexing from aggregate results 

THOR300:
This Pull Request: 
--- 
- Is adding a flow and deployment for the indexing from aggregate results. 
- This flow can be run concurrently as a sub deployment of a parent flow that controls batch indexing. 

Details of the staging run can be seen [here](https://linear.app/climate-policy-radar/issue/PLA-619/passage-indexer-to-use-aggregated-results#comment-da58306f).


flows/index_from_aggregate_results.py
```diff
@@ -108,3 +115,38 @@ async def index_aggregate_results_from_s3_to_vespa(
         "Successfully indexed all aggregated inference results for document import ID: %s",
         document_import_id,
     )
+
+
+@flow
+async def run_indexing_from_aggregate_results(
+    s3_bucket: str,
```
olaughter, line 122: **suggestion:** use the config option that we use as a standard everywhere and saves manually inputting this value

flows/index_from_aggregate_results.py
```diff
@@ -108,3 +115,38 @@ async def index_aggregate_results_from_s3_to_vespa(
...
+    aggregate_inference_results_s3_keys: list[str],
+) -> None:
+    """Index aggregated inference results from a list of S3 URIs into Vespa."""
+
+    logger = get_run_logger()
+    logger.info("Starting indexing from aggregate results...")
+
```
olaughter, line 129: **suggestion:** use prints and don't log that we are starting as prefect already will tell us if a flow has started

flows/index_from_aggregate_results.py
```diff
@@ -108,3 +115,38 @@ async def index_aggregate_results_from_s3_to_vespa(
         "Successfully indexed all aggregated inference results for document import ID: %s",
         document_import_id,
     )
+
+
+@flow
+async def run_indexing_from_aggregate_results(
+    s3_bucket: str,
+    aggregate_inference_results_s3_keys: list[str],
```
olaughter, line 123: **suggestion:** Would be a lot nicer to just use the run identifier here and let the flow worry about the `inference_results` subprefix as well as what specific ids where in there

tests/flows/test_index_from_aggregate_results.py
```diff
@@ -138,3 +141,27 @@ async def test_index_from_aggregated_inference_results__error_handling(
...
+
+    # Mock the get_vespa_search_adapter_from_aws_secrets function to return local_vespa_search_adapter
+    with patch(
+        "flows.index_from_aggregate_results.get_vespa_search_adapter_from_aws_secrets",
+        return_value=local_vespa_search_adapter,
+    ):
+        await run_indexing_from_aggregate_results(
+            s3_bucket=mock_bucket,
+            aggregate_inference_results_s3_keys=file_keys,
+        )
```
olaughter, line 167: **addition:** be good to assert something here to know its run, I'm not that interested in duplicating whats in the other test already, but we could at least take the state from `run_indexing_from_aggregate_results` and assert it has completed successfully

flows/index_from_aggregate_results.py
```diff
@@ -108,3 +115,38 @@ async def index_aggregate_results_from_s3_to_vespa(
         "Successfully indexed all aggregated inference results for document import ID: %s",
         document_import_id,
     )
+
+
+@flow
```
olaughter, line 120: **question:** Asking more for my own understanding as I assume there is reasoning behind this, but what's the logic for having two separate flows for the parent and child? Do flows work better that tasks for adding concurrency or do you imagine there being times were we'll want to run the child in isolation for a specific document or something?

flows/index_from_aggregate_results.py
```diff
@@ -108,3 +115,38 @@ async def index_aggregate_results_from_s3_to_vespa(
...
+    aggregate_inference_results_s3_keys: list[str],
+) -> None:
+    """Index aggregated inference results from a list of S3 URIs into Vespa."""
+
+    logger = get_run_logger()
+    logger.info("Starting indexing from aggregate results...")
+
+    s3_uri_list = [
+        S3Uri(bucket=s3_bucket, key=key) for key in aggregate_inference_results_s3_keys
+    ]
```
olaughter, line 132: **thought:** I've already flagged above and on the other PR that I don't think we want to be passing around s3 uris for these flows, also seeing this here in that we are just appending the bucket to the key and the user has had to specify both

flows/index_from_aggregate_results.py
jesse-c: I just commented about this, on the other one: https://github.com/climatepolicyradar/knowledge-graph/pull/435/files#r2120580030. I'd add to that too, that I think it's overkill having them as separate files.

jesse-c (commented):
Why is this a new `Deployment`, instead of refactoring the existing one? I think refactoring the existing one matches the ticket, it reduces 'noise' in Prefect by virtue of there being less `Deployment`s, and it would mean we don't need to modify the automations work.

flows/index_from_aggregate_results.py
THOR300: I'm not sure I entirely understand the question, are you asking why we have a parent and child flow/task or why we've decided to decorate the functions as flows / tasks respectively?

flows/index_from_aggregate_results.py
THOR300: Commented on the other PR.

flows/index_from_aggregate_results.py
olaughter: Yes? Let me attempt a reword to make the question from my side more specific: Adding subflows introduces more hierarchy to the flow and I'm interested in the logic behind doing it here when it could just be a task? I can't imagine us wanting to run this subflow in isolation and its also not really set up to do so. We also do this elsewhere and I'm not sure I have a clear understanding of what we gain from it

THOR300:
> Why is this a new `Deployment`, instead of refactoring the existing one? I think refactoring the existing one matches the ticket, it reduces 'noise' in Prefect by virtue of there being less `Deployment`s, and it would mean we don't need to modify the automations work.

Yeh super valid, I wanted to get a document level flow working that would get the passages updated correctly from the aggregated results, this deployment is just to get the flow deployed easily in the cloud as well as handling the concerns that we wouldn't want within the document level flow like instantiating the connection pool. 

I'm happy to look at refactoring this to be held in the `index.py` model and utilised in the `index_labelled_passages_from_s3_to_vespa` flow / deployment for example at a later date if that's preferred although at a glance it looks like it would be essentially a rewrite, e.g. no longer running on classifiers, getting s3 paths or running updates_by_s3 but run ids etc. 👍

flows/index_from_aggregate_results.py
```diff
@@ -44,19 +56,27 @@ async def _update_vespa_passage_concepts(
     return response
 
 
-@task
+@flow
```
THOR300, line 59: Had to convert back to a flow due to [this](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/0683dbc4-6885-7f86-8000-dd022087cc8c?entity_id=0683dbcb-da2e-7f3a-8000-d7a3c3be7bb6&state=cancelled&state=cancelling&state=completed&state=crashed&state=failed&state=paused&state=pending&state=running&state=scheduled&entity=flowRuns&entity=taskRuns&entity=events&entity=logs&entity=artifacts) run in prefect cloud: 


> Error encountered when computing cache key - result will not be persisted.
  File "/usr/local/lib/python3.10/site-packages/prefect/cache_policies.py", line 374, in compute_key
    return hash_objects(hashed_inputs, raise_on_failure=True)
  File "/usr/local/lib/python3.10/site-packages/prefect/utilities/hashing.py", line 89, in hash_objects
    raise HashError(msg)
prefect.exceptions.HashError: Unable to create hash - objects could not be serialized.
  JSON error: Unable to serialize unknown type: <class 'vespa.application.VespaAsync'>
  Pickle error: cannot pickle '_thread.RLock' object

The above exception was the direct cause of the following exception:

  File "/usr/local/lib/python3.10/site-packages/prefect/task_engine.py", line 169, in compute_transaction_key
    key = self.task.cache_policy.compute_key(
  File "/usr/local/lib/python3.10/site-packages/prefect/cache_policies.py", line 214, in compute_key
    policy_key = policy.compute_key(
  File "/usr/local/lib/python3.10/site-packages/prefect/cache_policies.py", line 384, in compute_key
    raise ValueError(msg) from exc
ValueError: Unable to create hash - objects could not be serialized.
  JSON error: Unable to serialize unknown type: <class 'vespa.application.VespaAsync'>
  Pickle error: cannot pickle '_thread.RLock' object

This often occurs when task inputs contain objects that cannot be cached like locks, file handles, or other system resources.

To resolve this, you can:
  1. Exclude these arguments by defining a custom `cache_key_fn`
  2. Disable caching by passing `cache_policy=NO_CACHE`

PR #448, opened 2025-06-02
fix: Don't define types twice

jesse-c:
This is something that fell through the cracks. I've put it in `utils` since that's imported by the `boundary`.

TOWARDS PLA-617


THOR300 (approved):
Looking good!

PR #452, opened 2025-06-03
Adding exception handling to the aggregate indexing flow.

THOR300:
This Pull Request: 
---
- Adds exception handling to the aggregate indexing flow to achieve the following requirement: 

> Passage indexing collects all failures rather than failing immediately

Towards: 
- https://linear.app/climate-policy-radar/issue/PLA-624/passage-indexing-to-use-interleaved-concurrency-and-result-collection

### DIFF
--- a/tests/flows/test_index_from_aggregate_results.py
+++ b/tests/flows/test_index_from_aggregate_results.py
@@ -161,16 +161,13 @@ async def test_run_indexing_from_aggregate_results(
     mock_s3_client,
     mock_bucket: str,
     mock_bucket_inference_results: dict[str, dict[str, Any]],
+    aggregate_inference_results_import_ids: list[DocumentImportId],
     mock_run_output_identifier_str: str,
     s3_prefix_inference_results: str,
     test_aggregate_config,
 ) -> None:
     """Test that we loaded the inference results from the mock bucket."""
 
-    document_import_ids = [
-        DocumentImportId(Path(file_key).stem)
-        for file_key in mock_bucket_inference_results.keys()
-    ]
     run_output_identifier = RunOutputIdentifier(mock_run_output_identifier_str)
 
     with patch(
@@ -179,7 +176,7 @@ async def test_run_indexing_from_aggregate_results(
     ):
         await run_indexing_from_aggregate_results(
             run_output_identifier=run_output_identifier,
-            document_import_ids=document_import_ids,
+            document_import_ids=aggregate_inference_results_import_ids,
             config=test_aggregate_config,
         )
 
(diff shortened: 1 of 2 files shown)


tests/flows/test_index_from_aggregate_results.py
```diff
@@ -229,3 +229,42 @@ async def test_run_indexing_from_aggregate_results(
...
+    mock_bucket_inference_results: dict[str, dict[str, Any]],
+    s3_prefix_inference_results: str,
+    test_aggregate_config,
+) -> None:
+    """Test that we handled the exception correctly during passage indexing."""
+
+    NON_EXISTENT_ID = DocumentImportId("non_existent_document")
+    document_import_ids = [
+        DocumentImportId(Path(file_key).stem)
+        for file_key in mock_bucket_inference_results.keys()
```
olaughter, line 250: This could theoretically be empty, perhaps theres a better way to build this to expect the specific ids that should be in there, that way its clear if anything goes wrong with the test setup

tests/flows/test_index_from_aggregate_results.py
```diff
@@ -229,3 +229,42 @@ async def test_run_indexing_from_aggregate_results(
...
+    ] + [NON_EXISTENT_ID]
+
+    run_output_identifier_str = Path(next(iter(mock_bucket_inference_results))).parts[1]
+    run_output_identifier = RunOutputIdentifier(run_output_identifier_str)
```
olaughter, line 254: This is a grim way of getting this from a test fixture under our control, would it not be simpler to just include the run identifier in the fixture somehow?

olaughter (approved):
Nice! Couple minor points below

flows/index_from_aggregate_results.py
```diff
@@ -155,13 +156,34 @@ async def run_indexing_from_aggregate_results(
...
+            index_aggregate_results_from_s3_to_vespa(
                 config=config,
                 run_output_identifier=run_output_identifier,
                 document_import_id=document_import_id,
                 vespa_connection_pool=vespa_connection_pool,
             )
+            for document_import_id in document_import_ids
+        ]
+
+        results = await asyncio.gather(*tasks, return_exceptions=True)
```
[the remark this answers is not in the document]
olaughter: Looks to be on its own PR, here: https://github.com/climatepolicyradar/knowledge-graph/pull/453

tests/flows/test_index_from_aggregate_results.py
THOR300: Noice I'll have a think 👍

THOR300: Updated the test fixtures to be explicit around document ids as opposed to dynamically loading from the found files 👍

PR #453, opened 2025-06-03
Feature/pla 624 passage indexing to use interleaved concurrency

THOR300:
This Pull Request: 
---
- Updates the passage indexer to run concurrently across multiple hosts. 
- This PR meets the requirement: 

> Passage indexing is done with interleaved concurrency

Towards: 
- https://linear.app/climate-policy-radar/issue/PLA-624/passage-indexing-to-use-interleaved-concurrency-and-result-collection


flows/index_from_aggregate_results.py
```diff
@@ -187,3 +204,86 @@ async def run_indexing_from_aggregate_results(
...
+    for result in results:
+        if isinstance(result, BaseException):
+            failures.append(result)
+        else:
+            successes.append(result)
+
+    if failures:
+        raise ValueError(
+            f"Some batches of documents had failures: {len(failures)}/{len(results)} failed."
+        )
```
THOR300, line 289: Todo: Add a deployment entry for this.

flows/aggregate_inference_results.py
```diff
@@ -112,6 +112,12 @@ def cache_bucket_str(self) -> str:
             )
         return self.cache_bucket
 
+    def to_json(self) -> dict[str, Any]:
+        """Convert the Config instance to a dictionary, handling complex types."""
+        result = asdict(self)
+        result["aws_env"] = self.aws_env.value  # serialize AwsEnv manually
+        return result
```
jesse-c, line 119: **polish:** I'd change this to a Pydantic model and then rely on its ability to serialise things, and if needed, hook into special serialisation for this 1 field.

flows/index_from_aggregate_results.py
```diff
@@ -125,17 +138,21 @@ async def index_aggregate_results_from_s3_to_vespa(
...
     """Index aggregated inference results from a list of S3 URIs into Vespa."""
 
     logger = get_run_logger()
 
-    if config is None:
-        config = await Config.create()
+    config = Config(**config_json)
+    logger.info(
+        f"Running indexing for batch with config: {config}, "
+        f"run_output_identifier: {run_output_identifier},  "
```
jesse-c, line 153: **quibble:** This is directly in the `FlowRun` parameters, so probably doesn't need to be repeated here. The other 2, are not (always, as the config may be specified) in the parameters, and thus are helpful to be printed.

flows/index_from_aggregate_results.py
```diff
@@ -187,3 +204,86 @@ async def run_indexing_from_aggregate_results(
...
+    on_crashed=[SlackNotify.message],
+)
+async def run_indexing_from_aggregate_results(
+    run_output_identifier: RunOutputIdentifier,
+    document_import_ids: list[DocumentImportId],
+    config: Config | None = None,
+    batch_size: int = DEFAULT_DOCUMENTS_BATCH_SIZE,
+    indexer_concurrency_limit: PositiveInt = DEFAULT_INDEXER_CONCURRENCY_LIMIT,
+) -> None:
+    """Index aggregated inference results from a list of S3 URIs into Vespa."""
```
olaughter, line 221: How come this needs another parent flow? Not sure I understand why this couldnt be implemented directly on `run_indexing_from_aggregate_results` / `index_aggregate_results_for_batch_of_documents`

flows/index_from_aggregate_results.py
```diff
@@ -19,16 +21,27 @@
...
     VespaDataId,
     VespaHitId,
+    function_to_flow_name,
+    generate_deployment_name,
     get_data_id_from_vespa_hit_id,
     get_document_passages_from_vespa__generator,
     get_vespa_search_adapter_from_aws_secrets,
 )
+from flows.inference import (
+    Profiler,
```
jesse-c, line 36: **fix:** This should be imported from `flow.utils`, not by a transitive dependency here that could change.

flows/index_from_aggregate_results.py
```diff
@@ -19,16 +21,23 @@
...
+    generate_deployment_name,
     get_data_id_from_vespa_hit_id,
     get_document_passages_from_vespa__generator,
     get_vespa_search_adapter_from_aws_secrets,
 )
+from flows.utils import Profiler, SlackNotify, iterate_batch, wait_for_semaphore
 
-DEFAULT_INDEXER_MAX_CONNECTIONS: Final[PositiveInt] = 50
+# How many connections to Vespa to use for indexing.
+DEFAULT_VESPA_MAX_CONNECTIONS_AGG_INDEXER: Final[PositiveInt] = 50
```
jesse-c, line 38: **fix:** This should be available as a parameter in the top-level flow.

flows/aggregate_inference_results.py
THOR300: Thanks, I did consider this but going to skip so we can get this in with the time we have!

olaughter (approved):

PR #454, opened 2025-06-03
Bump memory & make errors clearer

olaughter:
The last PR missed the mark slightly in that it still didn't result in red blocks being reported n the prefect ui, the reason being it was not actually failing the task. Making this clearer should surface errors faster.

Additionally the container ran out of memory. Theres a better long term version for this but bumping the memory is also a fast way to fix until then. I believe this is a problem because we are opening all the raw inference results before aggregating, so I will follow up with a fix and drop the memory again. But I'd quite like to have it running while I do that


THOR300 (approved):
LGTM

PR #456, opened 2025-06-03
refactor: Index both document passages and family documents together

jesse-c:
I've counted all concepts counts for all the updates that worked, and the failures are reported.

One thing I've experienced here is that I feel by not doing things the Prefect way[^1], that I'm fighting Prefgect. For example, we aim to be resilient and do _as much as possible_, amongst the failures.

This shows up in how we do tasks gathering and then raise _dumb_ exceptions, to get Prefect to show an overall, failed state. Instead, if we were more Prefect-first, I think we could avoid this munging of success and failure returns and reduce code as well.

As from PLA-622 and our discussion on naming (indexing both things), I've used that naming around here.

TOWARDS PLA-622

[^1]: https://climate-policy-radar.slack.com/archives/C078S46DJFP/p1748959406451039

### DIFF
--- a/flows/result.py
+++ b/flows/result.py
@@ -1,0 +1,100 @@
+"""
+A Result type implementation for Python inspired by Rust's Result type.
+
+This module provides a type-safe way to handle operations that can either succeed or fail,
+without relying on exceptions for control flow.
+"""
+
+from dataclasses import dataclass
+from typing import Any, Callable, Generic, TypeVar
+
+T = TypeVar("T")
+E = TypeVar("E")
+
+
+@dataclass
+class Ok(Generic[T]):
+    """Represents a successful result containing a value of type T."""
+
+    _value: T
+
+    def __str__(self):
+        """Return string representation of the Ok result."""
+        return f"Ok({self._value})"
+
+
+@dataclass
+class Err(Generic[E]):
+    """Represents a failed result containing an error of type E."""
+
+    _error: E
+
+    def __str__(self):
+        """Return string representation of the Err result."""
+        return f"Err({self._error})"
+
+
+Result = Ok[T] | Err[E]
+"""Type alias for a Result that can be either Ok[T] or Err[E]."""
+
+
+def map_ok(result: Result[T, E], fn: Callable[..., T]) -> Result[T, E]:
+    """Apply a function to the value inside an Ok result, leaving Err results unchanged."""
+    match result:
+        case Ok(value):
+            return Ok(fn(value))
+        case Err(_):
+            return result
+
+
+def map_err(result: Result[T, E], fn: Callable[..., E]) -> Result[T, E]:
+    """Apply a function to the error inside an Err result, leaving Ok results unchanged."""
+    match result:
+        case Ok(_):
+            return result
+        case Err(error):
+            return Err(fn(error))
+
+
+def unwrap_ok(result: Result[T, E]) -> T:
+    """Extract the value."""
+    match result:
+        case Ok(value):
+            return value
+        case Err(err):
+            raise ValueError(f"can't unwrap: {err}")
+
+
+def unwrap_err(result: Result[T, E]) -> E:
+    """Extract the error."""
+    match result:
+        case Ok(value):
+            raise ValueError(f"can't unwrap: {value}")
+        case Err(err):
+            return err
+
+
+def is_ok(result: Result[T, E]) -> bool:
+    """Check if a Result is an Ok variant."""
+    match result:
+        case Ok(_):
+            return True
+        case Err(_):
+            return False
+
+
+def is_err(result: Result[T, E]) -> bool:
+    """Check if a Result is an Err variant."""
+    match result:
+        case Ok(_):
+            return False
+        case Err(_):
+            return True
+
+
+@dataclass
+class Error:
+    """A simple and generic error with optional, helpful metadata"""
+
+    msg: str
+    metadata: dict[Any, Any] | None


flows/index_from_aggregate_results.py
```diff
@@ -56,27 +61,34 @@ async def _update_vespa_passage_concepts(
...
-async def index_aggregate_results_from_s3_to_vespa(
+# As of 2025-06-03, the Concept from the cpr_sdk isn't hashable
+@dataclass(frozen=True)
+class SimpleConcept:
+    id: str
+    name: str
+
+
+@flow(log_prints=True)
+async def index_document_passages(
```
olaughter, line 72: Hell yeah to this naming!

flows/index_from_aggregate_results.py
```diff
@@ -88,48 +100,98 @@ async def index_aggregate_results_from_s3_to_vespa(
...
+    concepts_counts: Counter[SimpleConcept] = Counter()
+
+    for text_block_id, serialised_concepts in aggregated_inference_results.items():
         if TextBlockId(text_block_id) not in list(passages_in_vespa.keys()):
-            text_blocks_not_in_vespa.append(text_block_id)
+            error = Error(
+                msg="text block not found in Vespa",
+                metadata={"text_block_id": TextBlockId(text_block_id)},
+            )
+            failures.append(Err(error))
```
olaughter, line 118: Nice to see Result in action!

flows/index_from_aggregate_results.py
```diff
@@ -88,48 +100,98 @@ async def index_aggregate_results_from_s3_to_vespa(
...
 
-    if text_blocks_not_in_vespa or update_errors:
-        raise ValueError(
-            f"Error with {document_import_id}: {text_blocks_not_in_vespa=}, {update_errors=}"
-        )
+        concepts = [
+            SimpleConcept(id=concept["id"], name=concept["name"])
+            for concept in serialised_concepts
+        ]
+        concepts_counts.update(concepts)
```
olaughter, line 147: **musing:** the fact that this is happening here within `index_document_passages`, makes me wonder if the structuring is off somehow. As the counts are built so specifically for `index_document_passages` but its also beyond the purpose of the function itself? I don't feel very excited about seeing a fancy refactor now when we are pressed for time, so this is probably a question for future thinking, but do you think a structure could work where the document is open in `run_indexing_from_aggregate_results` and passed in to both? I guess that would be unpleasant with passing the data into flows though...

flows/index_from_aggregate_results.py
```diff
@@ -88,48 +100,98 @@ async def index_aggregate_results_from_s3_to_vespa(
...
+
+    print(f"{len(failures)} failures, {len(successes)} successes")
+
+    return concepts_counts, failures
+
+
+@task
+async def index_family_document(
```
olaughter, line 155: **question:** Guessing this can just be a task because its really simple? Whereas the other is a flow because its looping over the whole doc?

flows/index_from_aggregate_results.py
```diff
@@ -88,48 +100,98 @@ async def index_aggregate_results_from_s3_to_vespa(
...
-    logger.info(
-        "Successfully indexed all aggregated inference results for document import ID: "
-        f"{document_import_id} into Vespa. "
-        f"Total passages updated: {len(aggregated_inference_results)}"
+    response: VespaResponse = await vespa_connection_pool.update_data(
+        schema="family_document",
+        namespace="doc_search",
+        data_id=document_import_id,
+        fields={
+            "concept_counts": concepts_counts_with_names
```
olaughter, line 173: Seeing it like this really reinforces having this as a single flow with passages, documents are such a minor indexing by comparison! 😆

olaughter (approved):

flows/index_from_aggregate_results.py
THOR300: The reason it's tricky to separate is because not all the passages from the file may be present in vespa and thus we can't just open the file and pass the content to two separate flows, we need to know what actually went in to vespa at this point! 

If we did refactor however we could return the passages we know were updated and use that in a separate flow however seeing as we've gone for a nice functional approach here.

THOR300: The reason it's tricky to separate is because not all the passages from the file may be present in vespa and thus we can't just open the file and pass the content to two separate flows, we need to know what actually went in to vespa at this point!

If we did refactor however we could return the passages we know were updated and use that in a separate flow however seeing as we've gone for a nice functional approach here.

jesse-c: I did consider that, and did do it at one point, but then started thinking about the return size. This was a funny PR as I had a lot of different ideas about how to do it, haha.

flows/index_from_aggregate_results.py
jesse-c: haha absolutely. It was a good approach to be emphasised as a strong contender (ty @THOR300).

flows/index_from_aggregate_results.py
```diff
@@ -56,27 +61,34 @@ async def _update_vespa_passage_concepts(
...
-@flow
-async def index_aggregate_results_from_s3_to_vespa(
+# As of 2025-06-03, the Concept from the cpr_sdk isn't hashable
+@dataclass(frozen=True)
+class SimpleConcept:
+    id: str
+    name: str
+
+
+@flow(log_prints=True)
```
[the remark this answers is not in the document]
jesse-c: @olaughter: Tagging you in here, as on another PR of mine, you began a small discussion about this.

flows/index_from_aggregate_results.py
jesse-c: ✓ Gave myself ~30 mins to do this and got it done, based on previous (↑) explorations yesterday. Still needs to be tested in staging.

THOR300 (approved):
Great to get the document level reporting!

PR #460, opened 2025-06-03
Support both incremental and historical runs

olaughter:
This adds a prefix alongside the run identifiers that holds all of the latest  aggregated results.

The idea here is to have all the latest ran aggregations compiled in a single place. Previously, all the documents to be drawn on for indexing would need to be within a single run identifier. Most of the time this is really useful and lets us only run on what has just been completed. But there are also times when we need to rerun on everything, or when we want to review what the current state should be. In this case it helps to also maintain all of the last aggregated documents. Another key benefit is for analytics, we can review all aggregations in one go, now in s3, later also in the data lake.

We previously had some bad experiences with using `latest` with models, but I do think this is different enough from that situation that its worth having.

---
Bonuses:
- Adds a little bit more logging
- Moves s3_uri to utils

### DIFF
--- a/flows/aggregate_inference_results.py
+++ b/flows/aggregate_inference_results.py
@@ -285,7 +285,7 @@ async def create_aggregate_inference_summary_artifact(
     task_runner=ConcurrentTaskRunner(),
 )
 async def aggregate_inference_results(
-    document_ids: list[DocumentImportId],
+    document_ids: None | list[DocumentImportId] = None,
     config: Config | None = None,
     max_concurrent_tasks: int = 5,
 ) -> RunOutputIdentifier:
--- a/flows/aggregate_inference_results.py
+++ b/flows/aggregate_inference_results.py
@@ -223,6 +223,9 @@ async def process_single_document(
         all_labelled_passages = get_all_labelled_passages_for_one_document(
             document_id, classifier_specs, config
         )
+        print(
+            f"Combining {len(all_labelled_passages)} labelled passages for {document_id}"
+        )
         vespa_concepts = combine_labelled_passages(all_labelled_passages)
 
         # Write to s3
@@ -237,6 +240,7 @@ async def process_single_document(
         s3_object_write_text(str(s3_uri), json.dumps(vespa_concepts))
         return document_id
     except ClientError as e:
+        print(e.response)
         raise AggregationFailure(
             document_id=document_id, exception=e, context=e.response
         )
--- a/flows/aggregate_inference_results.py
+++ b/flows/aggregate_inference_results.py
@@ -18,6 +18,7 @@ from flows.boundary import (
     DocumentImportId,
     TextBlockId,
     convert_labelled_passage_to_concepts,
+    s3_copy_file,
     s3_object_write_text,
 )
 from flows.inference import DOCUMENT_TARGET_PREFIX_DEFAULT
@@ -58,6 +59,11 @@ class S3Uri:
     def __str__(self) -> str:
         """Return the string representation of the S3 URI."""
         return f"{self.protocol}://{self.bucket}/{self.key}"
+
+    @property
+    def uri(self) -> str:
+        """Return the string representation of the S3 URI."""
+        return os.path.join(self.bucket, self.key)
 
     @property
     def stem(self) -> str:
@@ -144,7 +150,7 @@ def get_all_labelled_passages_for_one_document(
 
     for spec in classifier_specs:
         s3_uri = S3Uri(
-            bucket=config.cache_bucket_str,
+            bucket=config.cache_bucket,
             key=os.path.join(
                 config.document_source_prefix,
                 spec.name,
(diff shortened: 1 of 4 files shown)


flows/utils.py
```diff
@@ -92,6 +92,29 @@ def my_flow():
...
+
+    @property
+    def uri(self) -> str:
+        """Return the string representation of the S3 URI."""
+        return os.path.join(self.bucket, self.key)
+
+    @property
+    def stem(self) -> str:
+        """Return the stem of the S3 URI (the key without the extension)."""
+        return Path(self.key).stem
```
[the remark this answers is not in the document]
olaughter: This pr is just relocating the stem method I'm not really wanting to change its implementation, also s3Path is notorious for memory issues so I'd really rather not!

THOR300:
This could be really good as you say so we can quickly index from all the latest documents. 
- I'm wondering whether running two aggregating pipelines in parallel would cause any issues when trying to write to the same s3 key in `/latest` but it looks like that would be fine from the [docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/Welcome.html#ConsistencyModel)! 

> Amazon S3 does not support object locking for concurrent writers. If two PUT requests are simultaneously made to the same key, the request with the latest timestamp wins. If this is an issue, you must build an object-locking mechanism into your application.

flows/aggregate_inference_results.py
```diff
@@ -223,20 +206,37 @@ async def process_single_document(
...
+            bucket=config.cache_bucket,
             key=os.path.join(
                 config.aggregate_inference_results_prefix,
                 run_output_identifier,
                 f"{document_id}.json",
             ),
         )
         s3_object_write_text(str(s3_uri), json.dumps(vespa_concepts))
+
+        # Duplicate to latest
```
jesse-c, line 225: **suggestion:** Will there be a lot of space usage, for this duplication? Maybe instead of copying the objects, there could simply be a pointer written to a file in that last. I.e. it writes a text file that has the run output identifier, for _this_ run, AKA the latest run.

jesse-c (commented):
How would this `latest` prefix be used?

> But there are also times when we need to rerun on everything, ..

Has this had to be done lately? Is this just when we've been doing lots index runs as part of dev. work?

> .. or when we want to review what the current state should be.

I agree with the above problem, and this one. What's here is _a_ solution, but I have idea(s) for other possible solutions, around the irregular comments I've made over the last couple of months about tracking results. For example, effectively tracking the latest params somewhere centrally.

I'm not expecting necessarily consideration of _my_ ideas, but I think it could be helpful to have it expanded upon how this would be used.

jesse-c:
To add, sometimes a first solution is sufficient, so as long as it's clear how it would be used, it could go in! I know there's the whole "we'll get to it later" which invariably means how it's done forever, but that's not always true.

THOR300 (approved):
I've been doing some additional thinking on this and I do believe this is a good thing to do!

We really want the runs under the unique run prefix's to be "caches" e.g. they live for the duration of the run and can be cleaned up (or we'd quickly have ALOT of jsons in s3). 

So we have to IMO publish the results _somewhere_ before the indexer step. 

In regards to having a simple latest prefix it allows us to index on all quickly which is good for recoverability and is a view of the latest version of the data for data science let's say to interrogate pre data lake. 

In regards to how we manage this latest state I've been doing some reading and there's a few options for "best case" beyond this. The issue I see with just putting things in latest is that it's not idempotent from the indexers perspective. E.g. we can't run the indexer on `/latest` at two different points in time and get the same results as the objects in the bucket will have changed.

Some options (which I expect we won't have time for): 

1. Add object versioning to the bucket and track version pointers in a relational data store. 
This is a relatively common method which allows us to keep a track of the latest version of document. If we modelled this correctly we could even say "give me the view of latest at this date" which unlocks idempotent indexer runs.

2. Use a tool like [LakeFS](https://lakefs.io/?utm_campaign=Paid%20campaigns&utm_source=ppc&utm_content=paid_gsn_us_lakefs_230704&kw=lakefs&cpn=20625638839&utm_term=lakefs&utm_source=adwords&utm_medium=ppc&utm_campaign=lakeFS_230703&hsa_cam=20625638839&hsa_grp=154844088912&hsa_mt=b&hsa_src=g&hsa_ad=676149027349&hsa_acc=8353337013&hsa_net=adwords&hsa_kw=lakefs&hsa_tgt=kwd-2116779710627&hsa_ver=3&gad_source=1&gad_campaignid=20625638839&gbraid=0AAAAAoZtQ5HvW_ym6D0BGig1oOn3KXZBN&gclid=CjwKCAjwo4rCBhAbEiwAxhJlCa6bRrMRcCo4rTvxgzxWA4fbHmJfMOLIe8IxrDwpIevZTVKHupsB1hoCesoQAvD_BwE) for versioning the bucket. 
- "lakeFS uses metadata to manage data versions."
- "lakeFS supports data in all object stores including all major cloud providers S3..."

flows/aggregate_inference_results.py
olaughter: That's a good idea, I'd lean towards us waiting until the sizes become a problem and then switching to something like that once it hurts? I also expect the "latest" to be a tiny fraction of this before long as it would match the size of every full backfill we run

PR #461, opened 2025-06-03
Automatically update the vibe checker every night

harrisonpim:
at the request of the policy team! This PR:

- makes some small tweaks to the vibe checker build itself 
  - list of concepts to include comes from the `prod.yaml`
  - only shows a single classifier per concept, from the `ClassifierFactory` (NB might update this to use a few more variants in the future, if useful)
  - now running `.predict()` with each classifier in the script itself, on a sample of 250k passages
- adds add the vibe checker to the list of auto nightly rebuilt/deployed static sites

the whole process (including all predictions and the sync to s3) only takes a couple of mins to run on my laptop, so this shouldn't be too difficult for prefect to handle 🤞


mpjuhasz (approved):
LGTM! One open question on the backward compatibility of the hash change, otherwise great!

PR #465, opened 2025-06-04
Feature/pla 629 indexing doesnt handle  translated suffix

THOR300:
This Pull Request: 
---
- Is a bug fix for the indexer. 
- Objects in the unique run prefix are not document ids as they can contain the suffix `translated_en`. 
- This PR refactors the indexing functionality to run on `DocumentStem`'s and to make use of the `remove_translated_suffix` function to return a `DocumentId`.


olaughter (approved):
Looks good, although I think the naming distinction is a bit of a distraction, lets go with it for now

PR #466, opened 2025-06-04
fix: remove invalid memory value

olaughter:
This was already fixed in 805788d41ab7046919a9108adc9848d5eaf1a691 but somehow made its way back in


THOR300 (approved):
LGTM

PR #473, opened 2025-06-05
fix: Representing indexing all as appropriate Prefect state

jesse-c:
# Solution 2

From the discussion in Slack (see: issue), we're reducing the number of Prefect calls via removing flow/task approaches.

This is to reduce the number of Prefect API calls so we're less likely to be rate-limited. The downside being that we lose observability.

# Solution 1

Based on Prefect[^1], this is always being shown as a success state. This is misleading to us when evaluating the run. It should appropriately be a failed state too.

This begs the question of, are straying too far from the light of Prefect, and shouldn't be passing data in a failed flow run state?

FIXES PLA-640

[^1]: https://docs.prefect.io/v3/develop/manage-states#manage-states

### DIFF
--- a/flows/index_from_aggregate_results.py
+++ b/flows/index_from_aggregate_results.py
@@ -39,7 +39,7 @@ from flows.boundary import (
     get_document_passages_from_vespa__generator,
     get_vespa_search_adapter_from_aws_secrets,
 )
-from flows.result import Err, Error, Ok, Result, unwrap_err
+from flows.result import Err, Error, Ok, Result
 from flows.utils import (
     AsyncProfiler,
     Profiler,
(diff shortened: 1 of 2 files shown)


jesse-c:

- #472
    - #473 ◀

olaughter (approved):

flows/index_from_aggregate_results.py
```diff
@@ -299,7 +300,6 @@ async def create_indexing_summary_artifact(
     )
 
 
-@flow(log_prints=True)
```
THOR300, line 302: note: I'd personally be fine with this still being a flow, by removing the other two we'd be at a heathy third of the current levels and could always remove later.

THOR300 (approved):
Great! Just one non blocking comment 🙌

flows/index_from_aggregate_results.py
jesse-c: ✓

PR #475, opened 2025-06-05
feat: Also count number of passages in audit script

olaughter:
As a passage can have multiple spans the span count doesnt tell us the count of passages. This adds that number to help with understanding the state of vespa


THOR300 (approved):
LGTM

PR #477, opened 2025-06-06
refactor: run aggregations in batches

olaughter:
Most of the tasks are so fast that there was diminsihing returns on running them one at a time

### DIFF
--- a/flows/aggregate_inference_results.py
+++ b/flows/aggregate_inference_results.py
@@ -3,7 +3,7 @@ import json
 import os
 from collections import defaultdict
 from dataclasses import asdict, dataclass
-from typing import Any, TypeAlias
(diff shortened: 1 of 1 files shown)


THOR300 (commented):
Obviously this reduces the start up time of the tasks but why not add a deployment and invoke that so we run on multiple hosts? 

Theoretically ecs scales horizontally so we could pass n docs into each deployment run and run like 200 containers at once? Unlike indexing there's no database that can't tolerate the load!

olaughter:
> Obviously this reduces the start up time of the tasks which is great! But why not add a deployment and invoke that so we run on multiple hosts?

Sounds like a good next step! I am interested to see what the limit of one host is first and also think through some of the different options for spreading it, but yes lets follow up with some multi host runs soon

PR #480, opened 2025-06-06
refactor: make each document a task again

olaughter:
I moved this to the batch level because I thought it would make more sense visually in prefect, but I've since realised I could have my cake and eat it by having tasks at both levels


THOR300 (approved):
LGTM

PR #481, opened 2025-06-06
feat: no longer require run identifier for audits

olaughter:
Since merging 9f07055daab2f0656760fc0a0b8e88526cde5161 we can now review "latest" for this by default


THOR300 (approved):
🙌 Let's go!

PR #482, opened 2025-06-07
refactor: Use async calls for aggregator s3 interactions

olaughter:
This draws on the aioboto3, a wrapper for boto that runs aws interactions async. The refactor is minimal but I've only applied this to the aggregator for now to air out any strangeness. If this works well we should roll it out as standard elsewhere.

### DIFF
--- a/flows/aggregate_inference_results.py
+++ b/flows/aggregate_inference_results.py
@@ -5,20 +5,21 @@ from collections import defaultdict
 from dataclasses import asdict, dataclass
 from typing import Any, Sequence, TypeAlias
 
-import boto3
+import aioboto3
 from botocore.exceptions import ClientError
 from prefect import flow, task
 from prefect.artifacts import create_table_artifact
 from prefect.context import get_run_context
 from prefect.exceptions import MissingContextError
 from prefect.task_runners import ConcurrentTaskRunner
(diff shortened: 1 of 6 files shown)


flows/aggregate_inference_results.py
```diff
@@ -204,40 +205,41 @@ async def process_single_document(
...
-            key=os.path.join(
-                config.aggregate_inference_results_prefix,
-                run_output_identifier,
-                f"{document_id}.json",
-            ),
-        )
-        await asyncio.to_thread(
-            s3_object_write_text, str(s3_uri), json.dumps(vespa_concepts)
-        )
+        session = aioboto3.Session(region_name="eu-west-1")
```
THOR300, line 208: **thought**: Instantiate this higher and pass it in to the document level process so we can't create n sessions?

THOR300 (approved):
LGTM

flows/aggregate_inference_results.py
olaughter: Alrighty, I'll follow up separately with a change for that. Whats your thinking on this being a problem?

jesse-c (commented):
Agreed on using this going forward to if it works well.

PR #484, opened 2025-06-07
fix: Don't load ~80 docs into memory at a once

olaughter:
The control flow for this was loading into memory each labelled_passage file for all inference outputs. Meaning we had doc_size * num_classifiers worth of memory being utilised.

This refactors the way the combining happens to combine one document at a time to the full list. The goal being to only have the current document and the full document open at once, fixing the memory usage regardless of classifier count.

### DIFF
--- a/flows/aggregate_inference_results.py
+++ b/flows/aggregate_inference_results.py
@@ -1,9 +1,8 @@
 import asyncio
 import json
 import os
-from collections import defaultdict
 from dataclasses import asdict, dataclass
-from typing import Any, Sequence, TypeAlias
+from typing import Any, AsyncGenerator, Sequence, TypeAlias
 
 import aioboto3
 from botocore.exceptions import ClientError
@@ -122,9 +121,8 @@ async def get_all_labelled_passages_for_one_document(
     document_id: DocumentImportId,
     classifier_specs: list[ClassifierSpec],
     config: Config,
-) -> dict[SpecStr, list[LabelledPassage]]:
+) -> AsyncGenerator[tuple[ClassifierSpec, list[LabelledPassage]], None]:
     """Get the labelled passages from s3."""
-    labelled_passages = defaultdict(list)
 
     for spec in classifier_specs:
         s3_uri = S3Uri(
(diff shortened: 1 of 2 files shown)


flows/aggregate_inference_results.py
```diff
@@ -207,13 +174,45 @@ async def process_single_document(
...
+                labelled_passages,
+            ) in get_all_labelled_passages_for_one_document(
                 s3, document_id, classifier_specs, config
-            )
-            print(
-                f"Combining {len(all_labelled_passages)} labelled passages for {document_id}"
-            )
-            vespa_concepts = combine_labelled_passages(all_labelled_passages)
+            ):
+                # Validation not needed when its the first spec
```
jesse-c, line 186: **polish:** Could you add _why_ it's not needed?

flows/aggregate_inference_results.py
```diff
@@ -207,13 +174,45 @@ async def process_single_document(
...
+                    )
+
+                for passage, text_block_id in zip(
+                    labelled_passages, concepts_for_vespa.keys()
+                ):
+                    if passage.id != text_block_id:
+                        raise ValueError(
+                            f"The text_block id diverges for {spec} when compared with what has been collected so far:"
+                            f"{passage.id=} != {text_block_id=}"
+                        )
```
jesse-c, line 208: **praise:** Nice to see verbose reasons around like this.

flows/aggregate_inference_results.py
```diff
@@ -240,6 +239,10 @@ async def process_single_document(
                 ),
             )
             return document_id
+    except ValueError as e:
+        raise AggregationFailure(
+            document_id=document_id, exception=e, context="ValueError"
```
jesse-c, line 244: **question:** I wonder if this should capture _any_ exception, and not just `ValueError` ones?

jesse-c (approved):
LGTM

Another reasonable pattern to do elsewhere too.

flows/aggregate_inference_results.py
olaughter: Good call 👍

flows/aggregate_inference_results.py
olaughter: 🙇

flows/aggregate_inference_results.py
olaughter: Yeah that makes sense 👍

PR #485, opened 2025-06-07
fix: Initialise session at batch level

olaughter:
Rather than reinitialising the session for every document.

[As suggested here](https://github.com/climatepolicyradar/knowledge-graph/pull/482#discussion_r2133766947)


THOR300 (approved):
LGTM

PR #486, opened 2025-06-07
refactor: revise defaults aggregate params after runs

olaughter:
The result of having ran in production a few times trying out various configurations. These lower batch size is nice ones to work with as it reports back faster now we're actually running async. And the concurrent tasks don't seem to make the overall run faster over this amount.


THOR300 (approved):
LGTM

PR #492, opened 2025-06-10
chore: Disable some classifier specs.

jesse-c:
The web-app relies on this file to know which to make available.

FIXES PLA-655


THOR300 (approved):
LGTM - checked the classifiers against staging and confirmed that labs / staging don't need updating 👍

olaughter:
Occurs to me that we should hang on to the staging ones so policy can keep testing them there?

jesse-c:
> Occurs to me that we should hang on to the staging ones so policy can keep testing them there?

✓

olaughter (approved):

PR #495, opened 2025-06-16
fix: Make task hashable

jesse-c:
I saw this a lot[^1][^2] and it created a lot of noise, and may also cause  actual problems. I don't think the latter would be true though, as we don't use cached results, explicitly at least.

The session was successfully passed to the task, and used within it. E.g. you can see the 404s for objects not existing. It doesn't show up in the args though in Prefect's webapp.

**Test**

- https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/068500ad-4c45-7734-8000-d58eb360ffef?entity_id=068500b4-27b2-723d-8000-0b5cb49b3953 — Ran went okay without the cache function error.

TOWARDS PLA-658

[^1]:
```
Error encountered when computing cache key - result will not be persisted.
  File "/usr/local/lib/python3.10/site-packages/prefect/cache_policies.py", line 374, in compute_key
    return hash_objects(hashed_inputs, raise_on_failure=True)
  File "/usr/local/lib/python3.10/site-packages/prefect/utilities/hashing.py", line 89, in hash_objects
    raise HashError(msg)
prefect.exceptions.HashError: Unable to create hash - objects could not be serialized.
  JSON error: Unable to serialize unknown type: <class 'aioboto3.session.Session'>
  Pickle error: cannot pickle '_asyncio.Task' object

The above exception was the direct cause of the following exception:

  File "/usr/local/lib/python3.10/site-packages/prefect/task_engine.py", line 169, in compute_transaction_key
    key = self.task.cache_policy.compute_key(
  File "/usr/local/lib/python3.10/site-packages/prefect/cache_policies.py", line 214, in compute_key
    policy_key = policy.compute_key(
  File "/usr/local/lib/python3.10/site-packages/prefect/cache_policies.py", line 384, in compute_key
    raise ValueError(msg) from exc
ValueError: Unable to create hash - objects could not be serialized.
  JSON error: Unable to serialize unknown type: <class 'aioboto3.session.Session'>
  Pickle error: cannot pickle '_asyncio.Task' object

This often occurs when task inputs contain objects that cannot be cached like locks, file handles, or other system resources.

To resolve this, you can:
  1. Exclude these arguments by defining a custom `cache_key_fn`
  2. Disable caching by passing `cache_policy=NO_CACHE`
```
[^2]: https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/0684fda2-449e-7cdb-8000-732e3b78df1a?entity_id=0684fdc6-8809-7f22-8000-928b99be820a

### DIFF
--- a/tests/flows/test_aggregate_inference_results.py
+++ b/tests/flows/test_aggregate_inference_results.py
@@ -21,6 +21,7 @@ from flows.aggregate_inference_results import (
     process_single_document,
     validate_passages_are_same_except_concepts,
 )
+from flows.utils import DocumentImportId
 from scripts.cloud import ClassifierSpec
 from scripts.update_classifier_spec import write_spec_file
 from src.labelled_passage import LabelledPassage
@@ -268,7 +269,7 @@ async def test_process_single_document__value_error(
     keys, bucket, s3_async_client = mock_bucket_labelled_passages_large
     _, classifier_specs = mock_classifier_specs
 
-    document_id = "CCLW.executive.10061.4515"
+    document_id = DocumentImportId("CCLW.executive.10061.4515")
 
     # Replace the doc with a broken  one
     new_data = [
@@ -284,7 +285,7 @@ async def test_process_single_document__value_error(
 
     session = aioboto3.Session(region_name=test_aggregate_config.bucket_region)
     result = await asyncio.gather(
-        process_single_document(
+        process_single_document.fn(
             session,
             document_id,
             classifier_specs,
--- a/flows/aggregate_inference_results.py
+++ b/flows/aggregate_inference_results.py
@@ -2,9 +2,11 @@ import asyncio
 import json
 import os
 from dataclasses import asdict, dataclass
+from functools import partial
 from typing import Any, AsyncGenerator, Sequence, TypeAlias
 
 import aioboto3
+import prefect.tasks as tasks
 from botocore.exceptions import ClientError
 from prefect import flow, task
 from prefect.artifacts import create_table_artifact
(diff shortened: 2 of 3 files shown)


tests/flows/test_aggregate_inference_results.py
```diff
@@ -268,7 +269,7 @@ async def test_process_single_document__value_error(
     keys, bucket, s3_async_client = mock_bucket_labelled_passages_large
     _, classifier_specs = mock_classifier_specs
 
-    document_id = "CCLW.executive.10061.4515"
+    document_id = DocumentImportId("CCLW.executive.10061.4515")
```
jesse-c, line 272: **note:** Fixes a type warning

jesse-c:
> Thanks for this!
> 
> This is super interesting! At first reading I was worried we were just returning the prior cached result as opposed to actually running the task. But reading the docs [here](https://docs.prefect.io/v3/concepts/caching#cache-key-functions) it looks like we only return the cached result as default if the task is invoked from the same flow context more than once.
> 
> > This implies that, by default, calling the same task with the same inputs more than once within a flow will result in cached behavior for all calls after the first.
> 
> Seeing as we don't want to return cached results, do we need to persist the results at all? Wondering whether [`persist_result=False`](https://docs.prefect.io/v3/concepts/caching#cache-key-functions) would be a valid addition to ensure that this isn't happening, looking at the codebase if it's not set it [defaults to `True`](https://github.com/PrefectHQ/prefect/blob/5486631396a72284b1035fd98e50dc6c837d2cd0/src/prefect/tasks.py#L507).

Since we don't use the results, we could turn off the default persistence, but I also don't _currently_ know of any downsides (i.e. costs) of them being on. Prefect has some retention policies so the results are deleted, IIRC.

I do think that for idempotency, persisting results could be a good thing, by my default interpretation of idempotency, that is the same inputs have the same outputs. E.g. if we run inference aggregation on `v8` outputs, that should never change .. but we do overwrite `v8` outputs when doing inference itself, so it does reveal where we may need changes (or to make space for) idempotency!

PR #496, opened 2025-06-16
fix: Correctly report overall inference aggregation state

jesse-c:
At the moment, in Prefect the state will be completed (green) instead of failed (red), if there was ≥ 1 failures.

Now instead of the intermediate `process_n_documents`, which wasn't reporting the appropriate overall state, it's been removed. Prefect's task runner is now used _properly_.

There's room for future work to add sub-deployment runs to get this running across different machines.

FIXES PLA-644


flows/aggregate_inference_results.py
```diff
@@ -1,9 +1,9 @@
-import asyncio
 import json
 import os
+from collections.abc import AsyncGenerator, Sequence
```
jesse-c, line 3: **note:** The previous imports were deprecated.

flows/aggregate_inference_results.py
```diff
@@ -182,12 +181,21 @@ def task_input_hash(
...
+    document_id = parameters.get("document_id", "unknown")
+    slug = generate_slug(2)
+    return f"aggregate-single-document-{document_id}-{slug}"
+
+
 @task(
     cache_key_fn=partial(task_input_hash, ["session"]),
+    task_run_name=task_run_name,
+    retries=1,
+    persist_result=False,
```
jesse-c, line 194: **note:** I believe this means the `cache_key` param won't be needed, but I haven't confirmed that, so I've left it for now.

flows/aggregate_inference_results.py
```diff
@@ -391,25 +372,23 @@ async def aggregate_inference_results(
         f"{len(classifier_specs)} classifiers, outputting to {run_output_identifier}"
     )
 
-    semaphore = asyncio.Semaphore(max_concurrent_tasks)
-
-    # Create tasks for each document
-    tasks = [
-        wait_for_semaphore(
```
jesse-c, line 398: **note:** The [mixed results run](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/06852db9-1066-789a-8000-6bce67322cd8?entity_id=2fe59b1e-c0f6-463d-b22b-0b10c76c88cd), and another [experiment of mine](https://climate-policy-radar.slack.com/archives/C078S46DJFP/p1749665828525009), show how using `.map` does have an interleaved approach to concurrency.

THOR300:
> > Super clean PR!
> > To confirm are we not going to split this out across multiple hosts via sub deployments then?
> > If we are then the original batching strategy could be useful for this and then we use the presented method of mapping task within the sub deployments / at the host level.
> 
> That's what the new issue I made, from standup this morning, will be for!

Lovely stuff 🙌

PR #501, opened 2025-06-19
refactor: Add reporting for inference batches

jesse-c:
It's hard to debug otherwise.

**Tests**

I did a Prefect [run](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/0685438f-9d78-7b7c-8000-d5a63dfe6dba?g_range=%7B%22type%22%3A%22span%22%2C%22seconds%22%3A-604800%7D&entity_id=7faa08bb-9e58-4be8-b7de-1837cdeb6fdf&state=cancelled&state=cancelling&state=completed&state=crashed&state=failed&state=paused&state=pending&state=running&state=scheduled&entity=flowRuns&entity=taskRuns&entity=events&entity=logs&entity=artifacts) and got a top-level artifact and also a batch-level artifact!

TOWARDS PLA-658

### DIFF
--- a/flows/index_from_aggregate_results.py
+++ b/flows/index_from_aggregate_results.py
@@ -5,7 +5,7 @@ import tempfile
 from collections import Counter
 from collections.abc import Awaitable
 from dataclasses import dataclass
-from typing import Any, Final, TypeVar
+from typing import Any, Final
 
 import boto3
 import httpx
@@ -47,6 +47,7 @@ from flows.utils import (
     collect_unique_file_stems_under_prefix,
     iterate_batch,
     remove_translated_suffix,
+    return_with_id,
     wait_for_semaphore,
 )
 from scripts.cloud import AwsEnv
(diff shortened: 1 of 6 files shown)


THOR300 (approved):
LGTM

jesse-c:
> Really nice, perhaps consider adding a test for the artifact before merge? There is one approach for doing so [here](https://github.com/climatepolicyradar/knowledge-graph/blob/main/tests/flows/test_aggregate_inference_results.py#L114-L116) that just grabs the artifact after the run meaning it can then be reviewed

I've done that, but, I don't think it's a good idea as it means we're not using the Prefect test harness, and also because it adds dependency on the state of Prefect outside of a test. E.g. just checking for the length of artifacts to be greater than 1, since there may be 2 artifacts in there, where each 1 came from 2 different tests. I'm 75% confident on this ATM.

PR #506, opened 2025-06-24
rewrite script to index everything into neo4j more efficiently

harrisonpim:
We had an old `neo4j/populate.py` script knocking around, which I had used in the past to set up local versions of the knowledge graph for visualisation etc, or to populate our managed neo4j aura instance with data for the policy team to play around with. That script was _painfully_ slow.

This PR rewrites the script to run uploads in batches, which _massively_ speeds up the indexing. There's still more we could do here (adding more passage/document/concept metadata to the uploaded nodes, etc) but for now, these changes enabled me to index a complete version of the KG (5 million nodes, 15 million edges) into aura in a matter of minutes ✨


THOR300 (approved):
Looks good to me, maybe remove the commented code if in favour of documentation or conditional paths etc.

PR #509, opened 2025-06-24
Add a new dedicated Identifier class

harrisonpim:
Adding this PR to kick things off, with the intention to hand over the rest of the work on [PLA-557](https://linear.app/climate-policy-radar/issue/PLA-557/overhaul-of-model-id-to-add-lineage-tracking) to @climatepolicyradar/deng, who will do a much better job than me at the rest of the implementation.

This PR adds a new `Identifier` class, whose `Identifier.generate()` supersedes the `generate_identifier` function with much nicer structures for validation and testing etc. There's still more work to do here (broken down in the top-level ticket) but I hope sets some sort of direction/intent for the subsequent bits of work 🙇‍♂️

### DIFF
--- a/scripts/build_dataset/02_download_litigation.py
+++ b/scripts/build_dataset/02_download_litigation.py
@@ -26,7 +26,7 @@ from rich.console import Console
 from rich.progress import track
 
 from scripts.config import raw_data_dir
-from src.identifiers import generate_identifier
+from src.identifiers import Identifier
 
 console = Console()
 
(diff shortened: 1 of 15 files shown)


THOR300 (approved):
This looks great, reading over the notion linked on the ticket I see this objective: 

> Classifiers already have a [[deterministic ID](https://github.com/climatepolicyradar/knowledge-graph/blob/main/src/classifier/classifier.py#L115)](https://github.com/climatepolicyradar/knowledge-graph/blob/main/src/classifier/classifier.py#L115) that is created from the classifier type (aka class name), the Wikibase ID, the preferred label, description, definition, and labels. This takes the form of a hash, such as `“2sgknw32”`. Using this instead means we can pass around less data, have flatter folder structures, lookup classifiers more easily and consistently, confirm model consistency across environments, and make our folder structures & code agnostic of what kind of classifier is being used. 
> 
> All of our folder structures would instead look like this: `<concept>/<id>` , e.g: `Q111/gg7h2j2s` .


So just to confirm my understanding could we have achieved this new referencing of classifiers with the deterministic hash and we're implementing this change as it's nicer to work with and better tested?

PR #512, opened 2025-06-30
rename src.argilla_v2 -> src.labelling

harrisonpim:
Part of a series of small spring-cleaning jobs that I'm going to push through before making more substantial additions to the KG codebase. This one is a particularly silly nit-pick! 

I think `argilla_v2.py` is a terrible module name, and it makes me sad every time I have to import it 😭  Where's argilla v1? Why do we only have v2 here? I think this is very unclear for new users of the codebase...

Ideally, I would have renamed this module `src.argilla`, but, because the external argilla library needs to be imported as `argilla`, the block of imports in `src/argilla.py` which call `from argilla import ...` start causing all sorts of circular import and type errors. As a compromise, I've gone with `src.labelling`. 

let me know if you disagree with this change, or have any better naming ideas!

I've also taken the opportunity to delete the task/labeller assignment code, as it's unused.


jesse-c (approved):

jesse-c:
Seems like a reasonable solution to the un-ideal things you described! It's a super thin "abstraction" too, explicitly calling it `labelling`

PR #514, opened 2025-06-30
feat(deps): Use Poetry v2

jesse-c:
It's been out for a while, and I imagine lots of us have updated locally.

I've removed the version pinning too, as we should know from logs, which version was used. This means we can stay up-to-date since these are checked for via Dependabot.

This also updates Prefect.


THOR300 (approved):
LGTM

Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM prefecthq/prefect:3.3.7-python3.10
 
 RUN pip install --upgrade pip
-RUN pip install "poetry==1.8.3"
+RUN pip install "poetry
```
harrisonpim, line 4:
```suggestion
RUN pip install poetry
```

harrisonpim (approved):

Dockerfile
jesse-c: 🙇🏻

PR #520, opened 2025-07-07
Refactor DocumentId to DocumentStem in aggregate inference results pipeline

THOR300:
This Pull Request: 
---
- Refactors the use of `DocumentId` to `DocumentStem` in aggregate inference results KG pipeline. 
- This is required as in the labelled passages pipeline we read in document stems and persist this to the output.
- This means that we have labelled passages with the translated suffix and thus cannot be read via document id. 
- Discussion on any splash damage of the misuse of document ids has been listed on the ticket.

### DIFF
--- a/flows/aggregate_inference_results.py
+++ b/flows/aggregate_inference_results.py
@@ -138,10 +138,6 @@ async def get_all_labelled_passages_for_one_document(
                 config.document_source_prefix,
                 spec.name,
                 spec.alias,
-                # TODO: This is a document stem.
-                #   The labelled_passages s3 prefix contains files with a _translated
-                #   suffix. Inference also identifies stems and persists this when
-                #   saving inference results.
                 f"{document_stem}.json",
             ),
         )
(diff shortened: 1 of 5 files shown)


flows/aggregate_inference_results.py
```diff
@@ -271,40 +270,42 @@ async def process_document(
                     key=os.path.join(
                         config.aggregate_inference_results_prefix,
                         "latest",
-                        f"{document_id}.json",
+                        f"{document_stem}.json",
```
olaughter, line 273: Will we need to rerun aggregation over everything to get everything consistent?

olaughter (approved):
Nicely work! Little sad to see the added complexity flowing from the stem tech debt, but nice to be consistent and unblock things 👍

flows/aggregate_inference_results.py
THOR300: I've put a comment on the ticket, TLDR is that it looks like we did process the translated docs but I'd recommend that we do reprocess everything to make sure. 

https://linear.app/climate-policy-radar/issue/PLA-725/resolve-documentid-documentstem-discrepancies-in-the-aggregate#comment-58913e4a

PR #522, opened 2025-07-08
Feature/pla 729 fix issue with indexing log verbosity

olaughter:
In response to the following error message in prefect:

Screenshot 2025-07-08 at 10 26 33


olaughter:
> The logger vs prints debate continues haha!

Haha, yep! I think I'm coming round to logger now because of being able to set levels. Still keen for us to standardise at some point though, there is [a ticket](https://linear.app/climate-policy-radar/issue/PLA-718/standardise-on-whether-to-use-print-or-logger) to fix some agreement on our preference

PR #524, opened 2025-07-09
Initial commit looking at returning results from batch level inference.

THOR300:


flows/inference.py
```diff
@@ -116,6 +118,48 @@ def to_json(self) -> dict:
...
+
+        return self.failed_document_stems != set()
+
+
+class InferenceException(Exception):
+    """Exception raised when inference fails."""
+
+    def __init__(self, message: str, data: dict[str, Any]):
+        self.message = message
+        self.data = data
```
THOR300, line 151: Is it good practice or is there a better way to provide this data?

flows/inference.py
```diff
@@ -713,7 +774,7 @@ async def classifier_inference(
     config: Config | None = None,
     batch_size: int = INFERENCE_BATCH_SIZE_DEFAULT,
     classifier_concurrency_limit: PositiveInt = CLASSIFIER_CONCURRENCY_LIMIT,
-) -> Sequence[DocumentStem]:
+) -> State:
```
THOR300, line 777: Does this return type need to be more specific?

flows/inference.py
```diff
@@ -796,63 +857,110 @@ async def classifier_inference(
...
-        )
+    inference_result = InferenceResult(
+        batch_inference_results=batch_inference_results,
+        unexpected_failures=unexpected_failures,
+    )
 
-    return filtered_file_stems
+    if (
+        any([result.failed for result in batch_inference_results])
+        or unexpected_failures
```
THOR300, line 886: Use failed property.

flows/inference.py
```diff
@@ -116,6 +118,69 @@ def to_json(self) -> dict:
...
+    successful_document_stems: set[DocumentStem]
+    failed_document_stems: set[tuple[DocumentStem, Exception]]
+    classifier_name: str
+    classifier_alias: str
+
+    @property
+    def failed(self) -> bool:
+        """Whether the batch failed, True if failed."""
+
+        return self.failed_document_stems != set()
```
THOR300, line 149: Do we need this, can't we just get the State object from the deployment result and then use that to decide whether the batch failed or not rather than an inferred value.

flows/inference.py
```diff
@@ -796,63 +878,108 @@ async def classifier_inference(
             *tasks, return_exceptions=True
         )
 
-    failures, successes = group_inference_results_into_states(results)
+    unexpected_failures: list[BaseException] = [
+        result for result in results if isinstance(result, BaseException)
+    ]
+    flow_runs: list[FlowRun] = [
+        result for result in results if not isinstance(result, BaseException)
+    ]
```
THOR300, line 886: Is this right, should we not check for `isinstance(result, FlowRun)`

flows/inference.py
```diff
@@ -649,12 +714,29 @@ async def run_classifier_inference_on_batch_of_documents(
...
+            f"{len(results)} documents."
+        )
+        return Failed(
+            message=message,
+            data=BatchInferenceException(
+                message=message,
+                data=batch_inference_result.model_dump(),
+            ),
+        )
+    return Completed(
```
THOR300, line 736: Update tests to check for these results.

THOR300:
Prefect run (running): 
- https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/068752a4-6a8d-7e4c-8000-cb06b2c3e1cb?entity_id=068752a4-6a8d-7e4c-8000-cb06b2c3e1cb

flows/inference.py
```diff
@@ -116,6 +119,143 @@ def to_json(self) -> dict:
...
+
+    def __init__(self, message: str, data: dict[str, Any]):
+        self.message = message
+        self.data = data
+
+
+class InferenceResult(BaseModel):
+    """Result from running inference on all batches of documents."""
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
olaughter, line 170: **question:** how come this is needed?

flows/inference.py
```diff
@@ -116,6 +119,143 @@ def to_json(self) -> dict:
...
+        self.data = data
+
+
+class BatchInferenceResult(BaseModel):
+    """Result from running inference on a batch of documents."""
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    successful_document_stems: set[DocumentStem]
+    failed_document_stems: set[tuple[DocumentStem, Exception]]
```
olaughter, line 142: **question:** Do these need to be converted into sets here? If deduplication is a risk, would we want to do that earlier then inference?

flows/inference.py
```diff
@@ -556,13 +696,13 @@ async def create_inference_on_batch_summary_artifact(
     )
 
 
-@flow(log_prints=True)
+@flow(log_prints=True, persist_result=True)
```
olaughter, line 699: **concern:** I could do with some background behind this. Using `persist_results` seems to be leading to flakiness (i.e. the first of the two staging runs), I'm also seeing the docstring of  `utils.py::get_deployment_results` saying its required for it to work which is an entanglement that makes me wonder if its the right choice.  We haven't needed to use this elsewhere, and my understanding of this feature was to avoid having to reload data in tasks, so I'm not clear on why we want to persist the inference document stems this way?

olaughter (commented):
Looking good, did have some points to discuss below

flows/inference.py
olaughter: Thanks, I'm not asking for them to be lists though. More wondering if we want to ensure uniqueness whether this is happening too late, and that perhaps they should be sets earlier?

flows/inference.py
THOR300: The thinking is that using the below syntax is nice for marking a flow as failed or completed whilst also returning the resultant data: 

```python
Completed(
        message="Successfully ran inference on all batches!",
        data=inference_result.model_dump(),
)

Failed(
    message=message,
    data=InferenceException(
        message=message,
        data=inference_result.model_dump(),
    ),
)
```

THOR300: Thinking of some options: 
- Persist results to s3. 
- Upon failure in the batch inference run raise a custom exception with the result data as an attribute. This can be caught in the parent flow and the result of the batch inference flow will still be considered failed.

olaughter: Feels like the complexity might be expanding if we need to write flow state to s3 😬 I think @jesse-c has done this elsewhere without having to do that and might be able to advise

jesse-c: I've made some comments on the [accompanying Slack thread](https://climate-policy-radar.slack.com/archives/C078S46DJFP/p1752591392268049?thread_ts=1752509565.186829&cid=C078S46DJFP).

flows/inference.py
```diff
@@ -756,47 +875,62 @@ def parameters(
                 counter=classifier_concurrency_limit,
                 batches=batches,
                 parameters=parameters,
-                unwrap_result=False,
+                unwrap_result=True,
```
THOR300, line 878: Update map_as_sub_flow to handle async functionality.

flows/inference.py
THOR300: Following the discussion with prefect I will be utilising this approach and persisting deployment results to s3 as outlined here: 
- https://docs.prefect.io/v3/advanced/results#result-storage

The entanglement of needing the correct params to be set on the flow is valid and potentially we can check the decorator params for the flow before triggering 👍

flows/aggregate.py
```diff
@@ -539,7 +539,7 @@ def parameters(batch: Sequence[DocumentStem]) -> dict[str, Any]:
         counter=n_batches,
         batches=batches,
         parameters=parameters,
-        unwrap_result=True,
+        unwrap_result=False,
```
THOR300, line 542: This fails now we have fixed `map_as_sub_flow ` to await the async tasks. We don't need the result from the sub deployment so let's just not unwrap. 

Alternatively we could persist results from the sub deployments of `aggregate_batch_of_documents` but again we don't actually need the result.

flows/index.py
```diff
@@ -753,7 +753,7 @@ def parameters(batch: Sequence[DocumentStem]) -> dict[str, Any]:
         counter=indexer_concurrency_limit,
         batches=batches,
         parameters=parameters,
-        unwrap_result=True,
+        unwrap_result=False,
```
THOR300, line 756: Same as: https://github.com/climatepolicyradar/knowledge-graph/pull/524/files#r2222969668

flows/inference.py
```diff
@@ -118,6 +122,111 @@ def to_json(self) -> dict:
...
+
+    @cached_property
+    def failed_document_stems(self) -> set[DocumentStem]:
+        """The set of document stems that failed to be processed."""
+
+        return set(
+            document_stem
+            for batch_inference_result in self.batch_inference_results
+            for document_stem, _ in batch_inference_result.failed_document_stems
+        )
```
THOR300, line 227: Could we pass in all the document stems that processing was called on and use that to get the successful and failed stems?

flows/inference.py
```diff
@@ -65,6 +67,8 @@
 
 CLASSIFIER_CONCURRENCY_LIMIT: Final[PositiveInt] = 20
 INFERENCE_BATCH_SIZE_DEFAULT: Final[PositiveInt] = 1000
+AWS_ENV: str = os.environ["AWS_ENV"]
+S3_BLOCK_RESULTS_CACHE: str = f"s3-bucket/cpr-{AWS_ENV}-prefect-results-cache"
```
THOR300, line 71: Ticketed to get this from the default job variables instead: 
- https://linear.app/climate-policy-radar/issue/PLA-755/get-s3-block-for-results-cache-from-job-variables

flows/inference.py
THOR300, line 227: This class could be simplified by removing the failed_document_stems method as well as the classifier fields as these aren't currently used (were added and then the code refactored). They may prove useful soon though?

flows/utils.py
```diff
@@ -509,12 +510,18 @@ async def map_as_sub_flow(
...
+                                # on the type system, instead of doing `False`
+                                # and then allowing for a union of types in
+                                # the return.
+                                raise_on_failure=True,
+                            )
+                            if not fn_is_async
+                            else await result.state.result(
+                                raise_on_failure=True,
+                            )
                         )
```
jesse-c, line 525: **polish:** To avoid drift between the function calls and commentary:

```suggestion
                        result_fn = partial(
                        	result.state.result, 
                        	#  Doing it this way, makes it easier to rely
                          # on the type system, instead of doing `False`
                          # and then allowing for a union of types in
                          # the return.
                        	raise_on_failure=True,
                        )
                        flow_result: U = await result_fn() if fn_is_async else result_fn())
```

pyproject.toml
```diff
@@ -42,6 +42,7 @@ requests = "^2.32.3"
 aioboto3 = "^14.3.0"
 types-aioboto3 = {extras = ["s3"], version = "^14.3.0"}
 pytest-aioboto3 = "^0.6.0"
+prefect-aws = "^0.5.12"
```
jesse-c, line 45: **fix:** This is the deprecated way of bringing this in, IIRC. Instead, do something like[^1]:

```
prefect = {extras = ["aws", "slack"], version = "^3.2.15"}
```
[^1]: https://github.com/climatepolicyradar/orchestrator/blob/main/pyproject.toml#L11

flows/full_pipeline.py
```diff
@@ -145,25 +147,25 @@ async def full_pipeline(
...
     )
 
-    if isinstance(inference_result, Exception):
+    if isinstance(inference_result_raw, Exception):
         logger.error("Inference failed.")
-        raise inference_result
-    logger.info(f"Inference complete. Document stems count: {len(inference_result)}")
+        raise inference_result_raw
+
+    inference_result: InferenceResult = InferenceResult(**inference_result_raw)
```
jesse-c, line 159: **question:** 
> This is such that they can be used to further drive the parent full pipeline on only succeeded documents.
Could I check my understanding on how this is implemented.

`inference` can now return `InferenceResult`[^1] for successes or `InferenceException[^2] for at least 1 batch failures.

That means that from this code, and the bit above about re-raising an exception (that would be an `InferenceException`), then the full pipeline would only continue if every batch was successful? Whereas, the quoted line here, and our general approach to pipelines, would want a best-effort continuation of the ones that did succeed?

So the inference pipeline could be appropriately marked as failure, for ≥ 1 failure, and in the full pipeline you'd need to extract the ≥ 1 successes, and then continue on to aggregation with those?

[^1]: 
[^2]: https://github.com/climatepolicyradar/knowledge-graph/pull/524/files#diff-f2fbcdde09b032ff14f38a9893469fb71b65a7250901908d7f07c3a983c6be79R1092-R1099

flows/utils.py
```diff
@@ -455,6 +455,7 @@ async def map_as_sub_flow(
     batches: Generator[Sequence[T], None, None],
     parameters: Callable[[Sequence[T]], dict[str, Any]],
     unwrap_result: bool,
+    fn_is_async: bool = False,
```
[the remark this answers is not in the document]
THOR300: I'm sure there is a solution though that can be found.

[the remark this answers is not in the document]
jesse-c: Nice! There's also:

```
>>> from flows.inference import inference
>>> import inspect
>>> inspect.iscoroutinefunction(inference)
False
>>> inspect.iscoroutinefunction(inference.fn)
True
```

flows/inference.py
```diff
@@ -870,7 +997,7 @@ async def inference(
     config: Config | None = None,
     batch_size: int = INFERENCE_BATCH_SIZE_DEFAULT,
     classifier_concurrency_limit: PositiveInt = CLASSIFIER_CONCURRENCY_LIMIT,
-) -> Sequence[DocumentStem]:
+) -> State:
```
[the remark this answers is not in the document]
THOR300: Yeh but we need to return the successful document stems

flows/inference.py
```diff
@@ -118,6 +122,111 @@ def to_json(self) -> dict:
...
+            for batch_inference_result in self.batch_inference_results
+            for document_stem, _ in batch_inference_result.failed_document_stems
+        )
+
+        return set(
+            document_stem
+            for batch_inference_result in self.batch_inference_results
+            for document_stem in batch_inference_result.successful_document_stems
+            if document_stem not in failed_document_stems
+        )
```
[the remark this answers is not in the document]
jesse-c: You could share the internal implementation.

flows/inference.py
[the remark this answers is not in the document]
jesse-c: Explicitly returning a `State` is reinventing the automatic behaviour of Prefect's workflow engine, AFAIK.

A `Failed` state is automatically created from an exception[^1]. This stuffs the exception into the `data` field automatically[^2].

You can still treat it with normal Pythonic control flow approaches, and rely on Prefect. You've already got `return_state=True` too, so you don't need to explicitly return a `State` object.

[^1]: https://github.com/PrefectHQ/prefect/blob/0317ea2b5c508c980a4809dab11a551c49a64cd0/src/prefect/flow_engine.py#L954
[^2]: https://github.com/PrefectHQ/prefect/blob/main/src/prefect/states.py#L251

flows/full_pipeline.py
THOR300: Yeh I'll do tests on staging again before merge if this is approved. I've added a test for the full pipeline which did catch this and have now updating the code to handle the exceptions!

flows/inference.py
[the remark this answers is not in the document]
THOR300: Ahh nice, looks like we have to raise the exception for this to be a thang!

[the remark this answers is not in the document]
THOR300: Okay nice I've taken a look at this and I've got an action to implement it, do you see it as blocking here or happy to approve seeing as the PR is getting rather large and this would require validating how pydantic types are serialised and returned from the cache etc.

jesse-c (approved):
LGTM. A few of remaining things.

I couldn't imagine trying to use, and understand, Prefect if it wasn't open source!

flows/inference.py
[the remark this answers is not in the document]
THOR300: Ticketed here: 
- https://linear.app/climate-policy-radar/issue/PLA-760/remove-state-return-types-in-inference-pipeline

PR #526, opened 2025-07-09
Mixins for estimating model uncertainty

harrisonpim:
closes SCI-436 SCI-435

This PR extends our classifiers to enable uncertainty estimation, via a set of mixins.

The `LLMClassifier` and `BERTBasedClassifier` have been explicitly extended with these new mixins. Estimating the models' level of confidence in their predictions allows us to identify the passages that they're least confident about, and flag them for manual review. Filtering our data for the hard examples, and skipping the obvious ones should make our active learning process much more efficient for the policy team!

Just like span agreement, model confidence can be assessed at various levels: passage-level, span-level, or a combination of both.

The additions here allow us to run code like

```py
from src.classifier.bert_based import LLMClassifier

concept = wikibase.get_concept("Q382") # flood (Q382)
classifier = LLMClassifier(concept)

text = "When my sink overflowed and my kitchen flooded, i was flooded with emotion"
predictions, uncertainty = classifier.predict_with_uncertainty(
    text, 
    num_samples=10, 
    method="passage"
)

if uncertainty > 0.7:
    print("Low confidence (uncertainty: {uncertainty:.2f})")
    print("send this passage to a human for manual review")
else:
    print(f"High confidence (uncertainty: {uncertainty:.2f})")
    print("save this as a new fine-tuning example for a bert based model")
```

### DIFF
--- a/src/classifier/bert_based.py
+++ b/src/classifier/bert_based.py
@@ -93,9 +93,8 @@ class BertBasedClassifier(Classifier, GPUBoundClassifier, UncertaintyMixin):
 
     def predict_batch(self, texts: list[str]) -> list[list[Span]]:
         """Predict whether the supplied texts contain instances of the concept."""
-        use_dropout = getattr(self, "_use_dropout_during_inference", False)
 
-        if use_dropout:
+        if getattr(self, "_use_dropout_during_inference", False):
             with self._dropout_enabled():
                 predictions = self.pipeline(texts, padding=True, truncation=True)
         else:


src/classifier/bert_based.py
```diff
@@ -65,19 +72,35 @@ def __init__(
...
+            self.model.train(was_training)  # Restore original state when we're done
 
     def predict(self, text: str) -> list[Span]:
         """Predict whether the supplied text contains an instance of the concept."""
         return self.predict_batch([text])[0]
 
     def predict_batch(self, texts: list[str]) -> list[list[Span]]:
         """Predict whether the supplied texts contain instances of the concept."""
-        self.model.eval()
+        use_dropout = getattr(self, "_use_dropout_during_inference", False)
```
THOR300, line 96: **question**: Why use dropout during inference as opposed to during training (assuming we're the ones running training)?

THOR300 (approved):
This looks great, I cannot say that I fully understand the methods used to calculate the uncertainty but approving to unblock 🙌

src/classifier/bert_based.py
jesse-c: NB: As I'm trying to understand this all a bit, here's my attempt at an explanation.

We're not the ones doing training, as this is an off-the-shelf BERT model. Ideally, the dataset created by this whole process will be used to distill/fine-tune our own model.

Keeping dropout on is an implementation of the Monte-Carlo dropout approach [^1].

[^1]: https://www.notion.so/climatepolicyradar/Brief-Uncertainty-based-active-learning-2139109609a480438620eaa0f6754ff1

src/classifier/uncertainty_mixin.py
```diff
@@ -0,0 +1,259 @@
...
+
+    Example:
+        class MyClassifier(Classifier, UncertaintyMixin):
+            # This automatically implements UncertaintyCapableClassifier protocol,
+            # allowing us to safely call predict_with_uncertainty()
+            pass
+    """
+
+
+class Uncertainty(float):
```
jesse-c, line 88: **nitpick:** I was curios if this could be done in a Pydantic-first way, to rely on built-in validators, and to make it fit into other Pydantic models more readily. I've come up with:
```suggestion
class Uncertainty(RootModel[Annotated[float, Field(ge=0.0, le=1.0)]])::
    """
    A validated float representing uncertainty as a number, inclusive of 0.0 and 1.0.

    Lower values represent high confidence, with 0 representing total certainty.
    Higher values represent low confidence, with 1 representing total uncertainty.
    """

    def __str__(self) -> str:
         return str(self.root)
         
    def __repr__(self) -> str:
        return f"Uncertainty({self.root})"
```

src/classifier/uncertainty_mixin.py
```diff
@@ -0,0 +1,259 @@
...
+    SupportsUncertainty, HasUncertaintyMethods, Protocol
+):
+    """
+    Protocol combining classifier prediction and uncertainty calculation capabilities.
+
+    This protocol represents the complete interface for classifiers that can perform
+    uncertainty estimation. It combines:
+
+    1. SupportsUncertainty: Ability to create variants and make predictions
+    2. HasUncertaintyMethods: Uncertainty calculation algorithms
```
jesse-c, line 70: **question:** Would those 2 protocols ever be used separately? If not, why not combine them into one?

src/classifier/uncertainty_mixin.py
jesse-c, line 88: **suggestion:** I'd rename this `Confidence`, since if this value has 0, then it's a certain value, and not an uncertain value, at all. A confidence value incorporates the possibility of being both. Another benefit would be to reduce the frequency of the term _uncertainty_, to ease with understanding this module.

src/classifier/uncertainty_mixin.py
```diff
@@ -0,0 +1,259 @@
...
+        if not predictions:
+            # If we have no predictions, we are certain that the concept is not present
+            return Uncertainty(0.0)
+
+        # Calculate binary prediction consistency (concept present/absent)
+        binary_uncertainty = self._calculate_passage_uncertainty(predictions)
+
+        # If all predictions are negative, uncertainty is based only on binary consistency
+        positive_predictions = [spans for spans in predictions if spans]
+        if len(positive_predictions) < 2:
```
jesse-c, line 185: **question:** The comment says _all_, so I'd expect it to be checking if collated list of spans was empty?
```suggestion
        if not positive_predictions:
```

src/classifier/uncertainty_mixin.py
```diff
@@ -0,0 +1,259 @@
...
+            predictions: List of span predictions from multiple sampling runs
+
+        Returns:
+            Uncertainty score between 0 and 1
+        """
+
+        binary_predictions = [1 if spans else 0 for spans in predictions]
+
+        if len(set(binary_predictions)) <= 1:
+            return Uncertainty(0.0)
```
jesse-c, line 220: ****nitpick:** Use a more FP approach that _I_ feel is more quickly understood. I found the way of using the length of the unique values to be a little _secondary_.

```suggestion
        if all(not spans for spans in predictions):
            return Uncertainty(0.0)

        binary_predictions = [1 if spans else 0 for spans in predictions]
```

src/classifier/uncertainty_mixin.py
jesse-c: Some code golf:

```suggestion
        if not list(more_itertools.flatten(predictions)):
```

jesse-c: There's another check of this [elsewhere](https://github.com/climatepolicyradar/knowledge-graph/pull/526/files#r2219720431), so I imagine I'm just misunderstanding!

src/span.py
```diff
@@ -227,6 +227,35 @@ def jaccard_similarity(span_a: Span, span_b: Span) -> float:
     return intersection / union
 
 
+def jaccard_similarity_for_span_lists(
+    spans_a: list[Span], spans_b: list[Span]
+) -> float:
```
jesse-c, line 232: **suggestion:** Could be good to return a value that is validated for being constrained to [0.0, 1.0]. Fortunately, there already is the `class Uncertainty` that could be used, if it was made a base class for both uncertainty elsewhere, and here, for a _unit_ value.

src/classifier/bert_based.py
```diff
@@ -65,19 +72,35 @@ def __init__(
         # Move model to training device (only used during training)
         self.model.to(self.training_device)
 
-    def __repr__(self):
-        """Return a string representation of the classifier."""
-        return f'{self.name}({self.concept.preferred_label}, base_model="{self.base_model}")'
+    @contextmanager
+    def _dropout_enabled(self):
```
jesse-c, line 76: **question:** Is this needed, since `get_variant_sub_classifier` returns a copy, with dropout being enabled, and that never changing for that copy?

jesse-c (commented):
LGTM. Most likely nothing major in my comments, as the possible issue around count checking may be misunderstanding.

src/classifier/uncertainty_mixin.py
harrisonpim: very nice!

src/classifier/uncertainty_mixin.py
harrisonpim: good question - i think you're right

src/classifier/uncertainty_mixin.py
harrisonpim: actually ended up rolling this back because using a `RootModel` requires that we keep calling `uncertainty.root` to get the actual value, which felt kind of ugly! still curious about more pydantic-first ways to do this though, so i'll keep an eye out 👀

harrisonpim: i think this might just be a misunderstanding - the code is deliberately set up so that higher the value, the more uncertainty is present in the predictions, ie

`Uncertainty(0.0)` means "no uncertainty"
`Uncertainty(1.0)` means "maximum uncertainty"

Renaming the class `Confidence` would necessitate a big refactor (inverting all of the calculations) but wouldn't actually change the meaning/function of the code, afaict!

harrisonpim: point taken on the ease of understanding this module though! might be worth adding more high-level explanation/documentation 🤔

src/classifier/bert_based.py
harrisonpim: yeah, if I remember correctly it's still necessary to manage the switch into training mode for monte-carlo inference! the `_use_dropout_during_inference` property being `True` just determines whether to use the contextmanager, but unless the contextmanager actually switches the model to training mode, dropout would still be disabled.

src/classifier/uncertainty_mixin.py
jesse-c: That's fair enough! 🙇🏻 I didn't think about without an inversion, that the suggested name would have its own confusion.

PR #528, opened 2025-07-14
Feature/remove group inference results function

THOR300:


THOR300:
Post refactor this removal is no longer relevant 🙌

PR #529, opened 2025-07-14
fix: update automation deployment name

olaughter:
I missed automations off the previous change. We changed the flow name and the prefix, but the prefix change wasnt being applied here leading to a 404


THOR300 (approved):
LGTM

PR #531, opened 2025-07-14
feat: Use assets to track aggregation

jesse-c:


flows/aggregate.py
```diff
@@ -393,13 +427,46 @@ async def aggregate_batch_of_documents(
...
+            ).submit(
+                document_stem=document_stem,
+                session=session,
+                classifier_specs=classifier_specs,
+                config=config,
+                run_output_identifier=run_output_identifier,
+            )
+        )
+
+    futures = PrefectFutureList(tasks)
```
jesse-c, line 469: **note:** Annoyingly, we can't use the nice `map()` here since we need to use `with_options()`.

flows/aggregate.py
```diff
@@ -485,4 +552,7 @@ def parameters(batch: Sequence[DocumentStem]) -> dict[str, Any]:
         failures=failures,
     )
 
+    if failures:
```
jesse-c, line 555: **note:** This was a bug, unrelated to this work.

flows/index.py
```diff
@@ -112,6 +115,19 @@ async def _update_vespa_passage_concepts(
     # randomly sample from a pseudo-random distribution and
     # conditionally print this extra info.
     if random.random() < 0.1:
+        # Example:
```
jesse-c, line 118: **note:** A little verbose, but thought it was helpful.

flows/index.py
```diff
@@ -536,15 +615,39 @@ async def index_batch_of_documents(
...
-        config=unmapped(config),
-        run_output_identifier=unmapped(run_output_identifier),
-        indexer_document_passages_concurrency_limit=unmapped(
-            indexer_document_passages_concurrency_limit
-        ),
-        indexer_max_vespa_connections=unmapped(indexer_max_vespa_connections),
-    )
+    # Create Vespa connection inside the task to avoid serialization issues
+    temp_dir = tempfile.TemporaryDirectory()
+    vespa: Vespa = get_vespa_search_adapter_from_aws_secrets(
```
jesse-c, line 620: **note:** This is created just to get access to the function for getting a document V1 path.

flows/inference.py
```diff
@@ -329,36 +344,166 @@ def document_passages(
             yield _stringify(text_block.text), text_block.text_block_id
 
 
-def serialise_labels(labels: list[LabelledPassage]) -> BytesIO:
-    data = [label.model_dump_json() for label in labels]
-    return BytesIO(json.dumps(data).encode("utf-8"))
+T = TypeVar("T", bound=BaseModel)
+
+
+def serialise_pydantic_list_as_jsonl(models: Sequence[T]) -> BytesIO:
```
jesse-c, line 350: **note:** So! Pydantic doesn't have a neat JSONL support. I did try using a Pydantic model that was just a list, but didn't care for that.

It would be nice to use `.jsonl`.

flows/inference.py
```diff
@@ -329,36 +344,166 @@ def document_passages(
...
+
+    First tries JSONL format, then falls back to original format (JSON array of JSON strings).
+    """
+    from pydantic import ValidationError
+
+    # Try JSONL format first
+    try:
+        return deserialise_pydantic_list_from_jsonl(content, model_class)
+    except ValidationError:
+        # Fall back to original format (array of JSON strings)
```
jesse-c, line 390: **note:** This is for PLA-611, so we don't need to re-run all inference.

flows/inference.py
```diff
@@ -329,36 +344,166 @@ def document_passages(
...
+        data = json.loads(content)
+        return [model_class.model_validate_json(passage) for passage in data]
+
+
+def serialise_labels(labels: Sequence[LabelledPassage]) -> BytesIO:
+    """Serialize labelled passages as JSONL format."""
+    return serialise_pydantic_list_as_jsonl(labels)
+
+
+class SingleDocumentInferenceResult(BaseModel):
```
jesse-c, line 400: **note:** A wrapper instead of using a tuple.

flows/inference.py
```diff
@@ -329,36 +344,166 @@ def document_passages(
...
-    s3 = boto3.client("s3", region_name=config.bucket_region)
-    s3.put_object(
-        Bucket=config.cache_bucket,
-        Key=key,
-        Body=body,
-        ContentType="application/json",
-    )
+    s3 = session.client("s3")
+    # Don't get rate-limited by AWS
+    semaphore = asyncio.Semaphore(10)
```
jesse-c, line 449: **note:** An arbitrary number.

flows/inference.py
```diff
@@ -482,21 +624,18 @@ async def run_classifier_inference_on_document(
         )
         doc_labels.append(labelled_passages)
 
-    store_labels(
```
jesse-c, line 485: **note:** Had to do this as I couldn't just easily make this parent function a `@task`/`@materialize` due to rate-limiting. We had tried it being one before.

flows/utils.py
```diff
@@ -435,16 +436,16 @@ async def wait_for_semaphore(
         return await fn
 
 
-async def return_with_id(
-    id: U,
+async def return_with(
```
jesse-c, line 439: **note:** It's now used for more than just IDs.

flows/index.py
```diff
@@ -170,6 +186,36 @@ class SimpleConcept:
...
+) -> str:
+    document_id: DocumentImportId = remove_translated_suffix(document_stem)
+    # Example: /document/v1/doc_search/family_document/docid/CCLW.executive.10014.4470
+    document_v1_path = vespa.get_document_v1_path(
+        id=document_id,
+        schema="document_passage",
+        namespace="doc_search",
+        group=None,
+    )
+    return f"vespa:/{document_v1_path}.x"
```
THOR300, line 216: **question**: So here are we suffixing with `.x` as a placeholder for the full data id that we don't yet know (just reading the comment on line 554)?

THOR300 (approved):
Looks good to me and nice to be integrating new prefect functionality so soon! It would be good to do a little overview at standup to update us all on the UI etc. post merge and run 🙌

jesse-c:
I'm getting 404s when trying to delete the test assets, so I've reached out to Prefect support via email.

PR #534, opened 2025-07-16
feat: spec reader interface for specs with more data

olaughter:
We are moving to having specs with more data included in them. This starts an interface for working with these new file structures. There was plenty of code existing spread out that interfaces with the old files, this begins to consolidate that for the new files.

More work will be needed to update before and after this point in the lifecycle (i.e. training, inference, etc). The basic reading of these files seemed like a good place to start as these tie everything else together.

### DIFF
--- a/flows/classifier_specs/spec_interface.py
+++ b/flows/classifier_specs/spec_interface.py
@@ -1,0 +1,73 @@
+from pathlib import Path
+
+import yaml
+from pydantic import BaseModel, Field
+
+from scripts.cloud import AwsEnv
+from src.identifiers import WikibaseID
+
+SPEC_DIR = Path("flows") / "classifier_specs" / "v2"
+
+
+class ClassifierSpec(BaseModel):
+    """Details for a classifier to run."""
+
+    wikibase_id: WikibaseID = Field(
+        description=(
+            "The wikibase id for the underlying concept being classified. e.g. 'Q992'"
+        ),
+    )
+    classifier_id: str = Field(
+        description=(
+            "The unique identifier for the classifier, built from its internals."
+        ),
+    )
+    wandb_registry_version: int = Field(
+        description=("The version of the classifier in wandb registry. e.g. 1"),
+    )
+    gpu: bool = Field(
+        description=("Whether the classifier should be run on a GPU."),
+        default=False,
+    )
+    only_run_on: list[str] = Field(
+        description=(
+            "The source of documents to classify. If not provided, the classifier will"
+            "be run on all sources. Sources are the first part of a document id, e.g. "
+            "'sabin' in 'sabin.document.1.1', or 'UNFCCC' in 'UNFCCC.document.1.1'"
+        ),
+        default=None,
+    )
+
+    def __hash__(self):
+        """Make ClassifierSpec hashable for use in sets and as dict keys."""
+        return hash((self.wikibase_id, self.classifier_id))
+
+    def __str__(self):
+        """Return a string representation of the classifier spec."""
+        return f"{self.wikibase_id}:{self.classifier_id}"
+
+
+def determine_spec_file_path(aws_env: AwsEnv, spec_dir: Path = SPEC_DIR) -> Path:
+    """Determine the path to the spec file for a given AWS environment."""
+    return spec_dir / f"{aws_env}.yaml"
+
+
+def load_classifier_specs(
+    aws_env: AwsEnv, spec_dir: Path = SPEC_DIR
+) -> list[ClassifierSpec]:
+    """Load classifier specs into python for a given environment."""
+    file_path = determine_spec_file_path(aws_env, spec_dir)
+
+    with open(file_path, "r") as file:
+        contents = yaml.load(file, Loader=yaml.FullLoader)
+
+    classifier_specs = []
+    for spec in contents:
+        classifier_specs.append(ClassifierSpec.model_validate(spec))
+
+    return classifier_specs
+
+
+if __name__ == "__main__":
+    classifier_specs = load_classifier_specs(AwsEnv("sandbox"))
+    print(classifier_specs)
--- a/flows/classifier_specs/spec_interface.py
+++ b/flows/classifier_specs/spec_interface.py
@@ -4,7 +4,7 @@ import yaml
 from pydantic import BaseModel, Field
 
 from scripts.cloud import AwsEnv
-from src.identifiers import WikibaseID
+from src.identifiers import Identifier, WikibaseID
 
 SPEC_DIR = Path("flows") / "classifier_specs" / "v2"
 
@@ -17,7 +17,7 @@ class ClassifierSpec(BaseModel):
             "The wikibase id for the underlying concept being classified. e.g. 'Q992'"
         ),
     )
-    classifier_id: str = Field(
+    classifier_id: Identifier = Field(
         description=(
             "The unique identifier for the classifier, built from its internals."
         ),
--- a/flows/classifier_specs/spec_interface.py
+++ b/flows/classifier_specs/spec_interface.py
@@ -1,4 +1,5 @@
 from pathlib import Path
+from typing import Optional
 
 import yaml
 from pydantic import BaseModel, Field
@@ -11,6 +12,14 @@ SPEC_DIR = Path("flows") / "classifier_specs" / "v2"
 
 class ClassifierSpec(BaseModel):
     """Details for a classifier to run."""
+
+    class ComputeEnvironment(BaseModel):
+        """Details about the compute environment the classifier should on."""
+
+        gpu: bool = Field(
+            description=("Whether the classifier should be run on a GPU."),
+            default=False,
+        )
 
     wikibase_id: WikibaseID = Field(
         description=(
@@ -26,9 +35,9 @@ class ClassifierSpec(BaseModel):
         description=("The version of the classifier in wandb registry. e.g. v1"),
         pattern=r"^v\d+$",
     )
-    gpu: bool = Field(
-        description=("Whether the classifier should be run on a GPU."),
-        default=False,
+    compute_environment: Optional[ComputeEnvironment] = Field(
+        description=ComputeEnvironment.__doc__,
+        default=None,
     )
 
     def __hash__(self):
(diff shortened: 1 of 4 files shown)


flows/classifier_specs/spec_interface.py
```diff
@@ -0,0 +1,73 @@
...
+        """Return a string representation of the classifier spec."""
+        return f"{self.wikibase_id}:{self.classifier_id}"
+
+
+def determine_spec_file_path(aws_env: AwsEnv, spec_dir: Path = SPEC_DIR) -> Path:
+    """Determine the path to the spec file for a given AWS environment."""
+    return spec_dir / f"{aws_env}.yaml"
+
+
+def load_classifier_specs(
```
olaughter, line 55: I'd anticipate this replacing the current `[parse_spec_file](https://github.com/climatepolicyradar/knowledge-graph/blob/c5f022b269092df10e15ce22fd5d5f2d82b2d05e/scripts/update_classifier_spec.py#L38)` eventually

flows/classifier_specs/v2/sandbox.yaml
olaughter, line 1: To be replaced with a generated version next

harrisonpim (approved):

flows/classifier_specs/spec_interface.py
```diff
@@ -0,0 +1,73 @@
...
+        description=(
+            "The wikibase id for the underlying concept being classified. e.g. 'Q992'"
+        ),
+    )
+    classifier_id: str = Field(
+        description=(
+            "The unique identifier for the classifier, built from its internals."
+        ),
+    )
+    wandb_registry_version: int = Field(
```
jesse-c, line 25: **nitpick:**
```suggestion
    wandb_registry_version: NonNegativeInt = Field(
```

jesse-c: FWIW this used to be a [string](https://github.com/climatepolicyradar/knowledge-graph/blob/feat%2Fuse-assets-for-aggregations/scripts/cloud.py#L25-L30), so we're explicitly putting in the constraint to never allow aliases, which can be arbitrary.

flows/classifier_specs/spec_interface.py
```diff
@@ -0,0 +1,73 @@
...
+        default=False,
+    )
+    only_run_on: list[str] = Field(
+        description=(
+            "The source of documents to classify. If not provided, the classifier will"
+            "be run on all sources. Sources are the first part of a document id, e.g. "
+            "'sabin' in 'sabin.document.1.1', or 'UNFCCC' in 'UNFCCC.document.1.1'"
+        ),
+        default=None,
+    )
```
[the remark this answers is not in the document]
jesse-c: Wouldn't it be corpora, and not specific family documents? I guess it being like this makes it more flexible—and I wouldn't be surprised if _someone_ asks for a document to be excluded.

Would this allow for https://github.com/climatepolicyradar/knowledge-graph/blob/feat%2Fuse-assets-for-aggregations/flows/inference.py#L236-L238 to be removed, and put in the classifier specifications files instead?

flows/classifier_specs/spec_interface.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def __hash__(self):
+        """Make ClassifierSpec hashable for use in sets and as dict keys."""
+        return hash((self.wikibase_id, self.classifier_id))
```
jesse-c, line 43: **question:** Wouldn't it only need the `classifier_id`, since presumably with the new fancy new IDs, they include the name + alias/version?

flows/classifier_specs/spec_interface.py
```diff
@@ -0,0 +1,73 @@
...
+            "The unique identifier for the classifier, built from its internals."
+        ),
+    )
+    wandb_registry_version: int = Field(
+        description=("The version of the classifier in wandb registry. e.g. 1"),
+    )
+    gpu: bool = Field(
+        description=("Whether the classifier should be run on a GPU."),
+        default=False,
+    )
```
jesse-c, line 31: **question:** Now that I see this written down, beyond the Notion document. I wonder if this could be changed for 2 reasons:

1. Why not have it as a capabilities thing, such as `capabilities: [cpu, gpu]` for a model that could be run on both? Otherwise, it would just be 1 of them. It feels unlikely, but it makes me think about doing batch inference requests on a GPU, for lots of _smaller_ requests.
2. If we have the size needed, we could save money by dynamically using a specific size? E.g. some models may need only a small GPU, and others a huge one. It could be optional, with a default GPU chosen.

jesse-c (commented):
What do you think it would look like if Policy was able to say "use this model of this other model for this concept"?

Seems reasonable so far!

flows/classifier_specs/spec_interface.py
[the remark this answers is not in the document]
olaughter: Added some examples and written up some thoughts [on the ticket for doing this properly](https://linear.app/climate-policy-radar/issue/PLA-574/spike-what-is-a-reasonable-approach-to-us-being-able-to-handle-corpus#comment-bf96f8fc), I think I will take this out for now though

flows/classifier_specs/spec_interface.py
olaughter: Ah yes, I believe you are right

flows/classifier_specs/spec_interface.py
olaughter: Ah, realising this should stay a string

flows/classifier_specs/spec_interface.py
```diff
@@ -0,0 +1,73 @@
...
+class ClassifierSpec(BaseModel):
+    """Details for a classifier to run."""
+
+    wikibase_id: WikibaseID = Field(
+        description=(
+            "The wikibase id for the underlying concept being classified. e.g. 'Q992'"
+        ),
+    )
+    classifier_id: str = Field(
+        description=(
```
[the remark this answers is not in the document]
olaughter: Check 👍

flows/classifier_specs/spec_interface.py
olaughter: I do like this, but am also hesitant to do too much with it as an idea up front. So I'll limit the current implementation to only specifying if a gpu is needed or not, but refactor to an approach that would be easier to expand later with this in mind

olaughter:
> What do you think it would look like if Policy was able to say "use this model of this other model for this concept"?

I'm wondering if when we revisit the training section we might be able to add some indicator _somewhere_ for this that policy could use, [as raised by Harrison here](https://www.notion.so/What-do-we-want-from-classifier-specs-2329109609a480c18a79de40c18b94cf?d=2329109609a480dc8f5c001c85630e6f&source=copy_link#2329109609a480ad9690c0e20b4dba52). 

There is also an option (depending on what hot swapping turns into), for us to have the ability for multiple classifiers to be processed and indexed for the same concept, but for only one to then be "active" i.e. seen by users when running searches

flows/classifier_specs/spec_interface.py
harrisonpim: I didn't think that was a problem! I've definitely seen that pattern in the wild, and I'm pretty sure i've used it myself 👀

harrisonpim: yep, here we go 

https://github.com/climatepolicyradar/knowledge-graph/blob/e6625fc3cbfd8dc87f4c474ff5b8de97e84b673c/src/span.py#L176-L202

olaughter: Ah interesting, I've also just looked and found examples elsewhere, so perhaps it is fine 😅

harrisonpim: fair!

jesse-c: Should do it in a FP way ;) `envs |> map(cls.load_c_s)`

flows/classifier_specs/spec_interface.py
```diff
@@ -0,0 +1,75 @@
...
+        ),
+    )
+    classifier_id: Identifier = Field(
+        description=(
+            "The unique identifier for the classifier, built from its internals."
+        ),
+    )
+    wandb_registry_version: str = Field(
+        description=("The version of the classifier in wandb registry. e.g. v1"),
+        pattern=r"^v\d+$",
```
jesse-c, line 36: **fix:** Use https://github.com/climatepolicyradar/knowledge-graph/blob/feat%2Fuse-assets-for-aggregations/src/version.py#L4-L7, as it's for W&B versions.

flows/classifier_specs/spec_interface.py
```diff
@@ -0,0 +1,75 @@
...
+    classifier_specs = []
+    for spec in contents:
+        classifier_specs.append(ClassifierSpec.model_validate(spec))
+
+    return classifier_specs
+
+
+if __name__ == "__main__":
+    classifier_specs = load_classifier_specs(AwsEnv("sandbox"))
+    print(classifier_specs)
```
jesse-c, line 75: **fix:** This should be removed before merging? Or is it, and just the sandbox file being present, demonstrative of this being a step in the larger work?

jesse-c (approved):

flows/classifier_specs/spec_interface.py
olaughter: Aha, great point, thank you

flows/classifier_specs/spec_interface.py
olaughter: Ah yes I'll remove this

flows/classifier_specs/spec_interface.py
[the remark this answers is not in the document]
olaughter: Just to close the loop on this, I've also [asked Alan to investigate](https://linear.app/climate-policy-radar/issue/PLA-574/spike-what-is-a-reasonable-approach-to-us-being-able-to-handle-corpus#comment-9a70126d) for us about wether "source" is a viable option for this and updated the ticket about spiking it

flows/classifier_specs/spec_interface.py
jesse-c: Did this get missed, or will come after this PR? 😅

olaughter: Oh shoot! I forgot to push the commit 🤦  new pr inbound

olaughter: https://github.com/climatepolicyradar/knowledge-graph/pull/535

PR #539, opened 2025-07-23
feat: Add progress reporting

jesse-c:
Use the Prefect [progress kind of artifact](https://docs.prefect.io/v3/how-to-guides/workflows/artifacts#create-progress-artifacts) for places where it either takes generally a while and/or there's a lot of things to enumerate through.

I've tried to make it as _for free_ as possible, so it's in the sub-flow mapping function, and I've created a gather equivalent.

More can be done as we go along in the repository.

**Demo**

Flow run: https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/06881402-67a0-797b-8000-7ccc08a829e7
1. 0%
CleanShot 2025-07-23 at 21 04 02

2. 50%

CleanShot 2025-07-23 at 21 06 16

3. 100%

CleanShot 2025-07-23 at 21 07 21

FIXES PLA-610

### DIFF
--- a/flows/utils.py
+++ b/flows/utils.py
@@ -10,16 +10,31 @@ from dataclasses import dataclass, field
 from functools import partial
 from io import BytesIO
 from pathlib import Path
-from typing import Any, Callable, NewType, TypeVar
+from typing import (
+    Annotated,
+    Any,
+    Callable,
+    Literal,
+    NewType,
+    ParamSpec,
+    TypeVar,
+    overload,
+)
+from uuid import UUID
 
 import boto3
 from botocore.exceptions import ClientError
+from prefect.artifacts import (
+    create_progress_artifact,
+    update_progress_artifact,
+)
 from prefect.client.schemas.objects import FlowRun, StateType
 from prefect.deployments import run_deployment
 from prefect.flows import Flow
 from prefect.settings import PREFECT_UI_URL
+from prefect.utilities.names import generate_slug
 from prefect_slack.credentials import SlackWebhook
-from pydantic import PositiveInt
+from pydantic import Field, PositiveInt, RootModel
 from typing_extensions import Self
 
 from scripts.cloud import (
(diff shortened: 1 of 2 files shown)


flows/utils.py
```diff
@@ -526,3 +568,61 @@ async def map_as_sub_flow(
...
+
+
+async def gather_and_report(
+    tasks: Sequence[Awaitable[T]],
+    return_exceptions: bool,
+    key: str,
+    desc_create: str,
+    desc_update_fn: Callable[[Sequence[Any], list[Any]], str] = default_desc,
+) -> Sequence[T] | Sequence[T | Exception]:
+    progress_artifact_id: UUID = await create_progress_artifact(
```
THOR300, line 604: I get some linting issues on this line and 622: 

> "UUID" is not awaitable
>   "UUID" is incompatible with protocol "Awaitable[_T_co@Awaitable]"
>     "__await__" is not presentbasedpyright[reportGeneralTypeIssues](https://docs.basedpyright.com/v1.29.4/configuration/config-files/#reportGeneralTypeIssues)

flows/utils.py
```diff
@@ -494,12 +530,18 @@ async def map_as_sub_flow(
...
-        *tasks,
+    def desc_update_fn(tasks, results) -> str:
+        return f"Finished sub-flow for {qualified_name}, progressing to {len(results)}/{len(tasks)} finished"
+
+    results: Sequence[FlowRun | BaseException] = await gather_and_report(
+        tasks=tasks,
         return_exceptions=True,
+        key=f"progress-sub-flows-{generate_slug(2)}",
+        desc_create=f"Starting sub-flows for {qualified_name} for {len(tasks)} tasks",
+        desc_update_fn=desc_update_fn,
```
THOR300, line 541: **thought**: Are we sure we need / want progress reporting for all concurrent sub deployment runs?

flows/utils.py
```diff
@@ -526,3 +568,61 @@ async def map_as_sub_flow(
...
+            else:
+                raise e
+        finally:
+            await update_progress_artifact(
+                artifact_id=progress_artifact_id,
+                progress=Percentage.from_lists(results, tasks),
+                description=desc_update_fn(tasks, results),
+            )
+
+    return results
```
THOR300, line 628: **suggestion**: It would be good to get test coverage for this functionality

THOR300 (approved):
Very nice as always! 🚀 

It’s a shame that these progress artefacts can’t support the following _“increment by 1”_ call. Then we could just update from the sub deployments.

flows/utils.py
jesse-c: It's the same as for the other `create_foo_artifact` calls, annoyingly.

flows/utils.py
jesse-c: I am sure, yup! It's a good first place to add progress reporting. Sub-flow runs are usually the biggest separation of sizes of work (hence why they're run on separate machines). Where else would you add progress reporting?

jesse-c: To focus on though it being default in the function, since that was your point, which I tangentially addressed: it's such an easy win to get it, that I couldn't think of any downsides to it being there. It helps with adoption too.

flows/utils.py
jesse-c: ✓

flows/utils.py
jesse-c: I've suggested, in Slack, walking through the code at some point.

flows/utils.py
jesse-c: If it does end up being distracting, or something else, we can make it optional later.

PR #544, opened 2025-07-24
Feature/pla 760 remove state return types in inference pipeline

THOR300:
This Pull Request: 
---
- Updates the inference flows to no longer return prefect `States` but now the types we want.

### DIFF
--- a/flows/inference.py
+++ b/flows/inference.py
@@ -1046,7 +1046,7 @@ async def inference(
     )
 
     if inference_result.failed:
-        return Fault(
+        raise Fault(
             msg="Some inference batches had failures!",
             metadata={},
             data=inference_result,
--- a/flows/inference.py
+++ b/flows/inference.py
@@ -21,7 +21,6 @@ from prefect.client.schemas.objects import FlowRun
 from prefect.concurrency.asyncio import concurrency
 from prefect.context import get_run_context
 from prefect.logging import get_run_logger
-from prefect.states import Completed, Failed, State
 from prefect.utilities.names import generate_slug
 from pydantic import BaseModel, ConfigDict, PositiveInt, SecretStr
 from types_aiobotocore_s3.type_defs import PutObjectOutputTypeDef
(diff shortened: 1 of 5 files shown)


THOR300:
**_Are there any limits on what can be returned from the flows as these are serialised and saved to s3 and then reloaded?_** 

Following the lengthy flow decorator code path down we can see what happens to the returned value should `result_storage` be enabled. The `ResultRecord` is instantiated . This is then passed into the `ResultStore` `apersist_result_record` method and finally serialised to bytes using as default the cloud pickle library. 

https://github.com/PrefectHQ/prefect/blob/526e51921eea72b44586cbfc570b2fb2ae0c4c66/src/prefect/states.py#L351
https://github.com/PrefectHQ/prefect/blob/526e51921eea72b44586cbfc570b2fb2ae0c4c66/src/prefect/results.py#L746
https://github.com/PrefectHQ/prefect/blob/526e51921eea72b44586cbfc570b2fb2ae0c4c66/src/prefect/serializers.py#L137
￼
From reading around this can handle serialising of simple pydantic objects but more complex ones and it may struggle. Our response types are quite simple right now but in future a custom serialiser with tests would most probably be beneficial. 

The documentation on this can be seen here:
    - https://docs.prefect.io/v3/advanced/results#result-serialization

Open to thoughts if people think this is necessary now.

flows/inference.py
```diff
@@ -1037,7 +1030,11 @@ def parameters(
             all_raw_successes.extend(raw_successes)
             all_raw_failures.extend(raw_failures)
 
-    all_successes = [BatchInferenceResult(**result) for result in all_raw_successes]
+    # The type of response when running as a sub deployment is:
```
michaelokarimia, line 1033: Is it worth highlighting as a comment in the code the point you've made about this currently working with simple types and that it could fail if we use more complex ones?

https://github.com/climatepolicyradar/knowledge-graph/pull/544#issuecomment-3113908102

michaelokarimia (approved):
Nice change, I'm liking the the more meaningful error response too. Left a comment asking if it's worth explictly adding a comment on if we should call out that this may not work with a more complex type https://github.com/climatepolicyradar/knowledge-graph/pull/544#issuecomment-3113908102

flows/inference.py
THOR300: Yeh good shout!

PR #546, opened 2025-07-28
Identifier consistency

harrisonpim:
a few small tweaks to the way that we generate classifiers identifiers, to make them a bit more consistent/coherent

### DIFF
--- a/src/classifier/classifier.py
+++ b/src/classifier/classifier.py
@@ -108,7 +108,7 @@ class Classifier(ABC):
     @property
     def id(self) -> Identifier:
         """Return a neat human-readable identifier for the classifier."""
-        return Identifier.generate(self.name, self.concept)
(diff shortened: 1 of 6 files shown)


src/classifier/classifier.py
```diff
@@ -107,11 +107,41 @@ def __eq__(self, other):
 
     @property
     def id(self) -> Identifier:
```
jesse-c, line 109: **thought:** Why not have this not provide a default implementation, so people always have to be ensure that they provide one that's sufficient? Or, like we were saying in the meeting that a `Concept` has its own ID, based on its parameters, then it'll always be sufficient?

harrisonpim: good point - not providing a default could force people to reimplement, which is probably best!

PR #548, opened 2025-07-29
decouple vibe-check config

harrisonpim:
Until now, we've used the `prod.yaml` from the `flows` dir to define which concepts should end up in the vibe checker (the same config which feeds inference and indexing etc). 

Now, the policy team want to see concepts in the vibe checking tool which aren't yet ready to be deployed further. Adding a totally separate, decoupled config file for the vibe checker should make this much easier 👌


jesse-c (approved):
LGTM. It's divorced from any version, so it may be hard to track what is seen in the vibe check tool vs is deployed, but I imagine that's okay for now—or at least until the canonical IDs are in.

PR #549, opened 2025-07-29
Feat: Simplify BatchInferenceResult 

THOR300:
This Pull Request: 
---
- Is the first in a series of stacked PR's aimed at implementing a bug fix for the full knowledge graph pipeline where we were considering documents as successful when they had failures across classifiers during inference. 
- This PR is updating the batch inference result to contain all the document stems and successful stems only. We can identify failed documents as those that explicitly don't have successful results as opposed to attempting to collect all failures. 

Follow up PR (staging tests were run from this PR): 
- https://github.com/climatepolicyradar/knowledge-graph/pull/550

### DIFF
--- a/flows/deploy_static_sites.py
+++ b/flows/deploy_static_sites.py
@@ -14,6 +14,26 @@ import subprocess
 
 from cpr_sdk.ssm import get_aws_ssm_param
 from prefect import flow, task
+from prefect.futures import wait
+
+
+def _wait_for_futures_and_check_failures(futures_dict: dict, task_type: str):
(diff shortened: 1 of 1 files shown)


tests/flows/test_inference.py
```diff
@@ -458,23 +458,26 @@ def test_remove_sabin_file_stems(
 
 
 def test_group_inference_results_into_states(snapshot):
```
michaelokarimia, line 460:
```suggestion
def test_get_inference_results_as_state_groups(snapshot):
```

I feel that the name of the method and subsequent test method could be made clearer

PR #550, opened 2025-07-29
Feature/simplify inference result

THOR300:
This Pull Request: 
---
- Is aimed at implementing a bug fix for the full knowledge graph pipeline where we were considering documents as successful when they had failures across classifiers during inference. 
- Simplifies the batch inference and inference result by removing references to unexpected failures and failed documents. We can identify failed documents as those that explicitly don't have successful results as opposed to attempting to collect all failures.

Staging Test: 
- [No aggregation with failed classifiers](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/06889e59-20f7-7d8c-8000-57156b16b20b?state=pending&entity_id=06889e71-c475-7824-8000-23087e7aead6)
- [Aggregation and Indexing Run due to successful docs](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/06889e76-7e78-7860-8000-2e82f99aa00d?entity_id=06889e94-f200-7856-8000-67e7f1a73651)


flows/inference.py
```diff
@@ -127,66 +128,71 @@ class BatchInferenceResult(BaseModel):
...
+    successful_document_stems: list[DocumentStem]
     classifier_name: str
     classifier_alias: str
 
     @property
     def failed(self) -> bool:
         """Whether the batch failed, True if failed."""
 
-        return self.failed_document_stems != [] or self.unknown_failures != []
+        return len(self.batch_document_stems) != len(self.successful_document_stems)
```
olaughter, line 140: question: What is the difference here following this change? Could you give an example of edge cases this means we now cover?

flows/inference.py
```diff
@@ -127,66 +128,71 @@ class BatchInferenceResult(BaseModel):
...
 
 class InferenceResult(BaseModel):
     """Result from running inference on all batches of documents."""
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
+    document_stems: list[DocumentStem]
+    classifier_specs: list[ClassifierSpec]
     batch_inference_results: list[BatchInferenceResult] = []
-    unexpected_failures: list[BaseException | FlowRun] = []
```
olaughter, line 149: question: are we not losing some context here by removing this? The advantage of being explicit is that we can know what failures we have managed to catch and which might have fallen through the cracks somewhere. It also means in the future we can support more specific behaviour for different kinds of failure

olaughter (commented):
Thanks for consolidating the two PR's for me Mark! I have some uncertainty about wether this is a helpful direction to go with our failure behaviour. And have added some questions below to talk it through

flows/inference.py
THOR300: Yeah there's no harm in the field being there, it's just as of this PR we're not actually using it to ascertain non/succeeded documents!

flows/inference.py
THOR300: Current Behavior on main:
- Success is defined implicitly: A document stem is marked as succeeded if it’s not in the set of known failures.
- Issue: When an unexpected error occurs during inference_batch_of_documents, we don’t capture the corresponding document stems, so they aren't marked as failed — and by the current logic, are incorrectly assumed to be succeeded.

New Behavior in the PR:
- Success is now explicit: Only document stems that are explicitly successful across all classifiers are counted as succeeded.
- This avoids false positives, i.e. cases where a document was neither marked as failed nor as succeeded but was incorrectly counted as succeeded.

Why this is better:
- It’s more robust and conservative — success is only counted when you’re absolutely sure.
- It closes a logical hole where documents with silent or unexpected failures could slip through.
- It likely fixes a bug, as evidenced by updated behaviour in the test case (test_inference.py:L801).

michaelokarimia (approved):
Would prefer the [function rename](https://github.com/climatepolicyradar/knowledge-graph/pull/550/files#r2243155557) but happy for this to be merged to main, in order to unblock the work

flows/inference.py
olaughter: Just to clarify, I don't think adding an artefact showing these is in scope, but we are losing some context through this change in that we are no longer passing up the failure details. And wonder if that is the direction we want to go in

THOR300: To clarify this field is only currently used on main for identifying whether inference was successful or not.

Regarding artifacts we're still reporting / creating output artefacts from this data we're just not passing it back up the chain to the parent `full-pipeline` to decide which documents to keep running on. 
- https://github.com/climatepolicyradar/knowledge-graph/blob/feature/simplify-inference-result/flows/inference.py#L904
- https://github.com/climatepolicyradar/knowledge-graph/blob/feature/simplify-inference-result/flows/inference.py#L1067

We can see this in this example prefect run where we have the successful classifiers at the top level and the document level results for the batches: 
- https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/06889e76-7e78-7860-8000-2e82f99aa00d?entity_id=06889e7d-2b40-7617-8000-0e0be20b8886&state=pending
- https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/06889e7d-2b40-7617-8000-0e0be20b8886?state=pending&entity_id=06889e7d-b1da-764b-8000-951af72902d1

image

flows/inference.py
olaughter: Having run through this on a call I'm happy with the behaviour, but have asked for:
- More documentation about the full_pipeline
- clearer test case for the described issue

flows/inference.py
```diff
@@ -127,66 +128,71 @@ class BatchInferenceResult(BaseModel):
...
-        return (
-            any([result.failed for result in self.batch_inference_results])
-            or self.unexpected_failures != []
-        )
+        return any([result.failed for result in self.batch_inference_results]) or len(
+            self.document_stems
+        ) != len(self.successful_document_stems)
 
     @cached_property
     def successful_document_stems(self) -> set[DocumentStem]:
```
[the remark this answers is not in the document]
THOR300: Yes prefer this, it distinguishes between the attribute on the `BatchInferenceResult` 🙌

PR #551, opened 2025-07-30
Feat: Dynamically Load Prefect Result Persistence S3 Block 

THOR300:
This Pull Request: 
---
- Looks at dynamically loading the name of the s3 block for persisting results and then setting result persistence using the `with_options` method on the `Flow`. 
- Quite nice that we only persist results optionally also.


tests/flows/test_full_pipeline.py
```diff
@@ -190,13 +199,13 @@ async def test_full_pipeline_no_config_provided(
...
-        assert call_args.kwargs["config"] == test_config
-        assert call_args.kwargs["classifier_specs"] is None
-        assert call_args.kwargs["document_ids"] is None
-        assert call_args.kwargs["use_new_and_updated"] is False
-        assert call_args.kwargs["batch_size"] == INFERENCE_BATCH_SIZE_DEFAULT
+        assert len(mock_inference.mock_calls) == 1
+        assert mock_inference.mock_calls == [
+            call.with_options(
+                # TODO: Make dynamic
+                result_storage="s3-bucket/s3-bucket/cpr-sandbox-prefect-results-cache"
```
THOR300, line 206: Prefix is wrong.

tests/flows/test_full_pipeline.py
```diff
@@ -122,20 +121,33 @@ def test_validate_aggregation_inference_configs_raises_value_error(
...
+class MockFlowWithOptions:
+    """Mock flow with options that returns a given value."""
+
+    def __init__(self, return_value):
+        """Mock __init__ method that sets the return value."""
+        self.return_value = return_value
+
+    async def __call__(self, **kwargs):
+        """Mock __call__ method that returns the return value."""
+        return self.return_value
```
THOR300, line 133: Could add functionality to store call args but this seems to be re-inventing the wheel.

THOR300:
Deployed to staging and tested: 
- https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/0689cb1e-6e65-75ea-8000-e261d367e115

flows/full_pipeline.py
```diff
@@ -137,14 +170,13 @@ async def full_pipeline(
...
-    inference_run: State = await inference(
+    inference_run: State = await inference_with_result_cache(
         classifier_specs=classifier_specs,
         document_ids=document_ids,
         use_new_and_updated=inference_use_new_and_updated,
         config=inference_config,
         batch_size=inference_batch_size,
         classifier_concurrency_limit=inference_classifier_concurrency_limit,
-        return_state=True,
     )
```
jesse-c, line 180: **quibble:** I'd just do it inline, like is done elsewhere[^1]. I find this to be better as it's right there where it's happening, much more succinct, prevents adding another tiny in-between function, and is ~2 more lines of code, versus ~40 lines of code.
```suggestion
    inference_run: State = await inference.with_options(result_storage=result_cache_s3_block_uri)(
            classifier_specs=classifier_specs,
        document_ids=document_ids,
        use_new_and_updated=inference_use_new_and_updated,
        config=inference_config,
        batch_size=inference_batch_size,
        classifier_concurrency_limit=inference_classifier_concurrency_limit,
        return_state=True,
        )
```

[^1]: https://github.com/climatepolicyradar/knowledge-graph/blob/main/flows/inference.py#L873-L880

jesse-c (approved):
LGTM

flows/full_pipeline.py
THOR300: Ahh yeah the reason why I did it this way was for testing purposes! When I was looking at mocking the job variable load and inference call etc. it was causing the PR to contain a lot of test changes which I was keen to avoid.

PR #552, opened 2025-07-30
test: always skip the tests that contact argilla

olaughter:
These tests were actually creating a dataset in argilla. Judging by the docstring it is not intended to be run on a normally in tests and that shows in ci where we ignore it. However when running locally its easy not to ignore this module, I've therefore set it to always skip for now. I've also ensured local env variables for argilla are not picked up on by defining them for pytest with test values.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,7 +79,7 @@ jobs:
         run: just vespa_dev_setup
 
       - name: Run tests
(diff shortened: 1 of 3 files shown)


olaughter:
@harrisonpim I'm conscious you are away at the moment so lets chat when you're back about what would be good for these tests longer term. My guess is that logner term we won't want them to be directly contacting with argilla or to be skiped entirely 😅

PR #555, opened 2025-07-30
PLA-769 Log count of excluded Sabin files

michaelokarimia:
We currently exclude Sabin documents from being processed in the pipeline, but this is done silently. This commit counts the number of filtered Sabin documents and then print it to the console so it is visible when examining logs. Discovered as part of PLA-761


olaughter (approved):
Nice!

PR #557, opened 2025-07-30
Feat: Improve Inference Concurrency 

THOR300:
This Pull Request: 
---
- Improves inference concurrency by collecting all batches of inference jobs for all classifiers and then passing this into the `map_as_sub_flow`.
- This is in response to a recent run where running one document through inference took multiple hours as the for loop approach meant we waited for all inference jobs to finish for each classifier before moving to the next. We have 70ish classifiers so 2mins*70 -> 2 hrs. 
- Unit tests: The inference test is being fixed currently by [this](https://linear.app/climate-policy-radar/issue/PLA-771/testing-fix-testsflowstest-inferencepytest-inference-test) ticket and `map_as_sub_flow` has no tests, new tests implemented here.
- Staging run: Done.


flows/utils.py
```diff
@@ -535,12 +535,25 @@ async def map_as_sub_flow(
...
+    unwrap_result: Literal[True],
+) -> tuple[Sequence[R], Sequence[BaseException | FlowRun]]: ...
+
+
+async def map_as_sub_flow(
+    fn: Flow[P, R],
+    aws_env: AwsEnv,
+    counter: PositiveInt,
+    batches: Generator[Any, None, None],
+    parameters: Callable[[Any], dict[str, Any]],
```
THOR300, line 556: Did we fix this previously before the move to map as sub flow>

THOR300:
### Proving what the problem is on main:
[Here]([https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/06887738-6533-72f6-8000-1326782deba8?g_range={%22type%22:%22span%22,%22seconds%22:-2592000}&state=pending&entity_id=0688776d-4c2a-71e8-8000-ee416e171679](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/06887738-6533-72f6-8000-1326782deba8?g_range=%7B%2522type%2522:%2522span%2522,%2522seconds%2522:-2592000%7D&state=pending&entity_id=0688776d-4c2a-71e8-8000-ee416e171679)) 
we run two documents through every classifier in production. 
What we can see is that we’re running one batch level inference job at once, waiting for this to complete and then scheduling the next. This can be shown by selecting the batch level inference flows and looking at the `Scheduled start`. 
This inference run then ultimately took over 2 hours to run inference for one document. 
So it's clear to me that on main we are not optimally running inference. 
This is due to the `for loop` approach over classifiers 
[here.](https://github.com/climatepolicyradar/knowledge-graph/blob/main/flows/inference.py#L1011)

THOR300:
# Proving that this change solves the problem: 

Staging Inference Run | 2 docs | All Classifiers 
- [Here](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/0688b6b4-7c5d-76ea-8000-4d71a54714f3?state=pending&entity_id=0688b6b4-7c5d-76ea-8000-4d71a54714f3) we see 20 inference jobs spawning (default indexer concurrency limit that's managed by semaphore) initially. 
- We can see that different classifiers are being run at the same time as opposed to the previous example. 
- Looking at the `Scheduled start` time of the batch jobs in the second/third waves we can that they align with the `End time` of the jobs within the prior wave. 
- Inference for the two documents across all classifiers took ___ mins. 


INSERT IMAGE

THOR300:
### Did we fix this before? 

- [This](https://github.com/climatepolicyradar/knowledge-graph/commit/885058772ab7220d41937d923317fde0c8506bea) previous inference run supports that we were previously running different inference classifiers at once. 
- This can be seen by the scanning through the jobs in the second "wave" for example.
- The `Scheduled start` time of some jobs in the second wave also is before the end time of the jobs in the first wave. E.g. we were not waiting for all of the jobs for one classifier to finish before starting the next. 

image

- Looking back over the history of the inference job I believe this is the commit where the functionality changed. Previously before `map_as_sub_flow` we had a list comprehension creating all tasks for an inference run at once. [Commit](https://github.com/climatepolicyradar/knowledge-graph/commit/885058772ab7220d41937d923317fde0c8506bea)
- `flows/inference.py:L787`

So should we revert back to this? 
- Maybe I'm biased because I've started this PR now haha but `map_as_sub_flow` is used across the repo and is working we just have to pass in the correct batches so I'd vote to keep what we have and implement this change.

flows/utils.py
```diff
@@ -535,12 +535,25 @@ async def map_as_sub_flow(
...
 
 
+@overload
 async def map_as_sub_flow(
     fn: Flow[P, R],
     aws_env: AwsEnv,
     counter: PositiveInt,
-    batches: Generator[Sequence[T], None, None],
-    parameters: Callable[[Sequence[T]], dict[str, Any]],
+    batches: Generator[tuple[ClassifierSpec, Sequence[DocumentStem]], None, None],
```
jesse-c, line 543: **fix:** This shouldn't be changed to be specific like this. At most, it should be:
```suggestion
    batches: Generator[T, None, None],
```
Not every where that calls this function should/could pass `tuple[ClassifierSpec, Sequence[DocumentStem]]`.

`T` is a generic, so you can past *any*thing you want.

flows/utils.py
jesse-c, line 556: **fix:** `Any` is frowned upon. This is similar to https://github.com/climatepolicyradar/knowledge-graph/pull/557/files#r2245738951.

jesse-c (changes requested):

flows/utils.py
jesse-c: The actual implementation shouldn't have an `@overload` that's introduced as well, AFAIK.

flows/utils.py
```diff
@@ -535,12 +535,25 @@ async def map_as_sub_flow(
 ) -> tuple[Sequence[FlowRun], Sequence[BaseException | FlowRun]]: ...
 
 
+@overload
```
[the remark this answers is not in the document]
jesse-c: The `@overload` is to have typing checking for the possible values of `unwrap_result` and then how that influences the return type. I'd say there is a common interface, but I'm not 100% sure of the usage of _interface_ here, as there's just the parameters.

FWIW there's no difference in which parameters are accepted.

flows/utils.py
THOR300: Thanks for the comments I'll try and outline justification for my approach as well as consolidate chats from this [comment](https://github.com/climatepolicyradar/knowledge-graph/pull/557/files#r2245826976) as well. 

So I've updated the parameters callable to now accept a `tuple[ClassifierSpec, Sequence[DocumentStem]]` to link a classifier to a batch, allowing us to call the `map_as_sub_flow` function on all the units of work we want processed during the inference run. 

This now means that the linter is not happy as we are no longer passing in a `Sequence[T]` for the batches etc. Therefore, I've added a new over load so as to satisfy this. 

I agree we don't want an overload explosion here but this is what I'm understanding as the recommended solution from the [documentation](https://typing.python.org/en/latest/spec/overload.html#argument-type-expansion). There's also an example [here](https://typing.python.org/en/latest/spec/overload.html#overload-definitions) listing callables with different params that looks similar to our use case. 

E.g. we shouldn't try and update the `batches` and `parameters` params to be unions of the possible types as the documentation doesn't recommend this and they are linked (can't pass a tuple into the aggregation definition of the parameters callable). 

---

Open to what you guys are suggesting instead of this approach! One thing I've seen is the constrained type var. Wondering whether we could d osomething as follows: 

```python 
AnyParam = TypeVar("AnyParam", Iterable[DocumentStem], Iterable[tuple[ClassifierSpec, Sequence[DocumentStem]]
```

flows/inference.py
```diff
@@ -1018,35 +1018,48 @@ async def inference(
...
+    ) -> Generator[tuple[ClassifierSpec, Sequence[DocumentStem]], None, None]:
+        """Create batches that pair each classifier with document batches."""
+        for document_batch in document_batches:
+            for classifier in classifiers:
+                yield (classifier, document_batch)
+
+    document_batches = iterate_batch(filtered_file_stems, batch_size)
+    classifier_document_batches = create_classifier_batch_generator(
+        document_batches, classifier_specs
+    )
```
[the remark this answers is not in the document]
THOR300: Yeah I wanted to distinguish between a batch of documents and a batch of classifier+documents! 

Regarding the classifier spec naming I've moved the function and updated the param name. My intention with `classifiers` was to ensure we weren't using the same variable name as is used in the outer scope of the function in the inference flow.

jesse-c (commented):
See: Newly submitted comments.

flows/utils.py
THOR300: Thanks for clarifying that all @jesse-c!

jesse-c: No worries! 🙇🏻

olaughter (approved):
This is looking really clean, nice work!

flows/utils.py
```diff
@@ -539,8 +537,7 @@ async def map_as_sub_flow(
     fn: Flow[P, R],
     aws_env: AwsEnv,
     counter: PositiveInt,
-    batches: Generator[Sequence[T], None, None],
-    parameters: Callable[[Sequence[T]], dict[str, Any]],
+    parameterised_batches: Generator[dict[str, Any], None, None],
```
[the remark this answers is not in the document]
jesse-c: Yup, this is unavoidable, with the other changes having been made.

PR #562, opened 2025-08-04
PLA-771 Fix "flaky_on_ci" tests

michaelokarimia:
Fixes inference test which was marked as flaky on CI

Rewrote the test so it checks for BatchInferenceResults, as the behaviour of the function has changed after the test was disabled.

I've removed some assertions fom this test that are tested by the [test_inference_on_batch_of_documents](https://github.com/climatepolicyradar/knowledge-graph/blob/PLA-771-fix-flaky-ci-tests/tests/flows/test_inference.py#L540)

image


olaughter:
In case you find it useful for next time, you can also [manually trigger these tests](https://github.com/climatepolicyradar/knowledge-graph/blob/main/.github/workflows/ci.yml#L9) to run on a chosen branch saving you from having to open up a PR to iterate with. The option can be found [on the action](https://github.com/climatepolicyradar/knowledge-graph/actions/workflows/ci.yml).

tests/flows/conftest.py
```diff
@@ -325,8 +325,12 @@ def mock_bucket_documents(mock_s3_client, mock_bucket):
 
 
 @pytest.fixture
-def mock_bucket_documents_b(mock_s3_client, mock_bucket_b):
-    fixture_files = ["PDF.document.0.1.json", "HTML.document.0.1.json"]
+def mock_bucket_sabin_documents(mock_s3_client, mock_bucket):
```
THOR300, line 328: **suggestion**: Seeing as this fixture contains documents that aren't all of the Sabin corpus I'd change this naming.

tests/flows/fixtures/Sabin.document.16944.17490.json
```diff
@@ -0,0 +1,149 @@
+[
```
THOR300, line 1: **suggestion**: This data looks like it's a `LabelledPassage` whereas the other documents in the fixture are `ParserOutput` for example `tests/flows/fixtures/HTML.document.0.1.json`. We pass documents of the `ParserOutput` model into inference.

tests/flows/test_inference.py
```diff
@@ -248,44 +249,80 @@ async def test_text_block_inference_without_results(
...
+        try:
+            inference_result = await inference(
+                classifier_specs=[expected_classifier_spec],
+                document_ids=input_doc_ids,
+                config=test_config,
+            )
+        except ValueError:
+            # Expected to fail due to test environment mock limitations
+            # See below for assertations against the results
+            pass
```
THOR300, line 311: **suggestion**: Here rather than catching the exception we could use `return_state=True`

THOR300 (approved):
Looking great, huge to get this in!

tests/flows/test_inference.py
michaelokarimia: Added a type check assertion

.github/workflows/ci.yml
```diff
@@ -79,7 +79,7 @@ jobs:
         run: just vespa_dev_setup
 
       - name: Run tests
-        run: just test -v -m "'not flaky_on_ci and not transformers'"
+        run: just test -v -m "'not transformers'"
```
olaughter, line 83: So good to see!

tests/flows/test_index.py
```diff
@@ -337,7 +337,7 @@ async def test_run_indexing_from_aggregate_results__invokes_subdeployments_corre
     mock_run_output_identifier_str: str,
     test_aggregate_config,
 ) -> None:
-    """Test that run passage level indexing correctly from aggregated restuls."""
+    """Test that run passage level indexing correctly from aggregated results."""
```
olaughter, line 340: This is the second time I've seen this typo this week! 😆

olaughter: Time to get our keyboards cleaned, maybe! haha

tests/flows/test_inference.py
```diff
@@ -248,44 +249,77 @@ async def test_text_block_inference_without_results(
...
+    ]
+    expected_classifier_spec = ClassifierSpec(name="Q788", alias="v13")
+
+    with patch("flows.utils.run_deployment") as mock_inference_run_deployment:
+        flow_run_counter = 0
+
+        async def mock_awaitable(*args, **kwargs):
+            nonlocal flow_run_counter
+            flow_run_counter += 1
+            from prefect.results import ResultRecord
```
olaughter, line 278: question: any reason against having this at the top of the module?

tests/flows/test_inference.py
```diff
@@ -248,44 +249,77 @@ async def test_text_block_inference_without_results(
...
+                                expected_doc_stems
+                            ),  # all documents were classified successfully
+                            classifier_name=expected_classifier_spec.name,
+                            classifier_alias=expected_classifier_spec.alias,
+                        )
+                    )
+                ),
+            )
+
+        mock_inference_run_deployment.side_effect = mock_awaitable
```
olaughter, line 298: Nice! Good to know it was the deployment that was the core issue in the end

tests/flows/test_inference.py
```diff
@@ -248,44 +249,77 @@ async def test_text_block_inference_without_results(
...
+    expected_doc_stems = [
+        DocumentStem(Path(doc_file).stem) for doc_file in mock_bucket_documents
+    ]
+    expected_classifier_spec = ClassifierSpec(name="Q788", alias="v13")
+
+    with patch("flows.utils.run_deployment") as mock_inference_run_deployment:
+        flow_run_counter = 0
+
+        async def mock_awaitable(*args, **kwargs):
+            nonlocal flow_run_counter
```
olaughter, line 276: question: what are we using the flow_run_counter for, if its just to provide a unique name for the flow run I wonder if  the name I wonder if just using a hash would be sufficient and avoid the need for a nonlocal?

tests/flows/test_inference.py
olaughter: suggestion: we could also ensure this mock is being called, and with what we expect (to avoid regressions in the future where we break the mock and start calling actual infrastructure for example)
```
       assert mock_inference_run_deployment..assert_called_once_with(...)
```

olaughter (approved):
Really nice work untangling this! Hope its been a useful intro to the codebase

PR #563, opened 2025-08-04
ci: add stale pr detection

olaughter:
We have a number of stale PRs across our codebase, I'd like to find a way of keeping these cleaned up and am interested in trying this out here in the knowledge graph repo first. I've taken 40 days as the amount of time before a PR should actually be closed for now. My view is that after this time the PR is probably never going to be merged because the rest of the repo has changed, so it either should have gotten merged earlier or needs to be reopened. The specific timing could be adjusted though.

Note that for exceptions the label `stale-exempt` can be added to a PR.

See for reference: https://github.com/actions/stale

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,26 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: '12 9 * * 1' # every monday at 9:12 AM
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    env:
+      DAYS_BEFORE_STALE: 30
+      DAYS_BEFORE_CLOSE: 10
+    steps:
+      - uses: actions/stale@v9
+        with:
+          stale-issue-message: 'This issue is stale because it has been open ${{ env.DAYS_BEFORE_STALE }} days with no activity. Remove stale label or comment or this will be closed in ${{ env.DAYS_BEFORE_ISSUE_CLOSE }} days.'
+          stale-pr-message: 'This PR is stale because it has been open ${{ env.DAYS_BEFORE_STALE }} days with no activity. Remove stale label or comment or this will be closed in ${{ env.DAYS_BEFORE_PR_CLOSE }} days.'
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,26 @@
...
+      DAYS_BEFORE_STALE: 30
+      DAYS_BEFORE_CLOSE: 10
+    steps:
+      - uses: actions/stale@v9
+        with:
+          stale-issue-message: 'This issue is stale because it has been open ${{ env.DAYS_BEFORE_STALE }} days with no activity. Remove stale label or comment or this will be closed in ${{ env.DAYS_BEFORE_ISSUE_CLOSE }} days.'
+          stale-pr-message: 'This PR is stale because it has been open ${{ env.DAYS_BEFORE_STALE }} days with no activity. Remove stale label or comment or this will be closed in ${{ env.DAYS_BEFORE_PR_CLOSE }} days.'
+          close-issue-message: 'This issue was closed because it has been stalled for ${{ env.DAYS_BEFORE_CLOSE }} days with no activity.'
+          close-pr-message: 'This PR was closed because it has been stalled for ${{ env.DAYS_BEFORE_CLOSE }} days with no activity.'
+          exempt-pr-labels: 'stale-exempt'
```
THOR300, line 19: praise: Ahh nice this is good that we can retain the ones we want!

olaughter:
> Too confirm with this would the alerting of a closed PR be a github notification or silent?

It'll be a comment on the PR. So I guess that means that the author and reviewers will get a notification

THOR300 (approved):
Nice! Good to get this in. 

You're most probably already on top of this but it would be good to go over the current stale PR's before they get auto closed next Monday post merge of this. For example at a quick scan I see an old [one](https://github.com/climatepolicyradar/knowledge-graph/pull/369) from matyas.

olaughter:
> If we're to do this, I think we need to better at a. paying attention to our daily summary of open PRs and b. don't leave DSci hanging, as we hitherto have.

Yeah lets make a point to do this, we can also extend the times before prs are considered stale if we end up finding the timing it is too tight from that perspective

PR #564, opened 2025-08-04
feat: include env in the wandb `artifact` alias

olaughter:
The model connection between artifacts and s3 stores is a one to one, which means unless we revisit this our models will all have a coupling with a specific environment. This is now the only tangible difference between versions of a model. This adds the env as an alias to each model. We already do this in the registry at promotion, but as models stay in their env it makes sense to do it in training too

The benefits to this is are:

* Clarity for humans when reviewing wandb which model is for which environment
* Easily identify the environment for a model programmatically in a more dynamic way.

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -109,7 +109,7 @@ def create_and_link_model_artifact(
     artifact.add_reference(uri=uri, checksum=True)
 
     # Log the artifact to W&B, creating it within a wandb project
-    artifact = run.log_artifact(artifact)
+    artifact = run.log_artifact(artifact, aliases=[storage_link.aws_env.value])
     artifact = artifact.wait()
 
     return artifact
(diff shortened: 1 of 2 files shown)


michaelokarimia (approved):

jesse-c:
This information is already in the metadata[^1][^2], enabling the 2 benefits listed here? Perhaps that should be removed now, so the data isn't in 2 places?

[^2]: https://github.com/climatepolicyradar/knowledge-graph/blob/main/scripts/train.py#L91-L94
[^1]:
CleanShot 2025-08-04 at 14 53 20

olaughter:
> Actually, you wouldn't want to remove it from the metadata, as it's used[1](#user-content-fn-1-1f42b1896db532d3fe13a258f8647efd):
> 
> ## Footnotes
> 1. https://github.com/climatepolicyradar/knowledge-graph/blob/main/scripts/update_classifier_spec.py#L83-L85 [↩](#user-content-fnref-1-1f42b1896db532d3fe13a258f8647efd)

Thanks. When I get to the `update_classifier_spec` I'll aim to switch it over to use the alias and then remove it from the metadata 👍

PR #567, opened 2025-08-05
refactor: remove Across code

olaughter:
We ended up leaving across account promotion unimplemented. Now that we are changing promote it makes sense to remove it so that we don't have to maintain something we are not using. We probably will want better handling of accounts and train/promoting one day. At which point if we decide we still want to continue with the approach we had started here we can still reach into the commit history.

### DIFF
--- a/tests/test_promote.py
+++ b/tests/test_promote.py
@@ -1,4 +1,5 @@
 import os
+from contextlib import nullcontext
 from unittest.mock import Mock, patch
 
 import pytest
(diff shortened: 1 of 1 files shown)


docs/docs/developers/justfile.md
```diff
@@ -42,16 +42,10 @@ Or to also track in W&B:
 just evaluate Q992 --track
 ```
 
-You can promote a model version from one AWS account/environment, to another. You can optionally promote that model to be the primary version that's used in that account.
+You can promote a model version to be a primary version that's used in that account.
```
jesse-c, line 45: **fix:** There's only every 1 primary model at a time, no?
```suggestion
You can promote a model version to be the primary version that's used in that account.
```

tests/test_promote.py
```diff
@@ -63,109 +55,75 @@ def test_version_sorting_with_larger_numbers():
...
+    init_mock.return_value.__enter__ = Mock()
+    init_mock.return_value.__exit__ = Mock()
 
-    monkeypatch.setattr(
-        "scripts.promote.copy_across_aws_envs",
-        lambda *args: ("mock-bucket", "mock-key"),
-    )
-
-    monkeypatch.setattr("wandb.init", lambda **kwargs: Mock())
+    monkeypatch.setattr("wandb.init", init_mock)
```
jesse-c, line 117: **polish:** I wonder if it could be overridden/mocked to use a `null context`?

jesse-c (approved):
LGTM

tests/test_promote.py
olaughter: Great idea

olaughter: Found I can cover the mocks use nicely with: `init_mock = Mock(return_value=nullcontext(Mock()))`

PR #570, opened 2025-08-05
refac: move Version tests to own file

olaughter:
Version was previously refactored out of `promote.py`, this brings the corresponding tests out to their own file too


jesse-c (approved):
Nice! 🧼

PR #571, opened 2025-08-05
Represent model path with a pydantic model

olaughter:
Following on from [discussion on a previous PR](https://github.com/climatepolicyradar/knowledge-graph/pull/556#discussion_r2244776090), where I agreed to follow up on this separately


THOR300 (approved):
LGTM 

Wondering whether a `TypeAlias` would have sufficed but this is all good.

PR #574, opened 2025-08-05
Update promote script to use classifier ID

olaughter:
Switches promote to use classifier ID. For this

* The project model now gets looked up using the `classifier_id` argument (which has replaced `classifier`).
* The version is no longer a necessary parameter so it is removed and the id:env is used to look up the classifier instead.
* The lookup of existing collection artifacts is revised to use the artifact id instead of just the version

Interestingly, it looks like we had a bug in the way we looked up existing registry items, as the code had some confusion about which version was being used (project artifact version versus collection artifact version, which are not in sync). This has been resolved by using the `classifier_id` to make sure we are actually looking up the right classifier.

### DIFF
--- a/scripts/promote.py
+++ b/scripts/promote.py
@@ -13,7 +13,8 @@ from scripts.cloud import (
     is_logged_in,
     parse_aws_env,
 )
-from src.identifiers import WikibaseID
+from scripts.utils import ModelPath
+from src.identifiers import Identifier, WikibaseID
 from src.version import Version
 
 log = logging.getLogger(__name__)
@@ -95,10 +96,11 @@ def main(
             parser=WikibaseID,
         ),
     ],
-    classifier: Annotated[
-        str,
+    classifier_id: Annotated[
+        Identifier,
         typer.Option(
-            help="Classifier name that aligns with the Python class name",
+            help="Classifier ID that aligns with the Python class name",
+            parser=Identifier,
         ),
     ],
     version: Annotated[
@@ -165,7 +167,8 @@ def main(
         # This also validates that the classifier exists. It relies on an
         # artifiact not existing. That is, when trying to `use_artifact`
         # below, it'll throw an exception.
-        artifact_id = f"{wikibase_id}/{classifier}:{version}"
+        model_path = ModelPath(wikibase_id=wikibase_id, classifier_id=classifier_id)
+        artifact_id = f"{model_path}:{version}"
         log.info(f"Using model artifact: {artifact_id}...")
         artifact: wandb.Artifact = run.use_artifact(artifact_id)
 
--- a/scripts/promote.py
+++ b/scripts/promote.py
@@ -103,13 +103,6 @@ def main(
             parser=Identifier,
         ),
     ],
-    version: Annotated[
-        Version,
-        typer.Option(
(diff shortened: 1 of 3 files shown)


THOR300 (approved):
Looking great! 

It could be good to get some documentation in here or near the code that details the differences / definitions for aliases, versions and ids. 🚀

scripts/promote.py
```diff
@@ -35,23 +36,33 @@ def throw_not_logged_in(aws_env: AwsEnv):
...
+def find_artifact_in_registry(
+    model_collection, expected_id: Identifier, expected_version: Version
 ) -> Optional[wandb.Artifact]:
     """Find an artifact with the specified version in the model collection."""
-    return next(
-        (art for art in model_collection.artifacts() if art.version == str(version)),
-        None,
-    )
+    for art in model_collection.artifacts():
+        classifier_id, version = art.source_name.split(":")
```
jesse-c, line 44: **question:** Is this because the `source_name` includes the canonical ID and the version, as seen here?

Image

scripts/promote.py
```diff
@@ -65,9 +76,11 @@ def check_existing_artifact_aliases(
         name=target_path,
     )
 
-    target_artifact = find_artifact_by_version(model_collection, version)
+    target_artifact = find_artifact_in_registry(
+        model_collection, classifier_id, version
+    )
 
-    # It's okay if there isn't yet an artifact for this version, and
+    # It's okay if there isn't yet an registry artifact for this version, and
```
jesse-c, line 83: **nitpick:** Tiny grammar correction:
```suggestion
    # It's okay if there isn't yet a registry artifact for this version, and
```

olaughter:
> FWIW I'm still unsure about this. 😆 I did a quick spot check the other week, and found them in sync. That may have been luck though.

They mostly are in sync, which I think is why it hasnt caused much of an issue, there are a few [examples on Q123](https://wandb.ai/climatepolicyradar/Q123/artifacts/model/RulesBasedClassifier/v1) which I saw as I was using it for testing

scripts/promote.py
jesse-c: I was wondering why the classifier ID alone isn't sufficient, since the solution we were aiming for was for post-train, to not need the version at all?

jesse-c (commented):
Initial pass.

scripts/promote.py
```diff
@@ -95,17 +108,11 @@ def main(
...
-    ],
-    version: Annotated[
-        Version,
+    classifier_id: Annotated[
+        Identifier,
         typer.Option(
-            help="Version of the model (e.g., v3)",
-            parser=Version,
+            help="Classifier ID that aligns with the Python class name",
+            parser=Identifier,
```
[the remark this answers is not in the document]
olaughter: That sounds sensible to me. I'll take that on and do it separately to this PR

scripts/promote.py
jesse-c: Nice! This may not be relevant, but, you can look at the `aws_env` field in the metadata, so you could just get the artifact by alias, and then look at the metadata, avoiding the version?

jesse-c: Excellent! I think this puts it in a very good spot.

PR #575, opened 2025-08-05
even more ids to the vibe check config

harrisonpim:
[as requested by the policy team](https://climate-policy-radar.slack.com/archives/C096B8HD856/p1754408849304529?thread_ts=1753981327.103759&cid=C096B8HD856)


THOR300 (approved):
LGTM

PR #579, opened 2025-08-06
docs: Update documentation following id/version changes

olaughter:
There are more pieces of the model lifecycle that need updating but this follows a skim over the docs for what has changed so far and updates them.

Following a request for more documentation [here](https://github.com/climatepolicyradar/knowledge-graph/pull/574#pullrequestreview-3091805133), more should come as I go though

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -7,7 +7,7 @@ Individual scripts' docstrings contain more information about their purpose and 
 
 ## Updating a Classifier
 
-First we run the training scripts. This will increment the version of the classifier in the projects registry in weights and biases. You can then run the promote script which is used for promoting a model to primary within or cross aws environment. Promotion will not affect the version / alias.
+First we run the training scripts. This will upload the classifier to s3 and link to it from its weights and biases project. You can then run the promote script which is used for promoting a model to primary within an aws environment. Promotion as adds the model to the weights and bias registry and setting it as primary gives it the environment alias.
 
 _Note: You will need a profile in your `.aws/config` file with an active terminal session to use the following command as the upload command requires s3 access._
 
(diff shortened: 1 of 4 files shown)


scripts/promote.py
```diff
@@ -123,7 +123,7 @@ def main(
     primary: Annotated[
         bool,
         typer.Option(
-            help="Whether this will be the primary version for this AWS environment",
+            help="Whether this will be primary for this AWS environment",
```
THOR300, line 126:
```suggestion
            help="Whether this will be primary artifact for this AWS environment, marked by alias.",
```

THOR300: **thought**: Thinking whether this is the entity that we're referencing here?

scripts/README.md
```diff
@@ -7,7 +7,7 @@ Individual scripts' docstrings contain more information about their purpose and
 
 ## Updating a Classifier
 
-First we run the training scripts. This will increment the version of the classifier in the projects registry in weights and biases. You can then run the promote script which is used for promoting a model to primary within or cross aws environment. Promotion will not affect the version / alias.
+First we run the training scripts. This will upload the classifier to s3 and link to it from its weights and biaas project. You can then run the promote script which is used for promoting a model to primary within an aws environment. Promotion as adds the model to the weights and bias registry and setting it as primary gives it the environment alias.
```
THOR300, line 10: `biases`

scripts/promote.py
olaughter: I'm not sure thats lcear and would rather avoid putting an implementation detail in the user facing abstraction so I will leave this one be

scripts/README.md
olaughter: Good catch

PR #582, opened 2025-08-06
chore(deps): Add Prefect-Coiled

jesse-c:
So we can use it in Prefect.

Builds on #559.

TOWARDS PLA-758


olaughter (approved):
Lets gooooooo

PR #586, opened 2025-08-07
refactor: Use uv

jesse-c:
This helps with the issue of Transformers and Torch not installing well.

It's also a continuation of our organisational migration.

Other changes:

- https://ai.pydantic.dev/api/agent/#pydantic_ai.agent.AgentRun.usage shows that `AgentRunResult` has an `output` field, not `data`.
- The Docker GitHub Actions are used to get better support in Actions, such as for caching
- A non-Prefect image is used, so we don't need to keep it up-to-date. The same version of Prefect from our dependencies is used
- A `.dockerignore` file has been added to avoid things, when doing `COPY . .`
- I cleaned up the tests' snapshots
- Pre-commit is updated. It's actually quite nice, as it'll catch missed changes if someone didn't run `uv lock`
- Docker image building is now tested as part of CI

Documentation used:

- https://docs.astral.sh/uv/guides/integration/github/
- https://docs.astral.sh/uv/guides/integration/docker/

TOWARDS PLA-786
TOWARDS PLA-772


jesse-c:

- #586 ◀
    - #587

.github/workflows/ci.yml
```diff
@@ -79,11 +78,58 @@ jobs:
         run: just vespa_dev_setup
 
       - name: Run tests
-        run: just test -v -m "'not transformers'"
+        run: just test
```
THOR300, line 81: nice!

.github/workflows/ci.yml
```diff
@@ -79,11 +78,58 @@ jobs:
...
+
+      - name: Build Docker image
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          push: false
+          load: true
+          tags: dummy/${{ github.event.repository.name }}:${{ steps.get_version.outputs.version }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
```
THOR300, line 122: **thought**: This seems like a nice change, now that we're no longer using the just commands for build and push in CI wondering whether they can updated. 
`push-image` could be removed and `build-image` could be updated so we don't have to set env vars to build locally.

pyproject.toml
```diff
@@ -1,95 +1,81 @@
...
-[tool.poetry.scripts]
+[project.scripts]
 train = "scripts.train:app"
 promote = "scripts.promote:app"
 demote = "scripts.demote:app"
 evaluate = "scripts.evaluate:app"
 infer = "scripts.infer:app"
-update_inference_classifiers = "scripts.update_classifier_spec:app"
+update-inference-classifiers = "scripts.update_classifier_spec:app"
 deploy = "scripts.deploy:app"
```
THOR300, line 77: **question**: Have a validated locally that these scripts still run with this uv definition?

THOR300 (approved):
Great - approving subject to passing test.

.github/workflows/ci.yml
jesse-c: Those could be reasonable changes. I'll leave them for now, and they could be done later!

PR #587, opened 2025-08-07
refactor: Use Python v3.12.x

jesse-c:
I first tried 3.11.x, then thought I'd shoot for the moon, and try 3.12.x.

**Test**

- https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/06894ed1-f904-72a9-8000-d7743b8b9292 — Still failed with the same transformers related issue. A non-transformer inference worked.
- https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/06894f05-2f64-789a-8000-021cb7c5949e — Succeeded, with just a simple concept.

FIXES PLA-786
TOWARDS PLA-772


THOR300 (approved):
LGTM (also subject to passing tests in the base branch)! 

I guess we could refactor to make use of some of the features that come with version 12 but no need to prioritise that now. 

Once such nice change from reading around would be using generics in classes with no TypeVar definition: 

```python 
@dataclass
class Ok[T]:
    """Represents a successful result containing a value of type T."""

    _value: T

    def __str__(self):
        """Return string representation of the Ok result."""
        return f"Ok({self._value})"
```

jesse-c:
> LGTM (also subject to passing tests in the base branch)!
> 
> I guess we could refactor to make use of some of the features that come with version 12 but no need to prioritise that now.
> 
> Once such nice change from reading around would be using generics in classes with no TypeVar definition:
> 
> ```python
> @dataclass
> class Ok[T]:
>     """Represents a successful result containing a value of type T."""
> 
>     _value: T
> 
>     def __str__(self):
>         """Return string representation of the Ok result."""
>         return f"Ok({self._value})"
> ```

Thank goodness, haha.

PR #588, opened 2025-08-07
add the add_alternative_labels method back to wikibasesession

harrisonpim:
I removed this method from the class in the last refactor because it hadn't been used in months, but the very next day, @AnneIsARealProgrammerNow came up with a request that made it necessary again 😂 

Now it's back, but fits the async-first pattern used in the rest of the class


jesse-c (approved):
Good timing! 😆

PR #590, opened 2025-08-07
fix nasty event loop conflicts in the labelling librarian

harrisonpim:
Despite all of the work that we've done over the last week or so, the labelling_librarian was still throwing `RuntimeError: Event loop is closed` errors 😖

Screenshot 2025-08-07 at 20 30 35

After a bit of digging, it seems like the error comes from the caching of calls to `WikibaseSession.get_concept()` - the cache was sync, but the calls are run async. That, in combination with prefect's own event loop management, seemed enough to tie the whole system in knots...

The other static sites have been generating just fine under the new async pattern, so I've switched to use the async-first pattern in `labelling_librarian` too. With any luck, this should _finally_ get the vibe checker and other tools publishing nightly again 🤞

### DIFF
--- a/static_sites/concept_librarian/template.py
+++ b/static_sites/concept_librarian/template.py
@@ -4,7 +4,6 @@ from pathlib import Path
 from jinja2 import Environment, FileSystemLoader
 
 from src.concept import Concept, WikibaseID
-from src.wikibase import WikibaseSession
 from static_sites.concept_librarian.checks import (
     ConceptIssue,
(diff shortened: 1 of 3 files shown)


THOR300 (approved):
LGTM

THOR300:
The test that failed during CI is also failing on main. We're merging this PR to unblock the policy team. Discussion on slack [here](https://climate-policy-radar.slack.com/archives/C06JZKM6PGQ/p1754899211783919).

PR #592, opened 2025-08-11
test: Use stable mangled data

jesse-c:
We started seeing failures on GitHub runners[^1]. Why did it start? Who knows what's changed in the runners themselves.

There's several fixtures for a document with that stem (AKA ID), and they don't all have the same number of labelled passages. The string matching depends on a stable number of labelled passages.

All fixtures:

```
tests/flows/fixtures/labelled_passages/Q223/v3/CCLW.executive.10061.4515.json
tests/flows/fixtures/labelled_passages/Q767/v3/CCLW.executive.10061.4515.json
tests/flows/fixtures/labelled_passages/Q1286/v3/CCLW.executive.10061.4515.json
tests/flows/fixtures/labelled_passages/Q123/v4/CCLW.executive.10061.4515.json
tests/flows/fixtures/labelled_passages/Q218/v5/CCLW.executive.10061.4515.json
```

`tests/flows/fixtures/labelled_passages/Q218/v5/CCLW.executive.10061.4515.json`
has 25.

To avoid this, use specific classifiers, and the associated fixtures, to force
an error.

FIXES PLA-787

[^1]: https://github.com/climatepolicyradar/knowledge-graph/actions/runs/16875588172/job/47799425877

PR #593, opened 2025-08-11
fix: Surface total failure of inference

jesse-c:
If inference entirely failed, don't simply return, as it gives the _full
pipeline_ and overall succeeded state, which is incorrect.

FIXES PLA-788


jesse-c:

- #592
    - #593 ◀

PR #595, opened 2025-08-11
feat: Allow GPU inference

jesse-c:
TOWARDS PLA-759


flows/inference.py
```diff
@@ -933,6 +932,49 @@ async def inference_batch_of_documents(
...
+        batch,
+        config_json,
+        classifier_name,
+        classifier_alias,
+    )
+
+
+@flow(log_prints=True, result_storage=S3_BLOCK_RESULTS_CACHE)
+@coiled.function(  # pyright: ignore[reportUnknownMemberType]
+    vm_type=[
```
jesse-c, line 957: **note:** When testing Coiled, I noticed I couldn't get availability sometimes. So, here's a list we'll accept.

flows/utils.py
```diff
@@ -47,6 +47,8 @@
 T = TypeVar("T")
 U = TypeVar("U")
 
+JsonDict = NewType("JsonDict", dict[str, Any])
```
jesse-c, line 50: **note:** We use this type in so many places. Thought I'd chuck this here. Pydantic has a `Json` type, but it's for JSON as a serialised string. I sew

deployments.py
```diff
@@ -78,21 +88,49 @@ def create_deployment(
     docker_repository = os.getenv("DOCKER_REPOSITORY", PROJECT_NAME)
     image_name = os.path.join(docker_registry, docker_repository)
 
-    default_variables = JSON.load(f"default-job-variables-prefect-mvp-{aws_env}").value
+    if gpu:
+        work_pool_name = f"coiled-{AwsEnv(os.getenv('AWS_ENV', 'sandbox'))}"
+        work_queue_name = "default"
+        # TODO: Use the Block created via the orchestrator's IaC
```
jesse-c, line 94: **note:** Depends on: https://github.com/climatepolicyradar/orchestrator/pull/235.

flows/inference.py
jesse-c: **thought:** Currently the classifier spec. v2 only has GPU as a boolean, but perhaps one day it could be more specific.

deployments.py
```diff
@@ -63,9 +64,18 @@ def get_schedule_for_env(
         return None
 
 
+# Match what Prefect uses for Flows:
+#
+# > .. we use the generic type variables `P` and `R` for "Parameters"
+# > and "Returns" respectively.
+P = ParamSpec("P")
+R = TypeVar("R")
```
THOR300, line 72: **thought**: As we've upgraded the python version now I don't _think_ we need these top level variables to use generics (although the description does add something)

THOR300 (commented):
This is real nice! 

Regarding testing - that sounds like the right approach chasing it up with coiled regarding local solutions! Reckon it’s worth ticketing? Wondering whether you can mock dask clusters or ec2 instances locally to facilitate this 🤔 .

On the question of why use ECS - yeh we absolutely could not use ECS and potentially with our plan this wouldn't incur any additional cost! It would be good first to estimate how many cpu hours we might use as I think that would be the limiter. 

[This](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/0689d1ec-7714-7fd1-8000-541ec4a6d8f4?entity_id=0689d206-49fb-7a10-8000-d265b9d37e04&state=pending) staging test looks like we're getting some errors?

jesse-c:
> This is real nice!
It's come together relatively well! Also, I have been looking at the clusters in Coiled to see logs, and I am also seeing lots of logs in Prefect too.
> 
> Regarding testing - that sounds like the right approach chasing it up with coiled regarding local solutions! Reckon it’s worth ticketing? Wondering whether you can mock dask clusters or ec2 instances locally to facilitate this 🤔 .
TBD! I'll see what comes out of my thread with Coiled.
> On the question of why use ECS - yeh we absolutely could not use ECS and potentially with our plan this wouldn't incur any additional cost! It would be good first to estimate how many cpu hours we might use as I think that would be the limiter.
True, that would be the decider. TBD too then. Still TBD (x3) on using Prefect serverless in our plan too.
> [This](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/0689d1ec-7714-7fd1-8000-541ec4a6d8f4?entity_id=0689d206-49fb-7a10-8000-d265b9d37e04&state=pending) staging test looks like we're getting some errors? Happy to support fixing these before merge.
That failure is to do with the IAM access. I'm testing the work for that over in https://github.com/climatepolicyradar/orchestrator/pull/236. 🙇🏻 I need to run it via Pulumi next.

deployments.py
jesse-c: This was a **very** interesting reminder! 🙏🏻 I read https://realpython.com/python312-typing/. I don't _think_ that the new changes in Python 3.12 can help with that, but, it did prompt me to make the helpful changes elsewhere! I removed the `T = TypeVar("T")`[^1].

[^1]: [`78cbb8e` (#595)](https://github.com/climatepolicyradar/knowledge-graph/pull/595/commits/78cbb8e9e7182df112ada3cc4aa932d5dcb1fd2b)

flows/inference.py
jesse-c: **done:** Moved it to `flows.utils`.

THOR300 (approved):
Huge! 

Happy to approve with the staging runs still broken to unblock development and get what we have here on to main. This gpu deployment also isn't being utilised in the KG pipeline yet so no risk 👍

PR #599, opened 2025-08-13
Feat: Run full pipeline on new documents

THOR300:
This Pull Request: 
---
- Adds an automation for triggering the `full_pipeline` when the s3 backup is run. 
- Currently disabled until we fix the targets classifiers.

### DIFF
--- a/automations.py
+++ b/automations.py
@@ -84,7 +84,7 @@ def create_target_automation(
             # > event.
             #
             # NB: Currently a set due to the Prefect Python class used
-            expect=set(["prefect.flow-run.Completed"]),
+            expect=set(["prefect.flow-run.Running"]),
         ),
         actions=[
             RunDeployment(
--- a/automations.py
+++ b/automations.py
@@ -144,6 +144,7 @@ async def a_triggers_b(
     description: str,
     ignore: list[AwsEnv],
     aws_env: AwsEnv,
+    expect_state: StateType = StateType.COMPLETED,
 ) -> None:
     """Automation to after Deployment A completes trigger Deployment B."""
     client = get_client()
@@ -190,6 +191,7 @@ async def a_triggers_b(
         description=description,
         parameters=b_parameters,
         enabled=enabled,
+        expect_state=expect_state,
     )
 
     print("deleting, if already exists...")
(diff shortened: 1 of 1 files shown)


automations.py
```diff
@@ -207,10 +207,10 @@ async def main() -> None:
...
         a_deployment_name=f"navigator-data-s3-backup-pipeline-cache-{aws_env}",
-        b_flow_name=inference.name,
-        b_deployment_name=generate_deployment_name(inference.name, aws_env),
-        b_parameters={"use_new_and_updated": True},
-        description="Start concept store inference with classifiers.",
+        b_flow_name=full_pipeline.name,
+        b_deployment_name=generate_deployment_name(full_pipeline.name, aws_env),
+        b_parameters={"inference_use_new_and_updated": True},
+        description="Start the knowledge graph full pipeline.",
         enabled=False,
```
jesse-c, line 214: **question:** When would it be enabled?

automations.py
```diff
@@ -84,7 +84,7 @@ def create_target_automation(
             # > event.
             #
             # NB: Currently a set due to the Prefect Python class used
-            expect=set(["prefect.flow-run.Completed"]),
+            expect=set(["prefect.flow-run.Running"]),
```
jesse-c, line 87: **fix:** Could you make this configurable, with the default behaviour being on completion. There's other automations that may use this (as there was previously), and on completion was the common use case.

jesse-c (commented):
First pass.

automations.py
jesse-c: lmao I missed that, my apologies.

automations.py
```diff
@@ -207,10 +207,10 @@ async def main() -> None:
...
         a_flow_name="navigator-data-s3-backup",
         a_deployment_name=f"navigator-data-s3-backup-pipeline-cache-{aws_env}",
-        b_flow_name=inference.name,
-        b_deployment_name=generate_deployment_name(inference.name, aws_env),
-        b_parameters={"use_new_and_updated": True},
-        description="Start concept store inference with classifiers.",
+        b_flow_name=full_pipeline.name,
+        b_deployment_name=generate_deployment_name(full_pipeline.name, aws_env),
+        b_parameters={"inference_use_new_and_updated": True},
+        description="Start the knowledge graph full pipeline.",
```
jesse-c, line 213: **question:** You may want to chuck in `expect_state=running?` in here, now that you've done the other bit!

jesse-c (approved):
LGTM

automations.py
THOR300: 🫠

PR #609, opened 2025-08-18
Aggregate Flow uses a common flows Config

michaelokarimia:
Initial step to refactor to use one config for all flows


flows/pipeline_config.py
olaughter, line 1: suggestion: you could just call this file `config.py` for brevity, and as its in the flows directory its already implied its for a pipeline

michaelokarimia: Sure I'll do that rename right at the end once I've completed the refactor

PR #610, opened 2025-08-19
ci: Avail space for Docker image building

jesse-c:
This has arisen post-migrating to uv with Python 3.12, utilising a GH cache for the images.

- Removing unused software (available space before 27G, after: 44G). This is done via background jobs, to significantly speed it up—from a few minutes to near instantaneous
- Don't install Python and uv outside of the Docker image
- Don't copy everything carte blanche
- Let building start immediately, since it's the longest job to run
- Do more post-build checks

**Test**

Built successfully: https://github.com/climatepolicyradar/knowledge-graph/actions/runs/17066912735/job/48386151210?pr=610. There's been others after this too.

I've cleaned out our [caches](https://github.com/climatepolicyradar/knowledge-graph/actions/caches) (`$ gh cache list`):

CleanShot 2025-08-19 at 10 48 36

I'm not sure if the loading of the cache contributed to the high disk usage.

**Appendix**

- In the future, we may want to have separate CPU and GPU images.
- Loading the image, to run the post-build checks, takes a bit of time (` DONE 79.8s`). We could remove them, since we have pretty high confidence in the Dockerfile. We lived without them, until recently. They could at least be put into 1 script, though I think the time saved would be negligible.
- I have experimented with multi-layer images to reduce the final image size. It's still TBD.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,6 @@ jobs:
   build:
     runs-on: ubuntu-latest
     name: "Build"
-    needs: [lint]
     steps:
       - name: Check out repository code
         uses: actions/checkout@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,8 +134,16 @@ jobs:
       - name: Confirm Prefect binary availability
         run: docker run --rm dummy/${{ github.event.repository.name }}:${{ steps.get_version.outputs.version }} prefect version
 
+      - name: Confirm Coiled binary availability
+        run: docker run --rm dummy/${{ github.event.repository.name }}:${{ steps.get_version.outputs.version }} coiled --version
+
       - name: Confirm dependencies availability
         run: docker run --rm dummy/${{ github.event.repository.name }}:${{ steps.get_version.outputs.version }} python -c "import aioboto3; print('aioboto3 imported successfully')"
+
+      # AWS_ENV is needed, for now, since there's a module with a
+      # side-effect that calls the OS environment.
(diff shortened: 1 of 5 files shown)


michaelokarimia (approved):
Nice one!

.github/workflows/ci.yml
```diff
@@ -83,27 +83,37 @@ jobs:
...
 
+      - name: Remove unused software
+        # This is a "works for now" step. The underlying runner may
+        # change. Don't block CI on this.
+        continue-on-error: true
+        run: |
+          # Space before
+          df -h
+          # .Net
+          sudo rm -rf /usr/share/dotnet || true &
```
michaelokarimia, line 98: We certainly don't need this and Andriod! Haha

jesse-c: It was *so* big too.

Dockerfile
```diff
@@ -1,26 +1,23 @@
...
-
-# Download the latest installer
-ADD https://astral.sh/uv/install.sh /uv-installer.sh
-
-# Run the installer then remove it
-RUN sh /uv-installer.sh && rm /uv-installer.sh
-
-# Ensure the installed binary is on the `PATH`
-ENV PATH="/root/.local/bin/:$PATH"
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
```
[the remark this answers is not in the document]
jesse-c: It was already downloading the installer, so if the concern is about remote reliance, it's no better or worse. I wonder if it's better, as Docker's registries are stable?

PR #612, opened 2025-08-19
fix: typo for stale workflow env variable

olaughter:
Looks like this is working! [Example PR marked as stale](https://github.com/climatepolicyradar/knowledge-graph/pull/369#issuecomment-3173910583), when checking that comment I realised this variable was mistyped.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -12,8 +12,8 @@ jobs:
     steps:
       - uses: actions/stale@v9
         with:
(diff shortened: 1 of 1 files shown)


jesse-c (approved):
LGTM. Would be neat if the Action let you use the declared `days-before-pr-sale` value.

olaughter:
> LGTM. Would be neat if the Action let you use the declared `days-before-pr-sale` value.

Actually does look like it could be [just a 1 hour contribution](https://github.com/actions/stale/blob/8f717f0dfca33b78d3c933452e42558e4456c8e7/src/main.ts#L10-L54) to make... 👀

PR #614, opened 2025-08-20
Make use of the BaseParserOutput method to get text blocks.

THOR300:
This Pull Request: 
---
- Is a bug fix for the function to retrieve document passages during the KG pipeline. 
- Prefect flow [run](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/068a5ab0-0f44-73c9-8000-370c357f9520?entity_id=068a5ab0-0f44-73c9-8000-370c357f9520&state=pending).
- When we look at the content of the parser output json in the embeddings input directory we see the cause of the issue. The content type is html but the data is pdf. This is due to changes earlier this year where we convert html docs to pdfs to extract text. 
image

### DIFF
--- a/flows/inference.py
+++ b/flows/inference.py
@@ -408,18 +408,8 @@ def document_passages(
     document: BaseParserOutput,
 ) -> Generator[tuple[str, str], None, None]:
     """Yield the text block irrespective of content type."""
-    match document.document_content_type:
-        case "application/pdf":
-            text_blocks = document.pdf_data.text_blocks  # type: ignore
-        case "text/html":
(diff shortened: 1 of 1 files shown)


THOR300:
> Approved, but are there any tests for this?

Hey hey - the pydantic model has tests where it's declared [here](https://github.com/climatepolicyradar/cpr-sdk/blob/main/tests/test_parser_models.py#L33).

PR #615, opened 2025-08-20
Feature/pla 709 reuse config objects for flows

michaelokarimia:
Changed to make only one common [config.py](../blob/main/flows/config.py) for all Prefect deployment flows, by merging the  configs used in [aggregate](../blob/main/flows/aggregate.py) and [inference](../blob/main/flows/inference.py) and [index](../blob/main/flows/index.py) together into a single file.

Removes some unnecessary validation that used to compare the multiple configs, which are now the same file
Towards PLA-612 as I had to tweak the conftest.py file


tests/flows/test_full_pipeline.py
```diff
@@ -10,124 +10,23 @@
...
-                cache_bucket="test",
-                bucket_region="test",
-                document_source_prefix="test",
-                aws_env=AwsEnv.production,
-            ),
-            "AWS environment mismatch",
-        ),
-    ],
-)
-def test_validate_aggregation_inference_configs_raises_value_error(
```
michaelokarimia, line 110: Wondering if I should refactor this validation check to compare if document target and document source are the same. They are now separate parameters in the shared config

olaughter: That's a really good idea, feel free to follow up afterwards so that you arent blocked on merging for it

olaughter (approved):
Great to see this tidy up

flows/full_pipeline.py
```diff
@@ -217,7 +163,7 @@ async def full_pipeline(
         document_stems=list(
             inference_result.fully_successfully_classified_document_stems
         ),
-        config=aggregation_config,
+        config=config,
```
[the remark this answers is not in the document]
michaelokarimia: Good shout, I've address this now

flows/config.py
```diff
@@ -0,0 +1,95 @@
...
+        description="s3 prefix for where inference targets are written to",
+    )
+
+    bucket_region: str = Field(
+        default="eu-west-1", description="AWS region for S3 bucket"
+    )
+    aws_env: AwsEnv = Field(
+        default_factory=lambda: AwsEnv(os.environ["AWS_ENV"]),
+        description="AWS environment",
+    )
```
jesse-c, line 45: **fix:** Having this side-effect of reaching out to the OS is annoying and frail, and I've wanted it to be fixed for a while.

It means that this module, and ones that import it, can't just be imported without having set the AWS environment environment variable, leading to having to fake it at different times.

Most lately, with validating the Docker image, when ensuring that our package has been installed. 

It should be read when an instance of the class is initialised.

This is like how we don't do a call to SSM to get the W&B key in the class definition, but at instantiation time.

flows/config.py
```diff
@@ -0,0 +1,95 @@
...
+
+    pipeline_state_prefix: str = "input"
+    local_classifier_dir: Path = Path("data") / "processed" / "classifiers"
+    wandb_model_org: str = "climatepolicyradar_UZODYJSN66HCQ"
+    wandb_model_registry: str = "wandb-registry-model"
+    wandb_entity: str = "climatepolicyradar"
+    wandb_api_key: Optional[SecretStr] = None
```
jesse-c, line 52: **polish:** Add `Field`s for all of these, with descriptions, as they have above.

flows/config.py
```diff
@@ -0,0 +1,95 @@
...
+                "pipeline_cache_bucket_name"
+            )
+
+        if not config.wandb_api_key:
+            config.wandb_api_key = SecretStr(get_aws_ssm_param("WANDB_API_KEY"))
+
+        return config
+
+    @property
+    def cache_bucket_str(self) -> str:
```
jesse-c, line 69: **polish:** This is the only field, out of all the fields that are optional, that has this check. I know this is because we don't want people to have to get those values, like W&B, themselves, and simply because we didn't create function like this for the other ones.

I haven't done it for a while, but, Pydantic has a lot of flexibility in validators.

None of those values should be optional. The `Config` class should instaniation should be so that it's not needed. For example, a _before validator_[^1] could be used.

This is really helpful for type checking, since them being `None` is never something we'd accept, but the typer checker doesn't know that, so it dutifully complains that a value may be `None`.

[^1]: https://docs.pydantic.dev/latest/concepts/validators/#field-validators

flows/config.py
```diff
@@ -0,0 +1,95 @@
...
+        """Return the cache bucket, raising an error if not set."""
+        if not self.cache_bucket:
+            raise ValueError(
+                "Cache bucket is not set in config, consider calling the `create` method first."
+            )
+        return self.cache_bucket
+
+    def to_json(self) -> dict:
```
jesse-c, line 77: **fix:** I don't know why we ever created this manually, when Pydantic has this built-in.

It's used with [`"config_json": config.to_json()`](https://github.com/climatepolicyradar/knowledge-graph/blob/fix%2Fstop-docker-cache-explosion/flows/inference.py#L1055), but even that then uses the Pydantic deserialisation with [`config = Config(**config_json)`](https://github.com/climatepolicyradar/knowledge-graph/blob/fix%2Fstop-docker-cache-explosion/flows/inference.py#L815).

Pydantic's built-in JSON serialisation should be used.

flows/config.py
```diff
@@ -0,0 +1,95 @@
...
+        default=AGGREGATE_DOCUMENT_SOURCE_PREFIX_DEFAULT,
+        description="S3 prefix for source documents are read from",
+    )
+    aggregate_inference_results_prefix: str = Field(
+        default=INFERENCE_RESULTS_PREFIX,
+        description="S3 prefix for aggregated inference results are written to",
+    )
+    inference_document_source_prefix: str = Field(
+        default=INFERENCE_DOCUMENT_SOURCE_PREFIX_DEFAULT,
+        description="s3 prefix of documents read as source for inference",
```
jesse-c, line 31: **polish:** To align with the others, and to be correct spelling.
```suggestion
        description="S3 prefix of documents read as source for inference",
```

flows/config.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+
+    inference_document_target_prefix: str = Field(
+        default=INFERENCE_DOCUMENT_TARGET_PREFIX_DEFAULT,
+        description="s3 prefix for where inference targets are written to",
```
jesse-c, line 36: **polish:** ↑
```suggestion
        description="S3 prefix for where inference targets are written to",
```

flows/full_pipeline.py
```diff
@@ -28,56 +26,15 @@
...
-            + f"{inference_config.aws_env}"
-        )
-
-
 async def inference_with_result_cache(
     classifier_specs: Sequence[ClassifierSpec] | None = None,
     document_ids: Sequence[DocumentImportId] | None = None,
     use_new_and_updated: bool = False,
-    config: InferenceConfig | None = None,
+    pipeline_config: Config | None = None,
```
jesse-c, line 37: **fix:** This change seems unnecessary, considering elsewhere isn't changed to have the new prefix. Even if that did happen though, I think it just adds noise. There aren't non-pipeline configs?

This applies too to where _pipeline_ was added, like at least 1 print statement.

I wonder if your motivation was because of this awfully named pipeline, that has a superfluous _pipeline_ in the name?

jesse-c (commented):
Some key changes that I think would be good to have.

Have you validated that Prefect is able to render these in the webapp?

flows/full_pipeline.py
michaelokarimia: Nope, it's just a typo, thanks for flagging

flows/config.py
michaelokarimia: Would it make sense to default it as a `str` e.g. `sandbox` and then set it later?

michaelokarimia:
Addressed these outstanding issues in #624

flows/config.py
jesse-c: I think that'd be a less ideal way than updating the model to get the value at instantiation. It doesn't solve the problem, it just shifts it around, IMO. We do have some default AWS env. values around, but it's generally avoided, especially in the pipelines.

PR #622, opened 2025-08-21
Reorganise config in scripts and src directories

harrisonpim:
As this project has grown, we've developed a few messy patterns for managing configuration across different concerns 

There are lots of dangling bits of config-adjacent code which belong in `src` but live in `scripts`, some which belong in `scripts` but live in `src`, and some which should be stored at a much more general, higher level. 

This PR isn't comprehensive (there's a lot of stuff in `flows` which i've left untouched), but goes some way towards addressing these issues:
- moves a lot of static config from `scripts/config` to `.env`
- moves the remaining dynamic `scripts/config` to `src/config`
- reorganises a few modules to sit in more appropriate places
  - `scripts/cloud` → `src/cloud`: this is more of a module than a script, and belongs in the src dir
  - `scripts/utils` → `src/classifier`: The `ModelPath` class and `get_local_classifier_path` function are only related to classifiers and feel more appropriate alongside them than in a generic `utils`


.env.example
```diff
@@ -16,11 +16,15 @@ ARGILLA_API_KEY=
...
 AWS_PROFILE=labs
 
 DOCKER_REGISTRY=
 DOCKER_REPOSITORY=
 
 WANDB_API_KEY=
+
+DATA_DIR_NAME=data
+MODEL_ARTIFACT_NAME=model.pickle
+EQUITY_COLUMNS=translated,world_bank_region,document_metadata.corpus_type_name
```
olaughter, line 30: question: Are these likely to change by environment or local setup? I wonder if they belong in the static config to keep the env simpler?

src/classifier/__init__.py
```diff
@@ -1,10 +1,39 @@
...
+    def __fspath__(self) -> str:
+        """Return the filesystem path representation for use with pathlib."""
+        return str(self)
+
+
+def get_local_classifier_path(target_path: ModelPath, version: str) -> Path:
+    """Returns a path for a classifier file."""
+    from src.config import classifier_dir, model_artifact_name
+
+    return classifier_dir / target_path / version / model_artifact_name
```
olaughter, line 36: *subjective nitpick:* I imagine if I was looking for these I'd likely be more likely to check `classifier/classifier.py` first, wondering if that would be a better home for them?

.env.example
olaughter: Just seen the default arguments in config and realising that they are set there but can be overridden here so not sure my comment really applies. On a smaller level, it does still make me wonder if the example doesn't need them mind

tests/flows/test_full_pipeline.py
```diff
@@ -19,7 +19,101 @@
     InferenceResult,
 )
 from flows.utils import DocumentImportId, DocumentStem, Fault
-from scripts.cloud import ClassifierSpec
+from src.cloud import AwsEnv, ClassifierSpec
+
+
+def test_validate_aggregation_inference_configs() -> None:
```
olaughter, line 25: This has changed under you with some config refactors 😄

olaughter (approved):
Neat! Love to see it

olaughter:
Happy to be asked for a re-review if the merge conflict ends up changing things substantially and you'd prefer another sweep

src/classifier/__init__.py
olaughter: That makes sense, and yes that would make it clearer (no pressure to change it anytime soon though)

PR #632, opened 2025-08-26
Combine the keyword- and rules-based classifiers

harrisonpim:
## why?

The `RulesBasedClassifier` class has always been a bit of a silly contrivance. It's a subclass of the `KeywordClassifier` which uses the same mechanism to handle negative labels (if a concept has a negative label, the classifier will ignore any positive keyword matches which contain the negative label).

When I originally created the `RulesBasedClassifier`, I hoped/assumed that we would gradually incorporate more and more special logic/rules into this class. In practice, it turns out that positive and negative keywords are basically all you need in 99% of cases 🤷 

A year on, I don't think we need to keep both classes. IMO the `KeywordClassifier` should handle negative keywords by itself and there should be no ambiguity about whether you should use a `KeywordClassifier` vs `RulesBasedClassifier`. 

This PR updates the `KeywordClassifier` to add the negative labels functionality, and deletes the `RulesBasedClassifier`.

## a note on backward compatibility of classifier code

when running the test suite with new classifier code, I saw a few errors from `test_inference`, eg:

```
FAILED tests/flows/test_inference.py::test_text_block_inference_with_results - AttributeError: 'KeywordClassifier' object has no attribute 'case_sensitive_positive_pattern'. Did you mean: 'case_insensitive_pattern'?
```

The saved classifier artefact used by the inference flow contains the old `KeywordClassifier` attribute names, `case_sensitive_pattern` and `case_insensitive_pattern`, which have been replaced by `case_sensitive_positive_pattern` in this PR.

To get around this issue within the context of this PR, i've created and saved a fresh version of the classifier:

```py
from src.classifier.keyword import KeywordClassifier

wikibase = WikibaseSession()
concept = wikibase.get_concept("Q788")
classifier = KeywordClassifier(concept)
classifier.save("tests/flows/fixtures/classifiers/Q788")
```

Running the code with that newly saved artefact turns everything green again ✅

But I'm wondering whether this hints at a broader code/artefact versioning/compatibility issue... If I remember correctly, @THOR300 recently saw a similar issue when working with a targets model that had been saved using a different version of the huggingface transformers library.

If we're making changes to our code or updating our dependencies, do we need to enforce compatibility with our saved classifier artefacts? Or should we retrain every classifier every time we (or our dependencies) push a breaking change? Should these principles apply to local artefacts and/or those in wandb, as in @THOR300's example? Have any of us thought about this issue deeply enough? I know I haven't 😬


jesse-c (approved):
I was wondering if we needed both, when I was doing the performance investigation the other day! Nice to see this now.

On the Slack thread, I said I don't think this would need retraining, but now that I see it, I might've been wrong. Could you test loading a saved `RulesBasedClassifier`, after it having being removed from the code, to see if we can still do inference?

The `transformers` issue wasn't about compatibility within how models are used in KG, necessarily. These models aren't dependent external packages. Retraining would be needed if interfaces/classes change though.

THOR300:
That is a weirdly similar attribute error haha! 

My understanding of the question is: 

_"Is it okay to make changes that introduce backwards compatibility issues with existing classifiers?"_ 

I'd say yes and lean towards not investing in backwards compatibility for the saved artefacts as retraining is quick and straightforward for us and we don’t expect to run old classifiers alongside newer ones. Iteration speed is also key.

PR #633, opened 2025-08-26
use the new p11 property to fetch more negative labels

harrisonpim:
The policy team have asked for a property in wikibase which will allow them to include all positive labels from one concept as negative labels on another.

For example, maybe we want to add all of the labels on [greenhouse gas (Q218)](https://climatepolicyradar.wikibase.cloud/wiki/Item:Q218) as negatives on  [fossil gas (Q650)](https://climatepolicyradar.wikibase.cloud/wiki/Item:Q650). Until now, they've needed to do that manually, label by label (see the crazy long list of negative labels on Q650). Instead, we could just let them link to Q218 and let the code handle the rest.

I've created a property [P11](https://climatepolicyradar.wikibase.cloud/wiki/Property:P11) which (with this PR) should achieve that!

testing this:

```py
wikibase = WikibaseSession()
concept = wikibase.get_concept("Q650")
print(concept.negative_labels)
```

```
['gas chromatography', 'organic gas', 'carbon gas', 'greenhouse gases', 'GHGs', 'gas turbine', 'clean gas', 'CO2e', 'green house gases', 'GHG', 'CO₂e', 'land fill gas', 'carbon dioxide equivalent', 'green house gas', 'carbon gases', 'greenhouse gas', 'landfill gas', 'generator gas', 'biogas', 'cooking gas']
```

labels like "GHG" are now pulled from the `negative_concept` `Q218` 🎉


jesse-c (commented):
LGTM. It's a nice feature!

**thought:** With the model hot swapping, I'm so far making concepts, models, and inference results revision aware, so there's lineage for the revision of the concept used. I'm taking note that there's place, like here, where I may need to track the revision.

jesse-c (approved):

PR #634, opened 2025-08-27
Bugfix: Inference scaling errors 

THOR300:
This Pull Request: 
---
- Is a bug fix for some issues that were identified when running the inference stage of the knowledge graph full pipeline at scale. 
- Logs created from metadata where too long to send to the prefect api. 

> String should have at most 25000 characters

- The number of assets materialised during inference exceeded prefects max limit. 

> The maximum number of related resources is 100

_Note: Lot's of details on the investigation on the ticket | Not running a staging test as this only crops up at scale | Auto-merge enabled._


flows/utils.py
```diff
@@ -774,7 +774,14 @@ def __str__(self) -> str:
         except Exception as e:
             print(f"could not represent fault's data as a string: {e}")
             data_str = ""
-        return f"{self.msg} | metadata: {json.dumps(self.metadata, default=str)} | data: {data_str}"
+
+        fault_str = f"{self.msg} | metadata: {json.dumps(self.metadata, default=str)} | data: {data_str}"
+
+        # Prefect logs should have no more than 25000 characters, so truncate
+        # the fault string if it's too long.
+        fault_str = fault_str[:20_000] + "..." if len(fault_str) > 20_000 else fault_str
```
jesse-c, line 782: **fix:** I think there's an inconsistency here where the comment says 25,000, but the check is for 20,000.

jesse-c, line 782: **polish:** There's a built-in function for this: https://github.com/climatepolicyradar/knowledge-graph/blob/main/flows/utils.py#L157-L161.
```suggestion
        fault_str = textwrap.shorten(fault_str, width=20_000, placeholder="...")
```

jesse-c:
> Is there any way that truncating the assets affects how the function is executed?
> 
> Looking at the `MaterialiseTask` object that is instantiated I don't see any issue with this. [PrefectHQ/prefect@`4a2335d`/src/prefect/tasks.py#L2102](https://github.com/PrefectHQ/prefect/blob/4a2335d9b05bb76459ecb36ba33ba605f1047e88/src/prefect/tasks.py#L2102)

Not that I can think of, no.

jesse-c (approved):

PR #635, opened 2025-08-27
fix: Version serialisation for pydantic 2

olaughter:
Fixes a bug with the way `wandb_registry_version` was serialised in the classifier specs. Version was still using pydantic 1 syntax for serialisation. This had led to classifier specs storing an integer instead of a string for versions and that in turn meant they could not be read as they failed validation.


src/version.py
```diff
@@ -58,9 +61,16 @@ def __lt__(self, other):
         return str(self) < other
 
     @classmethod
-    def __get_validators__(cls):
-        """Return a generator of validators for Pydantic compatibility."""
-        yield cls._validate
+    def __get_pydantic_core_schema__(
```
jesse-c, line 64: **thought:** I hope there's a better way than this in Pydantic v2.

jesse-c (approved):

src/version.py
olaughter: I think the ideal would be to convert Version to a pydantic model, but its used in a variety of ways and felt way out of scope for what Im working on

PR #638, opened 2025-08-27
Feature/pla 764 add source specific classifier selection 2

olaughter:
Create script to apply spec wide metadata updates, then use it to update all the envs with the sabin source restriction.

Following this I'll be able to update the logic in inference to use these specs.

### DIFF
--- a/flows/classifier_specs/v2/labs.yaml
+++ b/flows/classifier_specs/v2/labs.yaml
@@ -5,86 +5,142 @@
   wandb_registry_version: v12
   wikibase_id: Q123
 - classifier_id: 58br2fy7
+  dont_run_on:
+  - sabin
   wandb_registry_version: v17
   wikibase_id: Q368
 - classifier_id: 82s9rtjv
+  dont_run_on:
+  - sabin
   wandb_registry_version: v16
   wikibase_id: Q374
 - classifier_id: 67n38m49
+  dont_run_on:
+  - sabin
   wandb_registry_version: v16
   wikibase_id: Q404
 - classifier_id: qgz45jfc
+  dont_run_on:
+  - sabin
   wandb_registry_version: v20
   wikibase_id: Q412
 - classifier_id: z3j4g6px
+  dont_run_on:
+  - sabin
   wandb_registry_version: v15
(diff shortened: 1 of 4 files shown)


scripts/classifier_metadata.py
```diff
@@ -25,7 +27,55 @@
...
+                aws_env=aws_env,
+                update_specs=False,  # since we'll only update once all are done
+            )
+            futures.append(future)
+
+        for future in as_completed(futures):
+            future.result()
+
+    if update_specs:
+        get_all_available_classifiers([aws_env])
```
jesse-c, line 74: **polish:** This function needs a rename IMHO! Perhaps this, based on the function doc.
```suggestion
        refresh_all_available_classifiers([aws_env])
```

scripts/README.md
```diff
@@ -55,6 +55,12 @@ Clear the list to allow the classifier to run on anything
 just classifier_metadata Q123 abcd2345 sandbox --clear-dont-run-on
 ```
 
+For times when its necessary to update every live classifier that is mentioned in the spec for an environment, you can run the following:
```
jesse-c, line 58: **fix:** To keep the terminology consistent:
```suggestion
For times when its necessary to update every promoted classifier that is mentioned in the spec for an environment, you can run the following:
```

scripts/classifier_metadata.py
```diff
@@ -25,7 +27,55 @@
...
 @app.command()
-def main(
+def update_entire_env(
+    clear_dont_run_on: Annotated[
+        bool, typer.Option(help="Remove all existing items from dont_run_on")
+    ] = False,
+    add_dont_run_on: Annotated[
+        list[DontRunOnEnum] | None,
+        typer.Option(help="Adds a single item to the metadata."),
+    ] = None,
```
jesse-c, line 37: **polish:** Add a check that these are mutually exclusive.

jesse-c (approved):

scripts/classifier_metadata.py
olaughter: Agree. Will come back to this 👍

PR #640, opened 2025-08-27
Feature/pla 759 update inference to utilise coiled compute layer 1

olaughter:
Extends the update metadata script so gpus can be added, the script has been run and the outputs updated. For now this field is just a bool since we'll use the default coil vm, but we can muck around with this in the future if we would like to. 

See current draft inference implementation for drawing on these [in PR 602](https://github.com/climatepolicyradar/knowledge-graph/pull/602/files#diff-f2fbcdde09b032ff14f38a9893469fb71b65a7250901908d7f07c3a983c6be79R1031-R1036).

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -55,7 +55,7 @@ Clear the list to allow the classifier to run on anything
 just classifier_metadata Q123 abcd2345 sandbox --clear-dont-run-on
 ```
 
-Add a requirement for the classifier to run with a GPU (or use clear-require-gpu to remove and revert to using a cpu)
+Add a requirement for the classifier to run in a compute environment that has a GPU (or use clear-require-gpu to remove and revert to using a cpu)
 
 ```shell
 just classifier_metadata Q123 abcd2345 sandbox --add-require-gpu


scripts/classifier_metadata.py
```diff
@@ -22,6 +22,9 @@
 REGISTRY_NAME = "model"
 JOB_TYPE = "configure_model"
 
+type ComputeEnvironment = dict[str, str | int | bool]
```
THOR300, line 25: **thought**: We have a definition of a compute environment declared in this module, wondering whether it could be utilised or whether they could be combined? 

flows/classifier_specs/spec_interface.py:L18

scripts/README.md
```diff
@@ -55,6 +55,12 @@ Clear the list to allow the classifier to run on anything
 just classifier_metadata Q123 abcd2345 sandbox --clear-dont-run-on
 ```
 
+Add a requirement for the classifier to run with a GPU (or use clear-require-gpu to remove and revert to using a cpu)
```
THOR300, line 58: **thought**: _...in a GPU supported environment..._

THOR300 (approved):
Nice clean PR!

scripts/README.md
olaughter: Went with "run in a compute environment that has a GPU" for this

scripts/classifier_metadata.py
olaughter: This seemed promising to me on paper, but after looking into this a bit I'm realising I'm not a fan. The two have different shapes and are serving too different a purpose one suited for writing to wandb metadata the other for writing it in & out of yaml, I think it would lead to some convoluted entanglement so I'm leaving it be.

PR #641, opened 2025-08-27
Remove unused synchronous s3 operation function

michaelokarimia:
Towards PLA-720

### DIFF
--- a/flows/boundary.py
+++ b/flows/boundary.py
@@ -826,18 +826,3 @@ class ConceptModel(BaseModel):
         raise ValueError(
             f"Could not extract concept name from model name '{self.model_name}'"
         )
-
-
-def save_labelled_passages_by_uri(
-    document_object_uri: DocumentObjectUri,
-    labelled_passages: list[LabelledPassage],
-) -> None:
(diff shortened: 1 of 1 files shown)


flows/boundary.py
```diff
@@ -172,24 +171,6 @@ def get_vespa_search_adapter_from_aws_secrets(
     )
 
 
-def s3_object_write_text(s3_uri: str, text: str) -> None:
```
olaughter, line 175: Always pleased to see code getting deleted 😄

tests/flows/conftest.py
```diff
@@ -3,12 +3,12 @@
...
 import subprocess
 import xml.etree.ElementTree as ET
-from collections.abc import AsyncGenerator, Generator
+from collections.abc import Generator
 from contextlib import asynccontextmanager
 from datetime import datetime
 from io import BytesIO
 from pathlib import Path
-from typing import Any
+from typing import Any, AsyncGenerator
```
olaughter, line 11: This is interesting! I think that `Generator` should be coming from typing too 🤔

PR #645, opened 2025-08-27
Small changes to classifier eval

harrisonpim:
Couple of very small changes which should help to unblock/improve classifier development/evaluation for the policy team! Decided to bundle them together because they're all so small.

1. The vibe checker is still failing. The failure seems to be new, and related to the vibe checker not being a proper python module. Adding a couple of `__init__.py` files to the `static_sites` dir _should_ fix this. I feel like I've been trying to get the vibe checker working again for months, so i'm half-expecting a fresh round of bugs to start haunting me when this fix is merged 👹👻🐛
2. The vibe checker predictions should be shuffled before they're formatted as HTML ([see discussion in slack](https://climate-policy-radar.slack.com/archives/C02PM87FJEA/p1756289836690329)). I've added a single line to `generate_predictions_for_classifier` which should resolve this.
3. Argilla v2 will automatically count a passage as 'complete' as soon as a single labeller annotates it. The passage is then hidden it from all other annotators. Ideally, we want two labellers per passage, which is difficult/impossible to achieve in this new scheme ([see discussion in slack](https://climate-policy-radar.slack.com/archives/C078K2QNU0L/p1756218716196829)). adding the `TaskDistribution(min_submitted=2)` to the settings should fix this for future datasets 🤞


olaughter (approved):
Looks good! If this doesnt resolve and you'd like more brain on this then I have space tomorrow morning, let me know

harrisonpim:
thank you!

tbh i'm thinking about rewriting the vibe checker from scratch as a more modular, independent set of prefect tasks/flows/whatever. the current version feels very bloated, and so many of the recent issues could be avoided if it got some dedicated dev time, rather than just patchy bugfixes. 

when i've got something reasonable together, i'll prob call on you for another review 😉

PR #653, opened 2025-09-01
fix: timeout async tests to avoid flakiness

olaughter:
Some async tests had been hanging indefinitely. This ensures these tests time out. I added these to make the tests fail faster but they actually seem to resolve the issue. The hanging behaviour was odd in that the tests would just hang after passing successfully so perhaps this just helps with some cleanup or resolution after the tests finish...


tests/flows/conftest.py
```diff
@@ -101,7 +101,7 @@ async def mock_s3_async_client(
     with mock_aws():
         session = aioboto3.Session(region_name="eu-west-1")
         config = BotoCoreConfig(
-            read_timeout=60, connect_timeout=60, retries={"max_attempts": 3}
+            read_timeout=3, connect_timeout=3, retries={"max_attempts": 0}
```
michaelokarimia, line 104: Does `retries` in `Max attempts` being set to 0 mean it will never stop attempting to retry?

olaughter: All good, from [the documentation](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html): "Setting this value to 0 will result in no retries ever being attempted after the initial request"

olaughter:
I ended up cherry picking the commits onto https://github.com/climatepolicyradar/knowledge-graph/pull/611 so closing this one in favour of that one

PR #657, opened 2025-09-02
test: use a generator mock for run_deployment

olaughter:
Trying out a nicer model for mocking that acts more as a customisable mock builder then our hardcoded mocks we typically use. There are existing mocks for `run_deployment` in `test_infer` and `test_index`, if this turns out to be enjoyed here we could expand it there too.

This is an experiment that I've only applied to one test, so it does look a little overblown right now!

### DIFF
--- a/tests/flows/test_inference.py
+++ b/tests/flows/test_inference.py
@@ -51,6 +51,35 @@ from src.labelled_passage import LabelledPassage
 from src.span import Span
 
 
+@pytest.fixture
+def mock_deployment():
+    """A `run_deployment` mock wrapper that lets result state be customised."""
+
+    class MockDeployment:
+        def __init__(self, state, name="test-flow-run"):
+            """Mock run deployment, a state per call"""
+            self.state = state
+            self.name = name
+            self._mock_patch = None
+
+        async def mock_awaitable(self, *args, **kwargs):
+            """Generate FlowRun with next state from the iterator"""
+            flow_id = uuid.uuid4()
+            flow_name = f"{self.name}-{str(flow_id)[:8]}"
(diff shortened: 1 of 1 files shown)


jesse-c (approved):
For the kind of factory approach I spoke about, I'd have this as a function that returns the _thing_, instead of it being a fixture. If it's a fixture, I don't view too different to how we're currently doing it.

If it's a function, that means you can return as many as you like. Whereas, being a fixture, is that possible? An AI-driven search revealed that you should be able to call the fixture again, but that doesn't _feel_ right:

```
def test_example(my_fixture):
      copy1 = my_fixture
      copy2 = my_fixture()  # Call the fixture function again
```

It also gave an example of the factory pattern:

```
 Create a fixture that returns a factory function:
  @pytest.fixture
  def my_fixture_factory():
      def _create():
          return SomeObject()
      return _create

  def test_example(my_fixture_factory):
      copy1 = my_fixture_factory()
      copy2 = my_fixture_factory()
```

It seems like patching behaviour is orthogonal to the factory idea that I spoke about, so I'm not entirely sure overall! I was just thinking data, and not functions. Just don't think I'm strongly for or against, but it would reduce the space for the somewhat common behaviour being patched here.

olaughter:
Thanks for the detailed reply, some good food for thought on using more of a factory approach in the future 

> I'd have this as a function that returns the _thing_, instead of it being a fixture. 

That makes sense, I guess in this case multiple calls isn't necessary but I like the sound of taking this idea further

> It seems like patching behaviour is orthogonal to the factory idea that I spoke about, so I'm not entirely sure overall! I was just thinking data, and not functions.

That makes sense, this was part of what was needed in this case, but by adding functionality it does also tie to the context more than a pure data factory.

PR #658, opened 2025-09-02
refac: update aggregate for dont_run_on filtering

olaughter:
Because aggregate did expect every classifier to have outputs for every document.

Handle the special case that we filtered everything out, in practice I expect this to be unlikely because inference wouldnt have run for it either.


flows/aggregate.py
```diff
@@ -272,6 +284,10 @@ async def process_document(
                 ),
             )
             return document_stem
+    except AllSkippedFailure as e:
+        raise AggregationFailure(
+            document_stem=document_stem, exception=e, context=repr(e)
+        )
```
THOR300, line 290: **thought**: Is this definitely a failure, for example right now we're not running inference on Sabin docs so if we ran aggregation on a Sabin doc wouldn't aggregation be running as expected?

THOR300 (approved):
LGTM

olaughter:
> LGTM - staging test would be good as well.

Thanks for the nudge, I did run one last week but forgot to add it onto the pr description, I have added it back on now (spoiler it had passed successfully 😄)

PR #661, opened 2025-09-05
Make this project importable as `knowledge_graph`

harrisonpim:
There's a growing need/desire to import the code from this repo into other projects:
- the https://github.com/climatepolicyradar/policy-mentions-classifier used the `LLMClassifier` from this repo to produce some of its results
- TD folks have wanted to build on/with our work
- I want to set up a separate repo for a more serious vibe checker implementation, without bloating this one

At the moment, if someone runs `pip install git+https://github.com/climatepolicyradar/knowledge-graph`, they'll get a copy of the `src` directory and all of the main classes that we use for this work. It works! People can use our stuff! 🎉

However, because our `pyproject.toml` says `packages = ["src"]`, the exported module comes labelled as `src`. That means that in a separate project, users of our knowledge graph code still need to import the code from a module called `src`:

```py
from src.classifier import Classifier  # even though we're in a separate project, we're still importing from src??
```

If their project _also_ has a `src` module (as many do), things start to get ugly/confusing:

```py
from src.classifier import Classifier  # this "src" comes from the knowledge-graph repo
from src.my_cool_module import Something  # but this "src" is local to their project
```

Yuck! At this point, the code starts to break 😞

---

I think our knowledge graph code should probably be importable as `knowledge_graph` instead of `src`

```py
from knowledge_graph.wikibase import WikibaseSession
from knowledge_graph.concept import Concept
from knowledge_graph.classifier import Classifier
...
```

To achieve this:

- I've renamed our `src` directory to `knowledge_graph`
- I've updated the `pyproject.toml` to build a module called `knowledge_graph` from that `knowledge_graph` directory, which others should be able to install/import
- I've updated all of the other imports in eg `scripts`, `flows`, and `tests` which previously referred to `src`
- As noted in [a previous PR](https://github.com/climatepolicyradar/knowledge-graph/pull/632), changes to our dependencies can cause weird things to happen to our saved classifier artefacts. When running the tests, I saw failures on those which tried to load an existing classifier fixture with complaints about the `src.classifier` module not existing. Saving a new version of the same classifier resolved those issues.

```py
# save_a_new_Q788.py
from knowledge_graph.classifier import ClassifierFactory
from knowledge_graph.wikibase import WikibaseSession

wikibase = WikibaseSession()
concept = wikibase.get_concept("Q788")

classifier = ClassifierFactory.create(concept)

# the classifier's ID remains the same, as the concept data and classifier behaviour
# remain unchanged from the previous version. the path is therefore unchanged, and 
# the previous version of the file is overwritten
classifier.save(
    f"tests/flows/fixtures/classifiers/{concept.wikibase_id}/{classifier.id}/model.pickle"
)
```

closes SCI-488

### DIFF
--- a/deployments.py
+++ b/deployments.py
@@ -33,6 +33,7 @@ DEFAULT_FLOW_VARIABLES = {
     "cpu": MEGABYTES_PER_GIGABYTE * 4,
     "memory": MEGABYTES_PER_GIGABYTE * 16,
     "ephemeralStorage": {"sizeInGiB": 50},
+    "match_latest_revision_in_family": True,
 }
 
 
@@ -156,6 +157,7 @@ create_deployment(
     flow_variables={
         "cpu": MEGABYTES_PER_GIGABYTE * 16,
         "memory": MEGABYTES_PER_GIGABYTE * 64,
+        "match_latest_revision_in_family": True,
     },
     extra_tags=["type:sub"],
 )
--- a/knowledge_graph/classifier/bert_based.py
+++ b/knowledge_graph/classifier/bert_based.py
@@ -121,6 +121,7 @@ class BertBasedClassifier(Classifier, GPUBoundClassifier, UncertaintyMixin):
                 span = Span(
                     text=text,
                     concept_id=self.concept.wikibase_id,
+                    prediction_probability=prediction["score"],
                     start_index=0,
                     end_index=len(text),
                     labellers=[str(self)],
--- a/knowledge_graph/classifier/large_language_model.py
+++ b/knowledge_graph/classifier/large_language_model.py
@@ -1,5 +1,6 @@
 import asyncio
 import json
+import logging
 import random
 from abc import ABC, abstractmethod
 from typing import Annotated, Optional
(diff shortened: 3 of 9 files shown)


kdutia:
abstaining from a review because i'm curious about what happens with remote classifiers, but don't feel like i'm back up to speed enough yet to dig in!

harrisonpim:
> Is there a way to make a comparable change, with uv, so we don't need to move so many files?

Yeah, I hoped that would be the case too, but it looks like uv doesn't support the `include` `from` pattern 😞 

> Or, just retrain them all?

Based on the discussion in https://github.com/climatepolicyradar/knowledge-graph/pull/632, i'd assumed that the decision was to just recreate all classifiers in situations like this. It does feel a bit mad, but I think the import shim approach might end up being even madder? what do you reckon?

jesse-c:
> > Is there a way to make a comparable change, with uv, so we don't need to move so many files?
> 
> Yeah, I hoped that would be the case too, but it looks like uv doesn't support the `include` `from` pattern 😞

That's disappointing, from them.

> 
> > Or, just retrain them all?
> 
> Based on the discussion in #632, i'd assumed that the decision was to just recreate all classifiers in situations like this. It does feel a bit mad, but I think the import shim approach might end up being even madder? what do you reckon?

We just don't want to be retraining all models semi-frequently. I'd quickly checkin with @olaughter on if the new canonical IDs, in the v2 classifier specifications, would upset anything.

jesse-c (approved):

PR #663, opened 2025-09-08
Enabling the full-pipeline automation.

THOR300:
This Pull Request: 
---
- Enables the full-pipeline automation such that documents processed via the main document ingest in aws batch are run through the knowledge graph pipeline automatically. 
- This can now be done because targets have been re-enabled in production.

### DIFF
--- a/automations.py
+++ b/automations.py
@@ -213,7 +213,7 @@ async def main() -> None:
         b_deployment_name=generate_deployment_name(full_pipeline.name, aws_env),
         b_parameters={"inference_use_new_and_updated": True},
         description="Start the knowledge graph full pipeline.",
-        enabled=False,
+        enabled=True,
         aws_env=aws_env,
         expect_state=StateType.RUNNING,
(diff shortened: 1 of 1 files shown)


automations.py
```diff
@@ -213,7 +213,7 @@ async def main() -> None:
         b_deployment_name=generate_deployment_name(full_pipeline.name, aws_env),
         b_parameters={"inference_use_new_and_updated": True},
         description="Start the knowledge graph full pipeline.",
-        enabled=False,
+        enabled=True,
         aws_env=aws_env,
         expect_state=StateType.RUNNING,
         ignore=[AwsEnv.labs],
```
olaughter, line 219: **non-blocking question:** are we happy here that in sandbox when we run the integration tests they will also trigger the `full_pipeline`?

olaughter (approved):
🙌

automations.py
THOR300: Yeh I believe it's a good thing to do!

PR #671, opened 2025-09-09
Add a prefect flow to sync wikibase and neo4j

harrisonpim:
At the moment, refreshing neo4j with the latest data is an ad hoc job for the data-science team, or code-minded policy people ([see latest discussion](https://climate-policy-radar.slack.com/archives/C05DJCD4U3C/p1757346892228639)). I'd like to take that job off my plate, while giving the policy team peace of mind that the data they're looking at is always up to date.

This flow replaces the `populate_neo4j` script with a more robust flow. By default, it pulls the latest concept data from the concept store and updates the relevant parts of the graph in neo4j.

The flow also gives users the option to use a local copy of the aggregated inference results to update _everything_ in neo4j, including the `document-passage-concept` relationships. The flow's docstring includes instructions on how to get a copy of that data, and how to run the extra part of the flow.

I've also added a deployment which will run the default flow (ie concepts only) each night.


jesse-c (approved):
With the limited previous awareness of this, I feel like I've done a sufficient due diligence.

I worry about injection attacks via the Concept Store via some other source, in the Cypher queries. Presumably it's mature enough to handle those.

Before merging, could you do a test run in Prefect? CI deploys to Prefect, so it should be there already. Otherwise, you can also do a manual deploy to staging, and run it in there.

If I'm understanding correctly, you can currently never run a refresh in Prefect, since it doesn't download the data from S3 to the aggregation folder path.

PR #672, opened 2025-09-10
store prediction probabilities in model outputs

kdutia:
A small change for us to start storing prediction probabilities, as part of [SCI-509](https://linear.app/climate-policy-radar/issue/SCI-509/make-uncertainty-classifier-code-more-modular-and-instrumentable). Also updates the targets classifiers and `BertBasedClassifier` to store these.

It'd be nice if we could apply the threshold later on in the KG pipeline so as to keep all predictions, but that can come later as-and-when we decide we _need_ it. For now i've made the minimum amount of change to enable storing prediction probabilities

**Open questions:**

- Do we want to store the labeller that made the probability prediction, or assert that a `Span` can have only one labeller if it has a `prediction_probability`? I'm not familiar enough with this code / intentions around it yet to figure out whether this would be a good or a bad idea...

### DIFF
--- a/src/classifier/bert_based.py
+++ b/src/classifier/bert_based.py
@@ -121,6 +121,7 @@ class BertBasedClassifier(Classifier, GPUBoundClassifier, UncertaintyMixin):
                 span = Span(
                     text=text,
                     concept_id=self.concept.wikibase_id,
+                    prediction_probability=prediction["score"],
                     start_index=0,
                     end_index=len(text),
                     labellers=[str(self)],


jesse-c (approved):
Cool to see! Is there space for a unit test?

> It'd be nice if we could apply the threshold later on in the KG pipeline so as to keep all predictions, but that can come later as-and-when we decide we need it.

Is there another possible feature of getting the values into Vespa, and including them on document passages? I'm currently revisiting the Vespa data models, so could be good for me just to check if in practice the changes could enable this.

jesse-c:
> Cool to see!
> 
> > It'd be nice if we could apply the threshold later on in the KG pipeline so as to keep all predictions, but that can come later as-and-when we decide we need it.
> 
> Is there another possible feature of getting the values into Vespa, and including them on document passages? I'm currently revisiting the Vespa data models, so could be good for me just to check if in practice the changes could enable this.

Easy peasy. It could be on https://github.com/climatepolicyradar/navigator-infra/blob/feat%2Fadd-model-hot-swapping-schemas/vespa/navigator_app_dev/schemas/document_passage.sd#L93-L100, with a new field to match this field, sometime in the future.

Since reindexing isn't fun, and we're going to be running targets _soon_, maybe we do want to get the data in, but not surface it to anyone?

jesse-c:
> Yes – that would fab! I'm not sure how useful the data will be in Vespa (it depends on whether we can filter on it I guess), but no harm in putting it in. Could be useful for eg raising thresholds if we find they're too low.

Sounds good!

> Good shout – looks like there's room for checking that the targets classifier outputs spans with probabilities. I was beginning to think about a way we could check for this at runtime through something like a mixin but didn't get much further with those thoughts...

Am I mistaken from my naïve understanding here that these changes are to the targets classifiers, that would do this? 😅

src/span.py
```diff
@@ -39,6 +39,13 @@ class Span(BaseModel):
     end_index: int = Field(
         ..., gt=0, description="The end index of the span within the text"
     )
+
+    prediction_probability: Optional[float] = Field(
+        default=None,
```
harrisonpim, line 44: `None` or `1`?

harrisonpim (approved):
looks good!

on indexing extra stuff into vespa, i'm a little bit wary - we won't have scores for anything below the threshold, and it feels weird to only be able to raise the threshold and not lower it...

src/span.py
kdutia: I changed to `None` after our chat as I figured it'd be good to be able to identify whether a model has tried to set confidence or not. this way we can explicitly disambiguate between predictions with probabilities and without

jesse-c:
> @jesse-c sorry, I mistyped! I meant for _any_ classifier that we expect to output probabilities.
I see!
> I also realised that we mock the spans output in the targets unit tests (presumably for either speed or downloading from huggingface reasons), so I won't be able to add that check in for now regardless
Fair enough.
> on indexing extra stuff into vespa, i'm a little bit wary - we won't have scores for anything below the threshold, and it feels weird to only be able to raise the threshold and not lower it...
No worries. From my simple PoV, I wouldn't be thinking about restricting it based on if it's higher or lower, just if the model is about to give a normalised probability or not, and then just stick it in.

Features coming to mind being for searching, with a minimum confidence (à la temperature when using an LLM), or displaying the probability (à la in Queried).

Just thinking out loud, so no commitments or one-way doors here.

src/span.py
harrisonpim: fair!

PR #676, opened 2025-09-13
Search for concepts and help pages through a WikibaseSession

harrisonpim:
Context: Last week, I moved our concept store documentation into wikibase as a series of Help pages, eg [Help:Hierarchy_heuristics](https://climatepolicyradar.wikibase.cloud/wiki/Help:Hierarchy_heuristics)

This PR gives us the ability to:
- search for those pages
- retrieve the content of the pages as markdown
- also search for concepts

These methods alone aren't super interesting... But it's very easy to build an MCP server around a class like this these days, and I think that's where they _could_ start becoming useful 👀

If, at some point in the future, we want to give LLM agents direct access to our concept store data, I reckon they'll benefit _a lot_ from having access to all of the good thinking in eg [our style guide](https://climatepolicyradar.wikibase.cloud/wiki/Help:Style_guide)... Maybe we could set up a more advanced [librarian](https://djhoyps9k9bag.cloudfront.net/) which periodically suggests edits to concepts based on the latest heuristics etc. All speculation, obviously, but I think there are some interesting possibilities here which are enabled by these small additions!

### DIFF
--- a/knowledge_graph/wikibase.py
+++ b/knowledge_graph/wikibase.py
@@ -88,9 +88,9 @@ class WikibaseSession:
     MAX_PAGE_REQUESTS = 2000  # Suitable up to 1M pages (500*2000)
     MAX_CONCURRENT_REQUESTS = 3  # Limit concurrent requests to avoid rate limiting
     REQUEST_DELAY_SECONDS = 0.5  # Small delay between requests to be gentle on server
+    HELP_NAMESPACE = 12
     ITEM_NAMESPACE = 120
     ITEM_PREFIX = "Item:"
-    HELP_NAMESPACE = 12
 
     # Property IDs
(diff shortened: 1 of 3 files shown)


knowledge_graph/wikibase.py
```diff
@@ -946,3 +948,96 @@ async def add_alternative_labels(
...
+    @async_to_sync
+    async def get_help_page(self, page_title: str) -> str:
+        """Sync wrapper for get_help_page_async"""
+        return await self.get_help_page_async(page_title)
+
+    async def search_concepts_async(
+        self,
+        search_term: str,
+        limit: Optional[int] = None,
+        timestamp: Optional[datetime] = None,
```
kdutia, line 1002: could you add descriptions of what these parameters mean? `timestamp` is the one that's unclear to me - but think it's worth adding to all given this is in a shared/production-y bit of code!

PR #678, opened 2025-09-15
fix 'event loop is closed' error in wikibase async_to_sync method

kdutia:
Trying to run this code failed with the error `RuntimeError("Event loop is closed")`

``` py
CONCEPT_IDS = [
    "Q1285", # ban
    "Q913", # restorative justice
    "Q1016", # sexual minority
    "Q760", # extractive sector
    "Q715", # tax
]

wikibase = WikibaseSession()

for concept_id in CONCEPT_IDS:
    concept = wikibase.get_concept(WikibaseID(concept_id))
```

It seems what was happening was each call to `get_concept()` created a fresh event loop and then closed it, but `_client: httpx.AsyncClient` was bound to the first loop – thus the `RuntimeError`.

With Claude's help I've updated the code to use one persistent loop per thread – this seems like a better alternative to creating a new loop (thus httpx client) every time we want to make a call to Wikibase.

This also raises the importance of having some tests against the concept store I think!

### DIFF
--- a/knowledge_graph/wikibase.py
+++ b/knowledge_graph/wikibase.py
@@ -1,7 +1,9 @@
 import asyncio
+import atexit
 import functools
 import json
 import os
+import threading
 from datetime import datetime, timezone
 from logging import getLogger
 from typing import Any, Callable, Coroutine, Optional, TypeVar, cast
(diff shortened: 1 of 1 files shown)


harrisonpim (approved):
I've seen the same error a couple of times over the last few weeks, and had messed around with some similar solutions! Thanks for pushing this :)

PR #683, opened 2025-09-15
Add new classifiers to the specs, and update the existing ones

harrisonpim:
I've recreated all of the existing classifiers following #661, and added the recently signed off IDs from fossil fuels, renewables, and just transition.

the specs for staging and prod envs have been auto-updated, and i've manually added metadata so that:
- none of the models run on sabin
- targets models are set to run on GPUs


.pre-commit-config.yaml
```diff
@@ -14,10 +14,10 @@ repos:
...
-        types_or: [ python, pyi, jupyter ]
-        args: [ --fix ]
+        types_or: [python, pyi, jupyter]
+        args: [--fix]
       - id: ruff-format
-        types_or: [ python, pyi, jupyter ]
+        types_or: [python, pyi, jupyter]
   - repo: local
     hooks:
       - id: pyright
```
olaughter, line 23: Line 27 here still has a `src` reference that would be worth catching

flows/classifier_specs/staging.yaml
```diff
@@ -1,77 +1,89 @@
...
+- Q1362:v13
+- Q1368:v13
+- Q1369:v13
+- Q1370:v13
+- Q1371:v13
+- Q1651:v13
+- Q1652:v10
+- Q1653:v10
+- Q1744:v1
+- Q1754:v1
```
olaughter, line 89: The concepts-api relies on these at the moment meaning after this goes in and at the point the api is next deployed they will start appearing in search.

olaughter (commented):
Thanks Harrison! I'm guessing this will resolve [the issue ](https://climate-policy-radar.slack.com/archives/C07695T6J75/p1757932847954739)I just began looking into in prefect and is blocking regular runs right now:

```
  File "/app/flows/inference.py", line 312, in load_classifier
    classifier = Classifier.load(model_path)
  File "/app/knowledge_graph/classifier/classifier.py", line 166, in load
    classifier = pickle.load(f)
ModuleNotFoundError: No module named 'src'           
```

The fact that it is module rather then file is significant and makes me think it was to do with the classifiers being pickled from a repo with src, but now unpickled in one without. whereas if the file path itself was wrong it would be a FileNotFound.

I am a bit cautious of adding new classifiers all in one. Are these ready to go live on the site? @jesse-c and I have talked about how we want to roll these out — ideally with hotswapping but falling back to a more phased way assuming that hot swapping won't be available in time for NYCAW. It might be worth making two PRs out of this — one updating the existing ones to resolve the unpickling issue, another with the additions that we want tot trerat in their own right

jesse-c:
Agreed with splitting in two, it'll make it just slightly easier for us in tracking this, and getting the full pipeline run. I let @olaughter know too that we're fully up-to-date on this, and thank you for checking!

olaughter:
Thanks both! Exciting times!

jesse-c (approved):
LGTM

There are some removals, e.g. `Q639`, and am assuming that's fine!

harrisonpim:
i believe that gas (Q650) and oil (Q639) were significantly updated as part of this round of concept development, so they should end up back in the list as part of the follow-up PR #684

PR #694, opened 2025-09-17
Add tests for train flow and script

olaughter:
Adds some initial test coverage to both the flow and the script


tests/conftest.py
```diff
@@ -22,6 +23,8 @@ def aws_credentials():
     os.environ["AWS_SECRET_ACCESS_KEY"] = "testing"
     os.environ["AWS_SECURITY_TOKEN"] = "testing"
     os.environ["AWS_SESSION_TOKEN"] = "testing"
+    with mock_aws():
```
jesse-c, line 26: **question:** I think this may end up with nested `@mock_aws` calls for the tests that use that manually, and also use this fixture?

**quibble:** This fixture isn't being used anywhere any more? Odd. I think the name of the fixture is now a little misleading, as it does 2 things now.

jesse-c (approved):

PR #697, opened 2025-09-17
feat: Add artifact for full_pipeline, fix test_index artifact test

kaparker:
Adding simple prefect artifact for `full_pipeline` flow.
Tests added to check artifact exists for successful full pipeline runs.

Note: no artifact will be created if any of the flows called by full pipeline fail.

Additional fix to index test where testing artifact creation:
+ De-indent after `with pytest.raises(ValueError) as excinfo` as remaining tests were not being executed
+ Update ValueError message to be 3/3 documents (instead of 1/3)
+ Update artifact test to get artifact based on key

### DIFF
--- a/flows/aggregate.py
+++ b/flows/aggregate.py
@@ -195,9 +195,9 @@ def generate_s3_uri_output(
 
 
 def task_run_name(parameters: dict[str, Any]) -> str:
-    document_id = parameters.get("document_id", "unknown")
+    document_stem = parameters.get("document_stem", "unknown")
     slug = generate_slug(2)
-    return f"aggregate-single-document-{document_id}-{slug}"
+    return f"aggregate-single-document-{document_stem}-{slug}"
 
 
 @materialize(


tests/flows/test_index.py
```diff
@@ -437,18 +437,20 @@ async def test_run_indexing_from_aggregate_results__handles_failures(
                 batch_size=1,
             )
 
-            assert (
-                f"Some batches of documents had failures: 1/{len(document_stems)} failed."
-                in str(excinfo.value)
-            )
+        assert (
+            f"Some batches of documents had failures: 3/{len(document_stems)} failed."
```
kaparker, line 441: This previously was not being run as it was within the context manager.
Is it expected that 3 documents fail as part of this test? Previously it was asserting that 1 fail (assuming the "non-existent" file introduced in this test)

kaparker: Expected for all documents to fail as in all other tests running `index()` due to limitations with mocking in the test environment. The failures returned are `ObjectNotFound(None)` for all objects. 

The purpose of the test is to check the error handling and that the flow continues when value error is raised, this is tested by checking the error message and that the artifact is created after the value error is raised.

tests/flows/test_index.py
```diff
@@ -437,18 +437,21 @@ async def test_run_indexing_from_aggregate_results__handles_failures(
                 batch_size=1,
             )
 
-            assert (
-                f"Some batches of documents had failures: 1/{len(document_stems)} failed."
-                in str(excinfo.value)
-            )
+        # Expect all documents to fail in test environment due to mocking limitations
```
olaughter, line 440: 👍

tests/flows/test_index.py
```diff
@@ -437,18 +437,21 @@ async def test_run_indexing_from_aggregate_results__handles_failures(
...
-            )
+        # Assert that the summary artifact was created after flow completion
+        summary_artifact = await Artifact.get(
+            "indexing-aggregate-results-summary-sandbox"
+        )
+        assert summary_artifact and summary_artifact.description
+        assert (
+            summary_artifact.description
+            == "Summary of the passages indexing run to update concept counts."
+        )
```
olaughter, line 454: praise: Nice work taking the initiative and fixing this! It's pretty subtle that the indentation was masking it from running!

olaughter (approved):
Looks good!

PR #699, opened 2025-09-17
feat: include gpu metadata at train time

olaughter:
This way the classifier_metadata script doesnt need to be rerun every time.


kdutia (approved):
lgtm! 

question (that I feel like has probably been answered somewhere already) – do we want to plan _now_ to be able to handle multiple types of GPU? the one for BERT-based models would be a lot smaller for any open source generative models. although I imagine we might decide not to self-host the latter any time soon...

PR #705, opened 2025-09-18
Fix Span.from_xml when there are no annotations

kdutia:
Small fix for `Span.from_xml` (used by LLM classifiers which return `<concept>` tags) for when no concepts are annotated in the text.

### DIFF
--- a/tests/test_span.py
+++ b/tests/test_span.py
@@ -413,3 +413,16 @@ def test_span_from_xml_invalid_concept_annotation(xml: str, is_valid: bool):
     else:
         with pytest.raises(SpanXMLConceptFormattingError):
             _ = Span.from_xml(xml, concept_id=WikibaseID("Q123"), labellers=["me"])
+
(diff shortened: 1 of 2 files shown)


tests/test_span.py
```diff
@@ -413,3 +413,16 @@ def test_span_from_xml_invalid_concept_annotation(xml: str, is_valid: bool):
...
+
+@pytest.mark.parametrize(
+    "xml",
+    [
+        "",
+        "no concept tags here",
+        "Some <b>bold</b> but no concept tags.",
+    ],
+)
+def test_span_from_xml_empty(xml: str):
```
harrisonpim, line 426:
```suggestion
def test_whether_parsing_text_with_no_concept_tags_returns_an_empty_list(xml: str):
```

really minor thing, but i try to give my tests really long, descriptive names - i find it makes debugging much easier when tests start failing!

PR #706, opened 2025-09-18
add classifier ensemble and ProbabilityCapableClassifier; remove UncertaintyMixin

kdutia:
After a good chat with @harrisonpim yesterday we decided that there's a careful balance somewhere between the `UncertaintyMixin` and the `VotingClassifier` I defined in #674 (now closed).

This PR implements the first stages of that:

- a `ProbabilityCapableClassifier` mixin, so we know which classifiers give us probabilities
- an `Ensemble` class, which is a collection of classifiers for the same concept
- replacing `UncertaintyMixin` with a simpler `VariantEnabledClassifier` protocol

The next PR as part of SCI-509 will be to add a set of ensemble measures, including uncertainty, for the outputs of the ensemble class.

### DIFF
--- a/knowledge_graph/span.py
+++ b/knowledge_graph/span.py
@@ -217,6 +217,9 @@ class Span(BaseModel):
 
         tags = re.findall(r"</?concept>", xml)
 
+        if not tags:
+            return
+
         if not (
             tags[0] == "<concept>"
             and len(set(tags)) == 2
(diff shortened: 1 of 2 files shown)


tests/test_classifier_ensemble.py
```diff
@@ -0,0 +1,51 @@
...
+)
+from knowledge_graph.classifier.keyword import KeywordClassifier
+from knowledge_graph.classifier.stemmed_keyword import StemmedKeywordClassifier
+from knowledge_graph.concept import Concept
+from knowledge_graph.identifiers import ClassifierID, WikibaseID
+from tests.common_strategies import concept_strategy
+
+
+@pytest.mark.xdist_group(name="classifier")
+def test_whether_ensemble_rejects_classifiers_with_different_concepts():
```
harrisonpim, line 16: looove the long test names here btw ;)

PR #713, opened 2025-09-18
ci: don't redeploy sandbox

olaughter:
Currently on merge to main, we deploy sandbox, then we redeploy sandbox before deploying everything else. Deploys can take over 5 minutes each so this is pretty painful. Its still a good check that deploys are valid before trying to deploy to other envs so Im just taking out the last deploy to sandbox

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -11,23 +11,11 @@ permissions:
   contents: write
 
 jobs:
-  deploy_prefect_sandbox:
+  deploy_prefect_labs:
     if: |
       (github.event.workflow_run.conclusion == 'success' &&
       github.ref == 'refs/heads/main') ||
       github.event_name == 'workflow_dispatch'
-    uses: ./.github/workflows/prefect_deploy.yml
-    with:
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -50,7 +40,10 @@ jobs:
       PREFECT_API_URL: ${{ secrets.PREFECT_API_URL }}
 
   deploy_prefect_prod:
-    if: github.ref == 'refs/heads/main'
+    if: |
+      (github.event.workflow_run.conclusion == 'success' &&
+      github.ref == 'refs/heads/main') ||
+      github.event_name == 'workflow_dispatch'
     needs: [deploy_prefect_sandbox]
```
kaparker, line 47: Does this dependency need removing too?

jesse-c (approved):
Seems reasonable to rely on the sandbox deployment from CI, and then to start all 3 CD deployments concurrently.

.github/workflows/cd.yml
olaughter: Woops yes, thank you!

PR #714, opened 2025-09-18
Raise clearer errors on fetching invalid concepts from wikibase

harrisonpim:
@jesse-c spotted an issue where `WikibaseSession` was incorrectly throwing a `ConceptNotFoundError` for Wikibase concepts that, while technically found by their ID (e.g., via `get_all_concept_ids_async`), were actually invalid due to missing preferred labels. This led to an inconsistent state where `get_concept_async` would then erroneously report these existing-but-invalid concepts as not found. [See thread](https://climate-policy-radar.slack.com/archives/C078K2QNU0L/p1758204287359459)

This PR adds an `InvalidConceptError` to our list of exceptions, and raises it when a fetched concept contains invalid data. Hopefully this will make future debugging easier if this happens again!

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -11,23 +11,11 @@ permissions:
   contents: write
 
 jobs:
-  deploy_prefect_sandbox:
+  deploy_prefect_labs:
     if: |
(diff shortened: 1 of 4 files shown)


jesse-c (approved):
Thank you!

PR #716, opened 2025-09-22
chore: once again temporarily disable targets

olaughter:
To get new classsifiers live quickly we then plan to follow up with targets afterwards. For reference:

https://climate-policy-radar.slack.com/archives/C078K2QNU0L/p1758289642269079

berne-targets


olaughter:
> Was this done in manually? If not, could we use the demote script?

These where done with the demote script

PR #717, opened 2025-09-22
Add semaphore for run_classifier_inference async task

kaparker:
Try to address the `RequestTimeTooSkewed` error [observed in inference runs](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/068cc263-70f1-7381-8000-b77741cad11b?entity_id=068cc263-70f1-7381-8000-b77741cad11b&state=cancelled&state=cancelling&state=crashed&state=failed&entity=flowRuns&entity=taskRuns&entity=artifacts&log-level=40) by limiting the number of async requests made to 200.

Set semaphore sessions to 200 as [s3 allows 5500 requests per second](https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html) and we are running 1000 docs per batch and around ~20 batches for all documents in the full pipeline.

### DIFF
--- a/flows/inference.py
+++ b/flows/inference.py
@@ -474,7 +474,7 @@ async def store_labels(
 ]:
     """Store the labels in the cache bucket."""
     # Don't get rate-limited by AWS
-    semaphore = asyncio.Semaphore(10)
+    semaphore = asyncio.Semaphore(200)
 
     tasks = [
         wait_for_semaphore(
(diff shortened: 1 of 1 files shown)


THOR300 (approved):
LGTM! 

Ran the maths as well: 

5,500 [ head / get requests ] / 20 [ concurrent jobs ] = 275 

So 200 looks like a safe estimate. 

3,500 [ head / get requests ] / 20 [ concurrent jobs ] = 175 

So 100 looks like a safe estimate.

jesse-c (commented):
Do you know why it shows up as that particular error?

olaughter (approved):
lgtm!

PR #718, opened 2025-09-22
quick improvements to LLMCLassifier prompt

kdutia:
based on @AnneIsARealProgrammerNow's work. I also amended section 6 with instructions to avoid wrapping both the whole passage and individual mentions with `<concept>` tags as I was seeing this sometimes

### DIFF
--- a/knowledge_graph/classifier/large_language_model.py
+++ b/knowledge_graph/classifier/large_language_model.py
@@ -38,9 +38,9 @@ class LLMResponse(BaseModel):
 
 
 DEFAULT_SYSTEM_PROMPT = """
-You are a specialist climate policy analyst, tasked with identifying mentions of 
-concepts in climate policy documents. You will mark up references to concepts with 
-XML tags.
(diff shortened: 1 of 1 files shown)


harrisonpim:
**silly tangent:** This made me realize we're kind of flying blind on performance impacts of changes to some of our classifiers. I trust that this change is a good one, but it would be fab if we could _prove_ that it's an imrpovement across all important cases. Some of our classifier classes (eg the `KeywordClassifier`) have a sufficiently simple mechanism/behaviour to be testable via standard pytest stuff, but the LLM- or BERT-based models are much harder to evaluate that way.

I wonder whether there's a future in which we run a set of classifier evals in CI on changes to particular files, and surface the resulting metrics diff in the PR? Obviously a long way off, but it feels somewhat related to our search eval conversations too 🤔

kdutia:
maybe it's not CI, as much as me providing details of the experiment i ran as a data scientist. if the experiment was a success then maybe i'd also promote the new version of the classifier

PR #722, opened 2025-09-22
fix: Add more debug info

jesse-c:
Seeing these, only in ECS, makes it extremely hard to track down where they came from. I haven't successfully been able to do so.

The print statements are in a Prefect task, that doesn't have `log_prints` set to `True`. That task is being called from a flow that _does_ have that set, but I guess it doesn't carry down to tasks.


jesse-c:

- #721
    - #722 ◀

PR #723, opened 2025-09-22
Fix unclosed connector issue

kaparker:
Update get_bucket_paginator to take ses…sion as parameter and move async loop inside session in list_bucket_file_stems

Testing the first few steps of `inference` by [running locally from prefect](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/deployments/deployment/d2322948-2813-44e7-9cf9-9917ca40d96e?tab=Runs), the unclosed client and connector errors appear during the `list_bucket_file_stems` function.

```16:14:05.759 | ERROR   | asyncio - Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x168775160>
16:14:05.759 | ERROR   | asyncio - Unclosed connector
connections: ['deque([(<aiohttp.client_proto.ResponseHandler object at 0x168765a30>, 294856.717941833)])']
```

This occured at the end of the function when returning the `file_stems` suggesting [that the async session had not been closed](https://stackoverflow.com/questions/46112848/python-package-aiohttp-has-a-warning-message-unclosed-client-session).

By moving the session to the `list_bucket_file_stems` function allows looping through the pages returned from the s3 paginator to be within the same session. When the `async for ` loop is outside of the session context, the unclosed client/session errors persist.

### DIFF
--- a/flows/index.py
+++ b/flows/index.py
@@ -11,9 +11,8 @@ from typing import Any, Final
 import aioboto3
 import httpx
 from cpr_sdk.models.search import Passage as VespaPassage
-from prefect import flow, unmapped
+from prefect import flow, task, unmapped
 from prefect.artifacts import create_markdown_artifact, create_table_artifact
-from prefect.assets import materialize
 from prefect.client.schemas.objects import FlowRun
 from prefect.futures import PrefectFuture, PrefectFutureList
 from prefect.logging import get_run_logger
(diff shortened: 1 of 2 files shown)


jesse-c (approved):
LGTM. 

Thank you for the good PR description. We haven't changed the test coverage in this PR, but I think that's okay, considering the existing coverage of this code.

From your PR description, I'm assuming you confirmed the error disappeared in a flow run on Prefect, since you did it to reproduce the error!

kaparker:
Tested inference run in staging and the error no longer appears in the inference run parent task but it is still within the batch tasks. Looking into this and will push a change and retest in staging

PR #724, opened 2025-09-22
fix: add logging for `RequestTimeTooSkewed` error

olaughter:
Adds some initial, extra logging for these errors, for starters it will tell us how skewed the requests are getting.

Also includes a driveby fix for a typo in the train script — ci feels too painful right now to have a whole run just for it.

### DIFF
--- a/flows/inference.py
+++ b/flows/inference.py
@@ -2,13 +2,14 @@ import asyncio
 import json
 import os
 from collections.abc import Generator, Sequence
-from datetime import timedelta
+from datetime import datetime, timedelta
 from io import BytesIO
 from pathlib import Path
 from typing import Any, Final, NamedTuple, Optional, TypeAlias
 
 import aioboto3
 import wandb
+from botocore.exceptions import ClientError
 from cpr_sdk.parser_models import BaseParserOutput, BlockType
 from more_itertools import flatten
 from mypy_boto3_s3.type_defs import (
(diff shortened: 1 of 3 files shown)


olaughter:
> I think ideally we'd have some test coverage but understand we are in fire fighting mode.

No its a good point and wont take me long realistically. brb

flows/inference.py
```diff
@@ -328,13 +330,42 @@ async def load_classifier(
...
-        Key=key,
-    )
+    try:
+        response = await s3_client.get_object(
+            Bucket=config.cache_bucket,  # pyright: ignore[reportArgumentType]
+            Key=key,
+        )
+    except ClientError as e:
+        log_client_error(e)
+        raise
```
jesse-c, line 368: **thought:** I don't think we should log messages that are in exceptions. 


Following that, I think here instead we should make the exception have a readable message, like in your log message. I know this is a little different here since you aren't outputting the exact same message, but, think we could use the pattern of wrapping errors in more context, at the handler(s) levels.

We have some other places where we do `print .. raise ...`.

jesse-c (approved):
LGTM

Non-blocking thought on how we approach exceptions.

flows/inference.py
jesse-c: That's fair enough. Maybe that pattern isn't as bad as I think it is.

PR #727, opened 2025-09-22
fix: Include timestamps when merging spans

jesse-c:
I investigated one of the sources of the missing timestamps:
`labelled_passages/Q769/7cgtdjf8/CCLW.document.i00001166.n0000.json`.

~69% are missing:

```
rg --fixed-strings 'timestamps":[]' CCLW.document.i00001166.n0000.json | wc -l
     102
```

```
rg --fixed-strings 'timestamps":["' CCLW.document.i00001166.n0000.json | wc -l
      47
```

The issue is that the keyword-based classifier can merge spans. It wasn't
included the timestamps though when doing so.


knowledge_graph/span.py
```diff
@@ -183,14 +195,24 @@ def intersection(cls, spans: list["Span"]) -> "Span":
...
             return spans[0]
         else:
+            # Combine unique labellers and their timestamps
+            labeller_to_timestamp: OrderedDict[str, datetime] = OrderedDict()
+
+            # Assume labellers and timestamps are ordered to match
+            # each other already.
+            for span in spans:
+                for i, labeller in enumerate(span.labellers):
+                    # Prefer the first seen timestamp for each labeller
```
olaughter, line 205: thought: I wonder if we want to make this behaviour more explicit by adding a note about it in the docstring

olaughter (approved):
Well done, very good catch

PR #741, opened 2025-09-23
train script can run arbitrary model types

kdutia:
Custom classifiers can be specified using the train script. this lets us bypass the `ClassifierFactory` for experimental new model types that we don't want to use in production yet

```
uv run train --wikibase-id Q1285 --classifier-type LLMClassifier --classifier-kwarg "random_seed=101" \
 --classifier-kwarg "model_name=gpt-4.1-mini" --track-and-upload
```

Any classifiers that don't use the classifier factory are marked with `experimental-model-type` in their weights and biases config.

Also:

- fixes `RuntimeError: "This event loop is already running` when using `LLMClassifier` inside the train script

## wandb screenshot

first LLMClassifier in wandb woo

image

if we want, it's quite easy to now extend wandb to get e.g. [confusion matrices](https://docs.wandb.ai/guides/track/log/plots/#model-evaluation-charts). judging by [their documentation](https://docs.wandb.ai/guides/tables/) it can be useful to log a `wandb.Table` of the predictions on the eval set, so you can compare judgements of each model on individual items from the eval set.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,7 +85,6 @@ jobs:
 
   deploy_prefect_sandbox:
     name: "Deploy"
-    needs: [lint, test]
     uses: ./.github/workflows/prefect_deploy.yml
     with:
       aws-env: sandbox
--- a/deployments.py
+++ b/deployments.py
@@ -225,9 +225,18 @@ create_deployment(
                 ],
                 "classifier_specs": [
                     # CPU-based
-                    {"name": "Q708", "alias": "v6"},
+                    {
+                        "wikibase_id": "Q708",
+                        "classifier_id": "x9kfsd8s",
+                        "wandb_registry_version": "v14",
+                    },
                     # GPU-based
-                    {"name": "Q1651", "alias": "v1"},
+                    {
+                        "wikibase_id": "Q1651",
+                        "classifier_id": "6rys3abe",
+                        "wandb_registry_version": "v13",
+                        "compute_environment": {"gpu": True},
+                    },
                 ],
             }
         ),
--- a/flows/aggregate.py
+++ b/flows/aggregate.py
@@ -406,7 +406,6 @@ async def collect_stems_by_specs(config: Config) -> list[DocumentStem]:
 
 @flow(  # pyright: ignore[reportCallIssue]
     timeout_seconds=None,
-    log_prints=True,
     task_runner=ThreadPoolTaskRunner(
         # This is valid, as per the docs[1].
         #
@@ -421,11 +420,16 @@ async def aggregate_batch_of_documents(
     run_output_identifier: RunOutputIdentifier,
 ) -> RunOutputIdentifier:
     """Aggregate the inference results for the given document ids."""
+    logger = get_logger()
+
+    logger.info(
+        f"Aggregating {len(classifier_specs)} classifiers on {len(document_stems)} documents"
+    )
+
     config = Config.model_validate(config_json)
 
     tasks: list[PrefectFuture[DocumentStem]] = []
 
-    print("submitting tasks")
     for document_stem in document_stems:
         tasks.append(
             process_document.submit(  # pyright: ignore[reportFunctionMemberAccess, reportArgumentType]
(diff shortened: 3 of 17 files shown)


knowledge_graph/classifier/large_language_model.py
```diff
@@ -164,6 +165,17 @@ def __setstate__(self, state):
...
+
+        # Check whether an event loop is already running. Python can't run nested
+        # async processes, so if there's a running loop then we use `nest_asyncio` to
+        # make the prediction be able to run.
+        try:
+            loop = asyncio.get_running_loop()
+            if loop.is_running():
+                nest_asyncio.apply()
+        except RuntimeError:
+            pass
```
harrisonpim, line 177: sticking an escape hatch here smells suboptimal to me...

i know that this isn't the main point of the PR, and i'd guess that it's really just included to unblock the rest, but i think it's worth figuring out why these issues are happening. for what it's worth, i considered a get-out like this when i spotted similar issues around the wikibasesession in prefect, which makes me thing this is a larger problem which we should solve more generally.

maybe there's a common decorator which we could add to methods like this? it could sit alongside the `async_to_sync` stuff in `wikibase.py`? or maybe in a new `knowledge_graph/async.py`?

scripts/train.py
```diff
@@ -316,24 +394,41 @@ async def run_training(
...
+    )
+
+    if classifier_type:
+        classifier = create_classifier(
+            concept=concept,
+            classifier_type=classifier_type,
+            classifier_kwargs=classifier_kwargs or {},
+        )
+    else:
+        classifier = ClassifierFactory.create(concept)
```
harrisonpim, line 413: idea: would it be neater if we moved the class override _into_ the `ClassifierFactory` itself? ie rather than having this if/else in the script, we could just say:

```py
classifier = ClassifierFactory.create(
    concept,
    override_class=classifier_type  # either an actual value specified by the user, or None
    override_kwargs=classifier_kwargs  # as above
)
```

i feel like the kwarg and class name matching functionality are particularly well suited to living in the `knowledge_graph/classifier/__init__.py` compared to the single purpose script

harrisonpim (commented):
nice! really happy to see this feature materialise

i've added a couple of points to chat about re structure, but overall this is looking great :)

scripts/train.py
```diff
@@ -40,11 +40,37 @@
 app = typer.Typer()
 
 
-def validate_params(track: bool, aws_env: AwsEnv) -> None:
+def parse_classifier_kwargs(classifier_kwarg: Optional[list[str]]) -> dict[str, Any]:
```
harrisonpim, line 43: i was pondering this parser and the overall developer experience when looking at the command in the pr description:

```
uv run train --wikibase-id Q1285 --classifier-type LLMClassifier --classifier-kwarg "random_seed=101" --classifier-kwarg "model_name=gpt-4.1-mini" --track-and-upload
```

do you think it's prefereable (or even possible?) to instead do something like:

```
uv run train --wikibase-id Q1285 --classifier-type LLMClassifier --random-seed=101 --model-name=gpt-4.1-mini --track-and-upload
```

what do you reckon? worth the effort for the clarity of intention and nicer dev x? not worth it because of ambiguous edge cases, risks around misspellings, difficulty parsing?

scripts/train.py
```diff
@@ -316,24 +368,38 @@ async def run_training(
...
+        wikibase_id=wikibase_id,
+        include_labels_from_subconcepts=True,
+        include_recursive_has_subconcept=True,
+        wikibase_config=wikibase_config,
+    )
+
+    classifier = ClassifierFactory.create(
+        concept=concept,
+        classifier_type=classifier_type,
+        classifier_kwargs=classifier_kwargs or {},
```
harrisonpim, line 383:
```suggestion
        classifier_kwargs=classifier_kwargs,
```

i think the `or {}` might be redundant here, based on the outputs of `parse_classifier_kwargs`?

harrisonpim (approved):

scripts/train.py
kdutia: i thought about this at first too, but from what i could see from the typer docs + claude code it's not possible

PR #748, opened 2025-09-24
Finance flow at start of labelling

AnneIsARealProgrammerNow:
Sharing the version of the notebook with some early experiments all the way down to the set-up we used to create labels which we are currently evaluating in Argilla.


kdutia:
I suggested Anne stick it here as it was more for code sharing than anything we want to use – closing this PR and adding it to a linear ticket to think about this :)

PR #749, opened 2025-09-24
fix: run inference with less concurrency

olaughter:
Tightens the semaphore with the aim of hitting less transient issues accessing s3

TOWARDS PLA-800


THOR300 (approved):
LGTM

PR #753, opened 2025-09-24
fix: Align classifiers profiles across scripts

jesse-c:
This was causing us grief. There's still more work to be done, for other
metadata to be in similar places.

I've also aligned other things, like additional options and options ordering.

### DIFF
--- a/justfile
+++ b/justfile
@@ -68,6 +68,35 @@ promote id +OPTS="":
 demote id aws_env:
     uv run demote --wikibase-id {{id}} --aws-env {{aws_env}}
 
+# Deploy (Get, train & deploy) new model to primary for the given AWS environment.
+# Example: just deploy-classifiers sandbox Q123
+deploy-classifier id aws_env +OPTS="":
+    uv run deploy new \
+        {{id}} \
+        --aws-env {{aws_env}} \
+        --train \
+        --promote \
+        {{OPTS}}
+
+# Deploy (Get, train & deploy) new models to primary for the given AWS environment.
+# Example: just deploy-classifiers sandbox 'Q123 Q368 Q374 Q404 Q412'
+deploy-classifiers ids aws_env +OPTS="":
+    #!/bin/bash
+    set -e
+
+    # Convert the ids argument to a list of --wikibase-id arguments
+    ids_args=""
+    for id in {{ids}}; do
+      ids_args="$ids_args --wikibase-id $id"
+    done
+
+    uv run deploy new \
+        --aws-env {{aws_env}} \
+        $ids_args \
+        --train \
+        --promote \
+        {{OPTS}}
+
 # run a model for a specific wikibase ID on a supplied string
 label id string:
     uv run python scripts/label.py --wikibase-id {{id}} --input-string {{string}}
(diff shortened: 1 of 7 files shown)


jesse-c:

- #753 ◀

scripts/deploy.py
```diff
@@ -122,6 +147,7 @@ def new(
                         classifier_id=classifier.id,
                         aws_env=aws_env,
                         primary=True,
+                        add_classifiers_profiles=add_classifiers_profiles,
```
jesse-c, line 151: **note:** Isn't really needed since it's put there in training, but I did it anyway.

scripts/deploy.py
```diff
@@ -45,10 +45,29 @@ def existing(
...
+    ] = None,
+    remove_classifiers_profiles: Annotated[
+        list[str] | None,
+        typer.Option(help="Removes 1 or more classifiers profiles."),
+    ] = None,
 ):
     """Deploy existing models from one environment to another."""
     validate_transition(from_aws_env, to_aws_env)
 
+    add_class_prof: set[str] = (
```
jesse-c, line 60: **note:** Still TBD about what can be shared where, so this is repeated.

justfile
```diff
@@ -155,24 +184,6 @@ serve-static-site tool:
 generate-static-site tool:
     uv run python -m static_sites.{{tool}}
 
-# Deploy (Get, train & deploy) new models to primary for the given AWS environment.
```
jesse-c, line 158: **note:** Moved it to be next to the related targets.

scripts/deploy.py
```diff
@@ -45,10 +45,29 @@ def existing(
...
+        set(add_classifiers_profiles) if add_classifiers_profiles else set()
```
kdutia, line 61: **nitpick:** might be more readable

```suggestion
        set(add_classifiers_profiles) or set()
```

kdutia (commented):
wasn't sure whether you wanted a ✅ yet as it's still in draft, but i made some nonblocking comments/suggestions

scripts/deploy.py
jesse-c: Sadly I don't think I can do this, it can be `None`[^1]? I might be missing something though!


[^1]:
```
>>> set(None) or set()
  File "<python-input-1>", line 1, in <module>
    set(None) or set()
    ~~~^^^^^^
TypeError: 'NoneType' object is not iterable
```

scripts/promote.py
```diff
@@ -125,14 +125,22 @@ def main(
...
+        list[str] | None,
+        typer.Option(help="Removes 1 or more classifiers profiles."),
+    ] = None,
 ):
     """
     Promote a model to the registry so it can be used downstream.
 
     The model should already have been trained, meaning it will exist
-    as an artefact in wandb, with a linked model in s3 for the environment.
+    as an artefact in W&B, with a linked model in s3 for the environment.
```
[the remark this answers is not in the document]
jesse-c: ✓

scripts/deploy.py
kdutia: oh whoops! tired brain

justfile
jesse-c: All yours!

olaughter: [First slice of this](https://github.com/climatepolicyradar/knowledge-graph/pull/755)

PR #754, opened 2025-09-24
docs: make it clear where we are deploying in ci

olaughter:
Motivated by the fact it is just not clear at the moment:
image

image

See current for comparison

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -12,6 +12,7 @@ permissions:
 
 jobs:
   deploy_prefect_labs:
+    name: "Labs"
     if: |
       (github.event.workflow_run.conclusion == 'success' &&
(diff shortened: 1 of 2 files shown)


jesse-c (approved):
This has annoyed me for a while too.

Image

↑ I wonder if instead, it could be "Sandbox / Deploy", to remove the duplicate _deploy_ text?

.github/workflows/cd.yml
```diff
@@ -12,6 +12,7 @@ permissions:
 
 jobs:
   deploy_prefect_labs:
+    name: "Labs Deploy"
```
jesse-c, line 15: **suggestion:**
Image

↑ I wonder if instead, it could be "Sandbox / Deploy", to remove the duplicate _deploy_ text?

```suggestion
    name: "Labs"
```

PR #755, opened 2025-09-24
refactor: move scripts just commands to scripts

olaughter:
This seems like a good light touch isolation for now, we essentially get all of the dev / ci cd commands in the top level file and keep the scripts related one within scripts. There is a more extreme version with files for different functions within each of those things but it just felt over the top


THOR300 (approved):
LGTM

PR #764, opened 2025-09-29
sample and label passages with LLMClassifier

kdutia:
new behaviour in this PR:

- amends `sample.py` to add W&B tracking and sampling from the larger, combined dataset
- new `predict.py` script runs prediction using an already trained model on labelled passages from W&B or local.

also, some centralising of useful methods:

- moving common functions to a new `wandb_helpers` submodule so we can start to ensure consistent IO with W&B
- in `inference.py` – moves function to run inference in parallel out of evaluation script and uses it in evaluate and predict

questions:

- i think there's a little bit of inconsistency about whether to put I/O on the objects or in submodules related to the task or thing you're IOing with. I've continued with this inconsistency here and have no real thoughts on what is best, but would like to hear opinions!


knowledge_graph/classifier/classifier.py
```diff
@@ -81,6 +84,40 @@ def predict_batch(self, texts: list[str]) -> list[list[Span]]:
         """
         return [self.predict(text) for text in texts]
 
+    def _batch_list(self, items: list, size: int) -> list[list]:
```
jesse-c, line 87: **polish:** We have an [existing function](https://github.com/climatepolicyradar/knowledge-graph/blob/main/flows/utils.py#L286-L289) for doing this, as a generator, that can be used instead.

knowledge_graph/classifier/classifier.py
```diff
@@ -81,6 +84,40 @@ def predict_batch(self, texts: list[str]) -> list[list[Span]]:
...
+        """Batch a list into `size`-lengthed lists"""
+        return [items[i : i + size] for i in range(0, len(items), size)]
+
+    def predict_iter(
```
jesse-c, line 91: **suggestion:** Bringing in console effects into non-scripts rubs me the wrong way. I think it should be kept to scripts, so as to not mix the I/O from both—and from Prefect. So it's clear what can/should happen in the different execution environments of local and Prefect.

> i think there's a little bit of inconsistency about whether to put I/O on the objects or in submodules related to the task or thing you're IOing with. I've continued with this inconsistency here and have no real thoughts on what is best, but would like to hear opinions!

My paragraph above touches on this.

jesse-c: This function too seems confusing since it feels like a dupe of `self.predict_batch`, which it in fact calls.

For these 3 additional functions, I would instead refactor `predict_batch` to take an optional batch size, and use the existing function.

jesse-c: That would mean handling the progress reporting in the script, where you're calling this from.

knowledge_graph/classifier/classifier.py
```diff
@@ -172,6 +210,24 @@ def load(cls, path: Union[str, Path], model_to_cuda: bool = False) -> "Classifie
             classifier.pipeline.device = torch.device("cuda:0")  # type: ignore
         return classifier
 
+    @classmethod
+    def load_from_wandb(
```
jesse-c, line 214: **question:** Could this be used then in lieu of https://github.com/climatepolicyradar/knowledge-graph/blob/main/flows/inference.py#L322-L336, in the inference pipeline? That would be very nice!

scripts/predict.py
jesse-c, line 1: **question:** This and `infer.py` seem like they should be 1 script?

knowledge_graph/inference.py
```diff
@@ -0,0 +1,30 @@
+from knowledge_graph.classifier import Classifier
+from knowledge_graph.labelled_passage import LabelledPassage
+
+
+def label_passages_with_classifier(
```
jesse-c, line 5: **suggestion:** I'm really reticent to add another module. I think ideally we'd start reducing them. Could https://github.com/climatepolicyradar/knowledge-graph/blob/main/flows/inference.py#L591-L627 be used instead? It also seems odd to pass in `LabelledPassage`s, which are de facto results of inference, to re-run inference on them.

knowledge_graph/labelled_passage.py
```diff
@@ -94,6 +97,20 @@ def from_argilla_record(cls, record: Record, client: Argilla) -> "LabelledPassag
 
         return cls(text=text, spans=spans, metadata=metadata)
 
+    @classmethod
+    def from_jsonl(cls, jsonl_path: Path) -> list[Self]:
```
jesse-c, line 101: **suggestion:** Could you use https://github.com/climatepolicyradar/knowledge-graph/blob/main/flows/inference.py#L411-L451 instead?

knowledge_graph/labelled_passage.py
```diff
@@ -179,3 +196,26 @@ def sanitise(text: str) -> str:
             }
         )
         return text.translate(normalize_translation)
+
+
+def save_labelled_passages_to_jsonl(
```
jesse-c, line 201: **suggestion:** Similarly so here, to use https://github.com/climatepolicyradar/knowledge-graph/blob/main/flows/inference.py#L411-L451.

knowledge_graph/inference.py
kdutia: > It also seems odd to pass in LabelledPassages, which are de facto results of inference, to re-run inference on them.

this is a pattern that we seem to use already – `scripts/sample.py` produces a set of `LabelledPassages` without any labels! I found it strange, and would have liked for there to be a `Passage` and `LabelledPassage` (or maybe there is, developed by the platform team in parallel to the dsci team doing this!)

kdutia:
closing this in favour of opening separate PRs, starting with https://github.com/climatepolicyradar/knowledge-graph/pull/788

others will be opened once i'm back from leave for the predict and sample scripts – tracked by [SCI-535](https://linear.app/climate-policy-radar/issue/SCI-535/open-prs-for-new-predict-and-sample-scripts)

PR #770, opened 2025-09-30
Feature/pla 871 bugfix failed knowledge graph full pipeline run

THOR300:


olaughter (approved):
Nice catch, thats insane!

PR #774, opened 2025-09-30
fix(train): Don't implicitly promote in train

jesse-c:
Having an AWS environment as an _alias_ means that that version has been
_promoted_. This shouldn't be implicitly done here. It should be explicitly done
in promotion.

The AWS environment is included in the metadata at training.

This means that all models trained since August 2025 have been implicitly
promoted, when they shouldn't have been.

TWOARDS PLA-838

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -155,8 +155,8 @@ def create_and_link_model_artifact(
     )
     artifact.add_reference(uri=uri, checksum=True)
 
-    # Log the artifact to W&B, creating it within a wandb project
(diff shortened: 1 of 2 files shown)


jesse-c:
> Good catch, I guess we inadvertently started considering adding to the registry as the promotion? Glad we are moving to using tags and profiles!
> 
> I thought we did use the alias somewhere, but I can't find where so wondering if I might be misremembering (Or maybe its already been resolved)

You're probably thinking about promotion? You've seen https://github.com/climatepolicyradar/knowledge-graph/pull/776 now too, so it should be all together!

PR #775, opened 2025-09-30
refactor: Use separate entries per classifiers profile

jesse-c:
The original data structure would've meant a much larger refactor, in inference and aggregation.

This approach means that inference, aggregate, and index require no changes. Inference will still take in a list of CSes (classifier spec) or will load a CSes file. Aggregationill still load a CS file. Indexing is unaffected.

W&B requires no changes as well, nor the scripts that interact with the metadata.

I haven't found anywhere that checks for disallowing multiple entries for CSes by Wikibase ID. For inference, it'll write them out to the appropriate path, separate from each other by the classifier ID.

**Test**

Ran `just update-inference-classifiers` and it correctly didn't find differences in the different environments, except for the one change in staging that I just did.

FIXES PLA-838


jesse-c:

- #774
    - #775 ◀

tests/scripts/__snapshots__/test_update_classifier_spec.ambr
```diff
@@ -0,0 +1,32 @@
...
+    dont_run_on:
+    - sabin
+  - wikibase_id: Q222
+    classifier_id: 2345abcd
+    classifiers_profile: experimental
+    wandb_registry_version: v1
+    compute_environment:
+      gpu: true
+    dont_run_on:
+    - sabin
```
olaughter, line 24: Yeah good call, this makes sense!

THOR300:
Looking good! 

Confirming my understanding: 
- A classifier / version is limited to one profile (primary, experimental etc.) 
    - So we may have Q123:v1 primary and Q123:v2 experimental in the staging classifier spec yaml for example. 
- Inference will as it does now just read in the classifier spec file and run on all entries. 
- Aggregation will aggregate both the experimental and primary values.
   - Currently aggregation outputs a [mapping of text block to VespaConcept](https://github.com/climatepolicyradar/knowledge-graph/blob/main/flows/aggregate.py#L234), so this will just be updated to output [this](https://github.com/climatepolicyradar/navigator-infra/blob/main/vespa/navigator_app_dev/schemas/document_passage.sd#L107) data structure for v2 concepts?

PR #778, opened 2025-10-01
Flexible matching of spaces dashes and hyphens for keyword classifier

AnneIsARealProgrammerNow:
- Modified create_pattern() to normalize labels for all labels (negative, positive, case sensitive and not)
- Spaces and hyphens now match interchangeably, as do different dashes
- Policy style guide already suggests labels should be created in this way.


harrisonpim:
FYI, I've added [a separate PR](https://github.com/climatepolicyradar/knowledge-graph/pull/821) which addresses this issue in a different way, with a load more testing and analysis 🙂 worth checking out [this thread](https://climate-policy-radar.slack.com/archives/C0806G4AZG8/p1760541461201419?thread_ts=1759141201.937399&cid=C0806G4AZG8) for some more context

kdutia:
closing as duplicates #821

PR #782, opened 2025-10-01
BERT model training improvements

kdutia:
in this PR:

- in `scripts/train.py`: allows training data to be loaded from W&B via a new `training_data_wandb_run_path` option. Whilst we might eventually store training data in Argilla, that needs more decisions about how we separate it from human-annotated eval sets and other sources of labelled data we might want to view – and maybe new conventions. 
- the BERT-based classifier now takes a `train_validation_data` argument
- a few small fixes for the BERT classifier after training some model variants
  - expands number of epochs to take advantage of early stopping; changes the early stopping threshold to 0 (the default) in response to it not being triggered before
  - adds new `WeightedTrainer`. 21% of the sampled passages from the GPT5 classifier were positive; this led to a run where precision was much higher than recall. My intuition for this was that it was a class imbalance issue. 
  - added gradient clipping as I was seeing some spikes in the training curve
  - made `batch_size` scale a little bit with training dataset size, to avoid epochs that are 2 batches or batches that are 2 passages
 
[Quick W&B report here](https://wandb.ai/climatepolicyradar/Q1016/reports/BERT-model-training-improvements--VmlldzoxNDU5NTQ5Ng) to justify the biggest changes to the training process. The final model is `genial-bird-114` below, which narrowly creeps into territory where it's competitive with gpt-4o-mini.

image


scripts/train.py
```diff
@@ -298,6 +323,12 @@ def main(
         list[str] | None,
         typer.Option(help="Adds 1 or more classifiers profiles."),
     ] = None,
+    training_data_wandb_run_path: Annotated[
+        Optional[str],
+        typer.Option(
+            help="W&B run path (entity/project/run_id) to fetch training data from instead of using concept's labelled passages",
```
jesse-c: **question:** Where do these runs come from?

PR #784, opened 2025-10-01
refactor: Use default work pools

jesse-c:
We've refactored the orchestrator to not create specific work pools. Instead,
the worker creates the `default` work pool, which is a blessed work pool. This
means that a. we don't need to specify the work pool and b. the previous
`mvp-$env` work pools are gone.


jesse-c:

- #784 ◀

THOR300 (commented):
question: The PR description mentions work pools but queues are removed here, just checking that's the intent? Assuming we mean we're now using the default work queues within the work pools so we don't need custom work queues?

jesse-c:
> question: The PR description mentions work pools but queues are removed here, just checking that's the intent?

I made a mistake! 🤦🏻‍♂️ Thank you. I've replaced pool with queue in the title + description.

PR #785, opened 2025-10-01
test: Reset manual OS env. setting

jesse-c:
This function is called in a test, and depending on the ordering, especially when the tests are run concurrently, can mess up other tests[^1].

Ideally there an effect like this isn't needed. For now, cleanup after itself.

TOWARDS PLA-883

[^1]: https://github.com/climatepolicyradar/knowledge-graph/actions/runs/18168137332/job/51717834628

### DIFF
--- a/scripts/evaluate.py
+++ b/scripts/evaluate.py
@@ -79,7 +79,12 @@ def load_classifier_remote(
     version: Version,
     wikibase_id: WikibaseID,
 ) -> Classifier:
-    """Load a classifier from W&B artifacts storage."""
+    """
+    Load a classifier from W&B artifacts storage.
+
+    NB: This function temporarily sets AWS_PROFILE to enable W&B to download
+    artifacts from S3. The profile is restored after download completes.
+    """
     artifact_id = f"{wikibase_id}/{classifier}:{version}"
     artifact = run.use_artifact(artifact_id, type="model")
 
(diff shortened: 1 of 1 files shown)


scripts/evaluate.py
```diff
@@ -79,7 +79,12 @@ def load_classifier_remote(
     version: Version,
     wikibase_id: WikibaseID,
 ) -> Classifier:
-    """Load a classifier from W&B artifacts storage."""
+    """
+    Load a classifier from W&B artifacts storage.
+
+    NB: This function temporarily sets AWS_PROFILE to enable W&B to download
+    artifacts from S3. The profile is restored after download completes.
```
olaughter, line 86: I'm curious about the underlying context of doing this because it seems like a bad smell to me. At first I wondered if it was because multiple profiles are needed but I don't see signs of that. Perhaps we just take out the profile setting so that users will need to choose a profile before running the script, following the pattern of other scripts?

olaughter (commented):
Thanks for jumping on this and catching where this is happening! Suggesting below we take a more extreme option and just delete the problem code 😄

scripts/evaluate.py
jesse-c: I think that's feasible, but would require more work. I just was going for the bare minimum to unblock myself/others for now!

olaughter:
TOWARDS PLA-883

olaughter:
> TOWARDS PLA-883

I guess this doesn't work 😆 added a link on the ticket instead

PR #786, opened 2025-10-02
Feature/pla 885 fix requestistooskewed error again

olaughter:
The connection pool defaults to 10! 

Meaning that when attempting to run concurrently there was a significant opportunity for the connections to time out. This is why the semaphore made a difference to our networking errors, because it was getting closer to 10. In order to actually get proper use of the concurrency we can instead increase the connection pool. Adding a buffer on top seems sensible as it will mitigate some not releasing back to the pool.

Note that I applied this to one of the runs [in production to confirm the theory](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/068de5be-dd66-7e35-8000-a4a583589adf?entity_id=068de5be-dd66-7e35-8000-a4a583589adf&state=failed&state=crashed&state=completed&state=running&entity=flowRuns&entity=artifacts&entity=logs&log-level=30). I'd typically see this as bad practice generally but it seemed right in this situation. Its a fix for something thats broken, and I wanted to be remove any doubt by tightly limiting the possible variables. It previously looked like I fixed the issue when running in staging. I have held off doing a larger scale run which would also be a good next step once merged.

See: b8f10305e66d81e0c01042e803dbd0383c02006e
See also docs: https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html

Drivebys:
- Allow `deployments.py` to be imported with out deploying as a side effect
- Tweak prefect artifact script to customise print behaviour
- Revert the tenacity retry approach which didn't actually help and just slowed down surfacing the error
- Increase the concurrency limit!


THOR300 (approved):
This looks great in getting us back to a position where we can control concurrency! Hats off to you around the max connection pools find. 

No worries re testing in prod here if it wasn't possible to reproduce e elsewhere easily as it makes sense to optimise for solving this problem fast. 

question: you mention in the second run most of the issues seen were read time out, did the skewed error still persist just in smaller proportion?

olaughter:
> did the skewed error still persist just in smaller proportion

Yes much smaller, I halved the semaphore again after that run, so that will help. Unfortunately it does mean this work is likely not finished yet and, will likely need another PR. But this one has made progress at least!

> wondering whether we want to set a max limit for s3 concurrency

Do you mean a check in the config to stop us from configuring it in silly ways? I guess we don't know yet what the max would be but I can look at following up on this once I find it (assuming I've understood your thought?)

PR #787, opened 2025-10-02
chore: re-enable targets

olaughter:
We had turned off targets previously as they where blocking other classifiers. This will allow the concepts api to pick up on them again. Who knows maybe this is the last time we have to do this!

To Do
- [ ] Finish indexing
- [ ] Redeploy concepts-api


THOR300 (approved):
LGTM

PR #788, opened 2025-10-02
better classifier prediction over many texts

kdutia:
The first of a set of PRs to replace #764, as that one was big enough that it led to a bit of an overwhelming review task. 

In this PR:

- `Classifier.predict` can now take `str` or `list[str]` as input arguments. This is similar to the implementation in [sentence-transformers](https://sbert.net/docs/package_reference/sentence_transformer/SentenceTransformer.html#sentence_transformers.SentenceTransformer.encode). The `abstractmethod` for `Classifier` is now `._predict`, and `.predict` handles batching logic.
- implements the longstanding needed typing fix for `predict_batch`: `texts` is now `Sequence[str]` rather than `list[str]`
- uses `nest_asyncio` more neatly in the LLMClassifiers to ensure that they can be run in an async outer loop

### questions you might have

**Won't adding progress bars to the `predict` method give me headaches if i'm running it in prefect?**

No! The progress bars are switched off by default, for this reason. Given a bunch of my job is running predict and train loops manually on my machine (often overnight), it's a requirement for me to be able to have a sense for how long they are. So, we can turn progress bars on for scripts we run locally i.e. experimental ones, and they'll be off for any prefect tasks.

**Is `labelling.py` a strange place to move the `label_passages_with_classifier` method to?**

I think it's okay! I moved it to a new `inference.py` submodule before but @jesse-c rightly said we're starting to get too many submodules. This aligns with chats i've been having with @harrisonpim that we should treat human and model labellers the same.

### DIFF
--- a/tests/common_strategies.py
+++ b/tests/common_strategies.py
@@ -43,7 +43,7 @@ def multi_word_label_strategy(
         list[str]
     ] = None,  # default to space-separated labels for most tests
 ):
-    """Generate a multi-word label, separated by the supplied separator characters"""
+    """Generate a multi-word label, separated by the supplied separator characters."""
     num_words = draw(st.integers(min_value=min_words, max_value=max_words))
     words = [draw(single_word_label_strategy) for _ in range(num_words)]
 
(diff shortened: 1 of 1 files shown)


tests/test_classifier.py
olaughter, line 1: suggestion: I'm thinking it would be good to have tests for each classifier to cover regressions

olaughter (approved):
This is really cool! I'm assuming we'd need to retrain all the classifiers to update the pickled models in order to use this in the pipelines?

jesse-c (commented):
> Classifier.predict_batch has the wrong output type to give us a clean interface to run lots of texts through a model. As is common (e.g. [in sentence-transformers](https://sbert.net/docs/package_reference/sentence_transformer/SentenceTransformer.html#sentence_transformers.SentenceTransformer.encode)), we want to be able to run multiple texts through a model without having to deal with batching the input or flattening the output each time. Another option would be to make predict have input type text: str | Sequence[str]. I'm not against this in principle, but the way inheritance is set up at the moment means I think this would need a bigger refactor.

I'm struggling to understand this!

```python
    def predict_batch(self, texts: Sequence[str]) -> list[list[Span]]:
```

```python
    def predict_many(self, texts: list[str], batch_size: int, show_progress: bool = False) -> list[list[Span]]:
```

They have same input (`list` and `Sequence` are functionally the same) and the same output? Am I seeing things? 😅

knowledge_graph/labelling.py
```diff
@@ -308,3 +309,33 @@ def _assert_datasets_of_the_same_type(*datasets: Dataset):
...
+
+    output_labelled_passages = [
+        labelled_passage.model_copy(
+            update={"spans": model_predicted_spans[idx]},
+            deep=True,
+        )
+        for idx, labelled_passage in enumerate(labelled_passages)
+    ]
+
+    return output_labelled_passages
```
harrisonpim: i don't think this belongs in this module! 

`labelling.py` is admittedly a bad name - it's supposed to handle the interface between this codebase and  our human labelling system (ie argilla). I'd much prefer to call this module `argilla.py`, but doing so causes all sorts of horrible import issues when importing `argilla` vs `knowledge_graph.argilla` 🫠 

1. if you can think of a clearer name for this file, pls suggest one 😅
2. i still think that this function needs a different home

harrisonpim (approved):

PR #796, opened 2025-10-08
Updating documentation for the Concept Store.

THOR300:
This Pull Request: 
---
- Adds some documentation for user management with wikibase. 

Note: This would have been useful to me as I had no idea whether users were created via the ui, cli, api, config, just command etc. and the wikibase documentation isn't all that easy to pick up. Our approach to rolling passwords also wasn't that clear. If people don't think this sits here however, more than happy to move / adapt.


michaelokarimia:
I recommend closing this PR and starting again, using `[skip ci]` keyword in the commits to prevent the CI/CD pipeline from being triggered
https://docs.github.com/en/actions/how-tos/manage-workflow-runs/skip-workflow-runs

PR #798, opened 2025-10-08
Update promote script to use specific artifact version instead of alias

kaparker:
* Update `artifact_id` to use `version` instead of `alias` (aws env) to get the artifact to be promoted 
* Determine `version` by retrieving all artifacts for the model path (wikibase id / classifier id) and select latest model version for the aws env
* Update README with changes
* Add comments to mocks in test_promote for clarity

TOWARDS PLA-890

### DIFF
--- a/.github/workflows/prefect_deploy.yml
+++ b/.github/workflows/prefect_deploy.yml
@@ -99,15 +99,23 @@ jobs:
       - name: Confirm package availability
         run: docker run --rm -e AWS_ENV=${{ inputs.aws-env }} ${{ steps.login-ecr.outputs.registry }}/${{ github.event.repository.name }}:${{ steps.get_version.outputs.version }} python -c "import flows.inference; print('flows.inference imported successfully')"
 
-      - name: Deploy to Prefect
+      - name: "Deploy to Prefect: Deployments"
         env:
           AWS_ENV: ${{ inputs.aws-env }}
           AWS_REGION: eu-west-1
           DOCKER_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
           DOCKER_REPOSITORY: ${{ github.event.repository.name }}
-          DOCKER_TAG: ${{ steps.get_version.outputs.docker_tag }}
           PREFECT_API_KEY: ${{ secrets.PREFECT_API_KEY }}
           PREFECT_API_URL: ${{ secrets.PREFECT_API_URL }}
-        run: |
-          uv run python -m deployments
-          uv run python -m automations
+          GIT_BRANCH: ${{ github.ref_name }}
+        run: uv run python -m deployments
+
+      - name: "Deploy to Prefect: Automations"
+        env:
+          AWS_ENV: ${{ inputs.aws-env }}
+          AWS_REGION: eu-west-1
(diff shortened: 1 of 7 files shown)


jesse-c (commented):
This is a confusing bit. The canonical IDs for classifiers, ["u86a8p2d"](https://wandb.ai/climatepolicyradar/Q1653/artifacts/model/u86a8p2d) in the screenshot below, were to replace needing to specify a model architecture _and_ a model version. Since aliases were being set before, "incorrectly", it did enable not having to need a version since made it so that the last trained version was always used. That's because an alias can only be on one version, and because we were setting it in training.

To emulate part of that behaviour, you'd need to change it use the last trained version, that has the appropriate AWS env. metadata.

@olaughter, could you double-check that, as I was never 100% on how the double layering of versioning (from the canonical ID first and then artifact version second) was supposed to function. I'm also confused how we end up with so many versions of artifacts, instead of new artifacts (AKA new canonical IDs), since I'd expect changes in a concept to mean that there's a new hash (AKA new canonical ID)?

Image

kaparker:
> @olaughter, could you double-check that, as I was never 100% on how the double layering of versioning (from the canonical ID first and then artifact version second) was supposed to function. I'm also confused how we end up with so many versions of artifacts, instead of new artifacts (AKA new canonical IDs), since I'd expect changes in a concept to mean that there's a new hash (AKA new canonical ID)?

It looks like the [canonical ID is set based on the classifier type](https://github.com/climatepolicyradar/knowledge-graph/blob/main/scripts/train.py#L403) so if the same classifier is being used then the ID is retained but for each model retraining a new version is created
``` 
# Note that as we use the id and the id changes whenever the model changes,
# the version would almost always be v0 in practice.
```

jesse-c:
> > @olaughter, could you double-check that, as I was never 100% on how the double layering of versioning (from the canonical ID first and then artifact version second) was supposed to function. I'm also confused how we end up with so many versions of artifacts, instead of new artifacts (AKA new canonical IDs), since I'd expect changes in a concept to mean that there's a new hash (AKA new canonical ID)?
> 
> It looks like the [canonical ID is set based on the classifier type](https://github.com/climatepolicyradar/knowledge-graph/blob/main/scripts/train.py#L403) so if the same classifier is being used then the ID is retained but for each model retraining a new version is created
> 
> ```
> # Note that as we use the id and the id changes whenever the model changes,
> # the version would almost always be v0 in practice.
> ```

It takes in so little data, so I'm still surprised[^1]. I think this is a mistake though. It should take in perhaps the regexes, since they count as "behaviour-driving parameters"[^2], @climatepolicyradar/datascience?

Anyway, this is beyond the scope of this issue.

What it whittles down to is that someone could train x 1000 times, and we'll end up with that many artifacts, and that's fine/not something we'll prevent.

I'd say that that means we should go with my idea of the latest version for that environment. That carries along the implicit behaviour anyway, and we could change it later.

[^1]: https://github.com/climatepolicyradar/knowledge-graph/blob/feat%2Fuse-aggregate-config-in-index/knowledge_graph/classifier/keyword.py#L111-L114
[^2]: https://github.com/climatepolicyradar/knowledge-graph/blob/feat%2Fuse-aggregate-config-in-index/knowledge_graph/classifier/classifier.py#L115

PR #799, opened 2025-10-08
ci: Log version used for Prefect and correctly run deployments

jesse-c:
This uses the same logging approach as in the automation script.

Originally I did this so I could get insight into what was being deployed, as I noticed our deployments are on an old version. I then found a bug in the script, which is now fixed.

I separated deployments and automations so it's easier to separate in CD on GitHub.

**Test**

```
$ AWS_ENV=sandbox uv run python -m deployments
      Built knowledge-graph @ file:///Users/jesse/src/github.com/climatepolicyradar/knowledge-graph
Uninstalled 1 package in 3ms
Installed 1 package in 3ms
using version: 0.8.0
15:24:38.654 | INFO    | __main__ - using version: 0.8.0
Successfully created/updated all deployments!

                        Deployments
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━┓
┃ Name                                 ┃ Status  ┃ Details ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━┩
│ train-on-gpu/kg-train-on-gpu-sandbox │ applied │         │
└──────────────────────────────────────┴─────────┴─────────┘

To schedule a run for this deployment, use the following command:

        $ prefect deployment run 'train-on-gpu/kg-train-on-gpu-sandbox'


You can also run your flow via the Prefect UI: https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/deployments/deployment/99548d7e-9679-4f22-90b9-4a10f3c94b45

...
```

CleanShot 2025-10-08 at 15 28 55

CleanShot 2025-10-08 at 15 45 46

### DIFF
--- a/justfile
+++ b/justfile
@@ -66,16 +66,16 @@ export-env-vars:
 prefect-login: export-env-vars
 	prefect cloud login -k ${PREFECT_API_KEY}
 
-deploy: prefect-login
+deploy:
     just deploy-deployments
     just deploy-automations
 
-deploy-deployments: prefect-login
-	aws ecr get-login-password --region ${AWS_REGION} | docker login --username AWS --password-stdin ${DOCKER_REGISTRY}
-	python -m deployments
+deploy-deployments:
+    aws ecr get-login-password --region ${AWS_REGION} | docker login --username AWS --password-stdin ${DOCKER_REGISTRY}
+    uv run python -m deployments
 
-deploy-automations: prefect-login
-	python -m automations
+deploy-automations:
+    uv run python -m automations
 
 # Run a static site locally
 serve-static-site tool:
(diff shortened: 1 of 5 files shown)


justfile
```diff
@@ -66,16 +66,16 @@ export-env-vars:
 prefect-login: export-env-vars
 	prefect cloud login -k ${PREFECT_API_KEY}
 
-deploy: prefect-login
+deploy:
     just deploy-deployments
     just deploy-automations
 
-deploy-deployments: prefect-login
```
kaparker, line 73: Is prefect-login no longer required? Or should `just prefect-login` be run prior to deployment commands?

jesse-c: No wonder uses it TBH, AFAIK. It's not used in CD. Buried in the commit message I've got:

> The Prefect API key is set in .env.

Though I would add that yes, you should be logged in already. You'll find out quickly if you're not! Having it required to run those targets, when the Justfile loads .env, didn't make sense. It also makes this closer to being usable in CD.

PR #800, opened 2025-10-08
feat(aggregate): Store metadata for runs

jesse-c:
Similar to #793, do so for aggregation. This data will be loaded by indexing. Indexing will use that data to enrich the aggregated results.

I considered updating the results that aggregation stored, but found this approach faster, and keeps the results stable as is. I think we could refactor them, but outside of the current project.

**Test**

- https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/068e698f-c324-7b6e-8000-40d7bc0f13c9?preview=true&tab=logs&log-level=0
- `s3://../inference_results/2025-10-08T17:05-cyan-starfish/metadata.json`

FIXES PLA-832


jesse-c:

- #800 ◀
    - #801

THOR300 (approved):
I don't have much context at the higher level as to what the data flow here is / how this fits in so I can't review with respect to that. But I can say that the implementation / code quality looks fabulous. 

What I'm understanding is that:
- Inference currently writes out run metadata. 
- Aggregation is not currently pulling that in but writing out it's own metadata along with the aggregation results. 
- Indexing will then load aggregation metadata and write out it's own. 

Considering that this `store_metadata` call doesn't allow any exceptions to raise to me means this is safe to merge. Also just one write per aggregate flow so no additional s3 shenanigans 👍

PR #1048, opened 2026-02-24
chore(train metadata): include BERT model config

olaughter:
When a model is a bert model its useful having the config to hand. For example we would have benefited from this today while debugging a dependency drift issue. including this in the wandb metadata means we can look this up when we need to for future trained models.


jesse-c (approved):
Were you able to do a test train locally with this?

kaparker (approved):

kdutia (approved):

olaughter:
This stopped feeling necessary after discovering the run can be found in wandb with all the necessary info. Closing

PR #1091, opened 2026-03-24
fix: Remove duplicate entries

jesse-c:
The models were retrained last week. The changes to the models though weren’t accounted for in the ID generation, so there was a new W&B version, but not a new canonical ID. The classifier ID in the Concept Store therefore wasn’t changed, since it was the same. This was why there was no update operations, and only the W&B model registry changes.

This removes the dupe entry, and gets it back to a good state.

I had to manually remove the `production` label from the old entries, in W&B. I ran `just update-inference-classifiers --aws-envs prod` to confirm that they weren't re-added, and that there are no changes.

TOWARDS PLA-1284

### DIFF
--- a/flows/classifier_specs/v2/production.yaml
+++ b/flows/classifier_specs/v2/production.yaml
@@ -430,33 +430,12 @@
   concept_id: anhcmh9a
   classifier_id: 72tmt7zm
   classifiers_profile: primary
-  wandb_registry_version: v15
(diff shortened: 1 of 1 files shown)


jesse-c:

- #1091 ◀

jesse-c:
Using https://github.com/climatepolicyradar/knowledge-graph/pull/1092 instead.

PR #1093, opened 2026-03-24
fix: add gpu specific limits

olaughter:
We were getting rate limited by aws previously for gpu instances and experienced crashes. This limits gpu classifier inference & cpu classifier inference independently. There is more that we can do to improve this, especially around distribution. But this is intended as a tactical, time-efficient improvement for starters

### DIFF
--- a/flows/inference.py
+++ b/flows/inference.py
@@ -78,7 +78,8 @@ BLOCKED_BLOCK_TYPES: Final[set[BlockType]] = {
     BlockType.FIGURE,
 }
 
-CLASSIFIER_CONCURRENCY_LIMIT: Final[PositiveInt] = 20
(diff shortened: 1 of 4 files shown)


kdutia (approved):
I wonder whether we're actually setting the limit on the wrong thing here? I.e. is the issue that BERT models require more compute, so we should limit concurrency more when we're using them – and they just so happen to do inference on GPU?

Not sure it matters, so have approved anyway 😅

olaughter:
> so we should limit concurrency more when we're using them

This is definitely factor, but not the specific issue we were seeing. Each sub flow deployment runs on a separate host, one classifier per machine, and aws was crashing new clusters due to a limit imposed from their side.

That being said:

> BERT models require more compute

Inference is massively under optimised from this perspective, so would like to get some improvements for it into the product prioritisation

PR #1098, opened 2026-03-30
Q912 llmclassifier fixes

kdutia:
- remove persona template (it performs worse consistently; i think we decided on Slack that you don't really need personas with reasoning models)
- remove old way of interacting with AutoLLMClassifier. We just sub "AutoLLMClassifier" into the `model_name` kwarg in `train` instead

### DIFF
--- a/scripts/custom_concept_training/q912_train_llm.py
+++ b/scripts/custom_concept_training/q912_train_llm.py
@@ -194,43 +194,6 @@ def train() -> None:
 
 
 @app.command()
-def autollm(
-    n_trials: int = typer.Option(5, help="Number of optimization trials to run"),
-    beta: float = typer.Option(
-        0.5, help="Beta value for f-beta score (default 1.0 = F1)"
-    ),
-    optimiser_model: str = typer.Option(
-        "openrouter:google/gemini-3-pro-preview",
-        help="Model to use for prompt optimization",
-    ),
-    final_model: str = typer.Option(
-        "openrouter:google/gemini-3-pro-preview",
-        help="Model to use for the final classifier",
-    ),
-) -> None:
-    """Train an AutoLLMClassifier that automatically optimizes its own prompt."""
-    concept_overrides = get_concept_overrides()
-
-    console.print("Training AutoLLMClassifier with automatic prompt optimization")
-    asyncio.run(
-        run_training(
-            wikibase_id=WIKIBASE_ID,
-            track_and_upload=True,
-            aws_env=AwsEnv.labs,
-            classifier_type="AutoLLMClassifier",
-            classifier_kwargs={
-                "model_name": "openrouter:google/gemini-3-flash-preview",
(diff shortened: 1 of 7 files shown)


scripts/custom_concept_training/q912_train_llm.py
```diff
@@ -41,30 +41,6 @@
...
-can also be a meaningful step towards real influence in projects
-and society.
-
-You will mark up references to concepts with XML tags.
-
-First, carefully review the following description of the concept:
-
-<concept_description>
-{concept_description}
-</concept_description>
```
harrisonpim, line 65: question: is the concept description not needed?

scripts/custom_concept_training/q912_train_llm.py
```diff
@@ -41,30 +41,6 @@
...
-"""
-
 
 def get_labelling_guidelines() -> str:
```
harrisonpim, line 45: sidebar: i feel increasingly uncomfortable about the coupling of concept definitions and code here - it feels like we're slowly undoing all of the good decoupling work that we did at the start of the knowledge graph work (programmes people focus on concepts, data scientists focus on classifier architectures). 

Obviously the _results_ of this work are fantastic, but structurally, it feels like scaling this in its current form is going to be very difficult. From a transparency POV, it's also not ideal that readers of the concept store won't be able to see our full definitions of some concepts without navigating our janky codebase...

No need to address this right now, but it would be good to start sharing thoughts on how we could eg overcome wikibase's character limit, or expose a new vbe-checker-y interface for editing prompts, or ???something???else??? to decouple the classifier development work again

scripts/custom_concept_training/q912_train_llm.py
AnneIsARealProgrammerNow: yeah it's still included in the default template. We don't need this persona-focussed override anymore.

scripts/custom_concept_training/q912_train_llm.py
kdutia: sounds like we need another "what are the issues with the knowledge graph workflow" workshop. i can add lots more :)

AnneIsARealProgrammerNow: For the transparency part, at minimum we could make it part of the Definition of Done for these classifiers to update the definition in the concept store and have a public version of the final prompt (e..g notion webpage?) that we link to in CC too (discussion page? new 'methodology' property?). 

The broader point on needing to do more decoupling & cutting LLM classifier development into manageable chunks with appropriate tooling: big yes, with chat interface experiments etc possibly being a step towards this?  But yeah this is not the forum.

AnneIsARealProgrammerNow (commented):
LGTM. I'm going to pull this before making other Q912 tweaks

kdutia:
merging because of @AnneIsARealProgrammerNow 's LGTM and the fact that this only touches experimental code

PR #1108, opened 2026-04-07
token-based BERT classifier

kdutia:
[BIO](https://en.wikipedia.org/wiki/Inside%E2%80%93outside%E2%80%93beginning_(tagging))-based BERT classifier + benchmark script

Tested on _Q1829 finance flow_, this scored a passage-level F1 score on par with the [best passage-based BERT model for the concept](https://wandb.ai/climatepolicyradar/Q1829/runs/cgscfdfs?nw=nwuserkdutia). Some layers had to be unfrozen for it to perform as well as it did.

It's natural that it'd perform worse than the passage-level classifier at span level as the passage-level classifier labels _everything_, but I'm not sure why it performs so badly. Either way, I think it's worth merging so we can do a fully-W&B-instrumented training run to compare some predictions. 

``` text
    Group   Agreement at        Precision   Recall   Accuracy   F1 score   Support                                                                                                                  
────────────────────────────────────────────────────────────────────────────────                                                                                                                 
    all     Passage level       0.78        0.87     0.83       0.82       644                                                                                                                      
    all     Span level (0)      0.60        0.76     0.67       0.67       827                                                                                                                      
    all     Span level (0.5)    0.38        0.48     0.51       0.42       855                                                                                                                      
    all     Span level (0.9)    0.23        0.27     0.42       0.25       862                                                                                                                      
    all     Span level (0.99)   0.09        0.11     0.36       0.10       881                                                                                                                      


# comparable results for passage level classifier

all   Passage level       0.75        0.91     0.82       0.82       644
all   Span level (0.5)    0.41        0.49     0.54       0.45       762

```

### DIFF
--- a/knowledge_graph/classifier/bert_token_classifier.py
+++ b/knowledge_graph/classifier/bert_token_classifier.py
@@ -587,6 +587,8 @@ class BertTokenClassifier(
             for name, _ in self.model.named_parameters():
                 if match := re.search(r"layers?\.(\d+)\.", name):
                     layer_indices.add(int(match.group(1)))
+                else:
+                    logger.warning(f"No layers found in the model with name {name}")
             if layer_indices:
                 max_layer = max(layer_indices)
                 unfrozen_layer_indices = {
--- a/knowledge_graph/classifier/bert_token_classifier.py
+++ b/knowledge_graph/classifier/bert_token_classifier.py
@@ -278,7 +278,7 @@ class BertTokenClassifier(
         concept: Concept,
         model_name: str = "answerdotai/ModernBERT-base",
         download_pretrained_model_on_init: bool = True,
-        unfreeze_layers: int = 0,
+        unfreeze_layers: int = 2,
     ):
         super().__init__(concept)
         self.model_name = model_name
--- a/knowledge_graph/classifier/bert_token_classifier.py
+++ b/knowledge_graph/classifier/bert_token_classifier.py
@@ -1,0 +1,734 @@
+import logging
+import os
+import random
+import re
+import tempfile
+from collections.abc import Sequence
+from contextlib import contextmanager
+from datetime import datetime
+
+import numpy as np
+import pandas as pd
+import torch
+from datasets import Dataset
+from rich.logging import RichHandler
+from sklearn.metrics import accuracy_score
+from sklearn.model_selection import train_test_split
+from sklearn.utils.class_weight import compute_class_weight
+from transformers import (
+    AutoModelForTokenClassification,
+    AutoTokenizer,
+    EarlyStoppingCallback,
+    EvalPrediction,
+    PreTrainedModel,
+    PreTrainedTokenizer,
+)
+from transformers.trainer import Trainer
+from transformers.training_args import TrainingArguments
+from typing_extensions import Self
+
+from knowledge_graph.classifier.classifier import (
+    Classifier,
+    GPUBoundClassifier,
+    ProbabilityCapableClassifier,
+    VariantEnabledClassifier,
+)
+from knowledge_graph.concept import Concept
+from knowledge_graph.identifiers import ClassifierID, WikibaseID
+from knowledge_graph.labelled_passage import LabelledPassage
+from knowledge_graph.span import Span
+
+logging.basicConfig(handlers=[RichHandler()])
+logger = logging.getLogger(__name__)
+
+# BIO label scheme
+O_LABEL = 0
+B_LABEL = 1
+I_LABEL = 2
+IGNORE_LABEL = -100
+NUM_LABELS = 3
+
+ID2LABEL = {O_LABEL: "O", B_LABEL: "B-CONCEPT", I_LABEL: "I-CONCEPT"}
+LABEL2ID = {v: k for k, v in ID2LABEL.items()}
+
+
+def _align_labels_with_tokens(
+    offset_mapping: list[tuple[int, int]],
+    spans: list[Span],
+    concept_id: WikibaseID | None,
+) -> list[int]:
+    """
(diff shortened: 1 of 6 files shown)


AnneIsARealProgrammerNow:
Quick thought that might be wildly off: for finance flow, a bunch of the labels were from excel sheets rather than argilla. Given that "support" shows hundreds of labels, I think these were included. They are effectively paragraph-level labels, so depending on how they were added, that might explain the low scores?

scripts/train.py
```diff
@@ -176,8 +176,10 @@ def move_model_to_cpu(classifier: Classifier) -> None:
 
     console = Console()
 
-    if isinstance(classifier, BertBasedClassifier):
-        classifier.move_model_to_device(torch.device("cpu"))
+    if isinstance(classifier, BertBasedClassifier) or hasattr(
+        classifier, "move_model_to_device"
+    ):
```
harrisonpim, line 179: suggestion: could we check whether a classifier is an instance of `GPUBoundClassifier` here, instead of the dual condition?

knowledge_graph/classifier/bert_token_classifier.py
```diff
@@ -0,0 +1,734 @@
...
+        texts = [passage.text for passage in labelled_passages]
+
+        tokenized = self.tokenizer(
+            texts,
+            padding=True,
+            truncation=True,
+            max_length=512,
+            return_tensors=None,
+            return_offsets_mapping=True,
+        )
```
harrisonpim, line 419: suggestion: we've still got the `max_length` capped here, which might end up causing problems down the line. i haven't looked at the distribution of lengths in the training data, but i suspect we already have passages longer than 512 tokens in there. could we use a data collator to resolve the padding limitation? https://huggingface.co/docs/transformers/en/main_classes/data_collator#transformers.DataCollatorForTokenClassification

knowledge_graph/classifier/bert_token_classifier.py
```diff
@@ -0,0 +1,734 @@
...
+            self.model.train(was_training)
+
+    def get_variant(
+        self, random_seed: int | None = None, dropout_rate: float = 0.1
+    ) -> Self:
+        """Get a variant of the classifier for Monte Carlo dropout estimation."""
+        variant = self.__class__(
+            concept=self.concept,
+            model_name=self.model_name,
+            download_pretrained_model_on_init=True,
```
harrisonpim, line 395: question: do we need to `download_pretrained_model_on_init` if we're going to fill it with the existing state dict on the next line?

knowledge_graph/classifier/bert_token_classifier.py
```diff
@@ -0,0 +1,734 @@
...
+
+
+def _compute_token_metrics(eval_pred: EvalPrediction) -> dict[str, float]:
+    """
+    Compute token-level metrics from token predictions.
+
+    Note: precision/recall/F1 here are token-level, not span-level. A 5-token
+    entity counts as 5 positives. This is used for early stopping during
+    training; the production quality signal is the span-level Jaccard metric
+    computed by evaluate_classifier.
```
harrisonpim, line 202: really good point, which i'm sure will affect training when we do big runs! claude suggests https://github.com/chakki-works/seqeval as a possible upgrade here?

knowledge_graph/classifier/bert_token_classifier.py
harrisonpim, line 1: There's a lot of duplication in here with our existing `BertBasedClassifier` which we could probably get rid of via a deliberate little refactor. Personally, I'd like to do the same so that we can more neatly integrate and experiment with a potential LORA-based classifier implementation. Maybe that's something we can come back to together?

knowledge_graph/classifier/bert_token_classifier.py
```diff
@@ -0,0 +1,734 @@
...
+
+        logger.info(stats_df.to_markdown(index=False, tablefmt="rounded_grid"))
+
+        # Determine which transformer layers to unfreeze (if any)
+        unfrozen_layer_indices: set[int] = set()
+        if self.unfreeze_layers > 0:
+            layer_indices = set()
+            for name, _ in self.model.named_parameters():
+                if match := re.search(r"layers?\.(\d+)\.", name):
+                    layer_indices.add(int(match.group(1)))
```
harrisonpim, line 589: suggestion: given that the base `model_name` and the `unfreeze_layers` are both tweakable by fallible human users, we probably want to log a warning here if the regex doesn't hit anything across the full model 🤔

harrisonpim (approved):
Looks great! Very exciting to be coming back round to token-level labelling 🕺 

Agree that it's worth doing a fully instrumented and logged run to figure out what's going on with those passage- vs span-level metrics. @AnneIsARealProgrammerNow's suggestion seems plausible to me but I haven't been able to find enough in the logged runs to figure out what's going on myself 😞

knowledge_graph/classifier/bert_token_classifier.py
kdutia: i thought about this, but figured that a token-based metric is likely to be smoother in early epochs than a span-based one

knowledge_graph/classifier/bert_token_classifier.py
kdutia: the state dict is just the values of the weights i believe, rather than the architecture itself. reloading the model on init saves us from potential clashes in compatibility between library and model versions

knowledge_graph/classifier/bert_token_classifier.py
kdutia: https://linear.app/climate-policy-radar/issue/SCI-787/investigate-silent-truncation-of-passages-to-512-tokens-in

PR #1110, opened 2026-04-07
Predict adhoc flow

kdutia:
Data science use this script as part of the model distillation process, to generate training data for a BERT model using an LLM. This puts it on Prefect 🤞 

[Related PR to add secret to SSM via Pulumi here](https://github.com/climatepolicyradar/orchestrator/pull/305)


flows/predict.py
```diff
@@ -0,0 +1,77 @@
...
+    session = boto3.session.Session(
+        profile_name=aws_env.value if use_aws_profiles else None,
+        region_name=config.bucket_region,
+    )
+    session.client("s3")
+
+    return config
+
+
+@flow()
```
THOR300, line 45: **thought**: If you want to be notified via slack upon failure of this flow you could look to implement the `SlackNotify` class here - https://github.com/climatepolicyradar/knowledge-graph/blob/3fae640ec502055116109510fe6287f3c20e7372/flows/wikibase_to_s3.py#L153

flows/predict.py
```diff
@@ -0,0 +1,77 @@
...
+    openrouter_api_key = SecretStr(
+        get_aws_ssm_param("OPENROUTER_API_KEY", aws_env=aws_env)
+    )
+    os.environ["OPENROUTER_API_KEY"] = openrouter_api_key.get_secret_value()
+
+    use_aws_profiles = os.environ.get("USE_AWS_PROFILES", "false").lower() == "true"
+    session = boto3.session.Session(
+        profile_name=aws_env.value if use_aws_profiles else None,
+        region_name=config.bucket_region,
+    )
```
THOR300, line 39: **thought**: This session will not be accessible outside of this function?

THOR300 (approved):

flows/predict.py
kdutia: thanks, this is useful to know how to do! there will be lots more work tidying up these flows, so i'll leave these for future

flows/predict.py
kdutia: it just needs the machine to auth with AWS to be able to access the models stored in S3 via W&B, so I thought this would do the job. it's the same as [calling `get_s3_session` in the script](https://github.com/climatepolicyradar/knowledge-graph/blob/92f09077c52e1167ed66b4aebc856f9d08c594c3/scripts/predict.py#L214)

PR #1113, opened 2026-04-13
fix: move retry handling to lowest levels & increase timeouts

olaughter:
This seems to be flakey in production suggesting its a heavier path, trying a more generous timeout

Additionally, three levels of abstraction in the trace, each level has retry logic. This simplifies by just putting it on the low level calls, this should improve the resilience because we are retrying per call rather than across a batch of calls, it also means we should have better interpretability, in that we had a lot of code trying to unwrap the error previously but can now just reraise instead.

### DIFF
--- a/knowledge_graph/wikibase.py
+++ b/knowledge_graph/wikibase.py
@@ -18,6 +18,7 @@ from pydantic import (
     ValidationError,
 )
 from tenacity import (
+    AsyncRetrying,
     before_sleep_log,
     retry,
     retry_if_exception_type,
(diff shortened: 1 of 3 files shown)


knowledge_graph/wikibase.py
```diff
@@ -231,17 +228,33 @@ async def _get_all_redirects(
...
+                        params={
+                            "action": "wbgetentities",
+                            "format": "json",
+                            "ids": "|".join(ids_to_fetch),
+                            "props": "info",
+                        },
+                        timeout=self.REDIRECT_REQUEST_TIMEOUT,
+                    )
+
+            if response is None:
```
THOR300, line 254: **suggestion**: Response should always have a value if the attempt passes due to `reraise=True`

THOR300 (approved):
Looking good, nice moving the retry to a more localised location rather than decorating the entire function. 

Compared against the example in the [docs](https://tenacity.readthedocs.io/en/latest/#retrying-code-block) and we're grand 🙌

knowledge_graph/wikibase.py
olaughter: It shouldnt get run but is also useful to have for regressions

PR #1114, opened 2026-04-13
Feature/pla 1342 create parallel inference pipeline for running chunks

olaughter:
Adds a deployment that will work for chunked-passage documents. This includes parameterising which document model is being used, which required an sdk bump.


flows/inference.py
```diff
@@ -570,14 +572,25 @@ def document_passages(
...
-            yield _stringify(text_block.text), text_block.text_block_id
+        if text_block.type not in BLOCKED_BLOCK_TYPES:  # pyright: ignore[reportAttributeAccessIssue]
+            yield _stringify(text_block.text), text_block.text_block_id  # pyright: ignore[reportAttributeAccessIssue]
+
+
+def document_passages_v2(
+    document: BaseParserOutputV2,
+) -> Generator[tuple[str, str], None, None]:
+    """Yield v2 text blocks as `(text, id)` tuples."""
+    text_blocks = document.get_text_blocks()
```
olaughter, line 583: This will be text, rather than serialised text which I'm going to double check is correct

olaughter: [Got a thumbs up on this](https://climate-policy-radar.slack.com/archives/C06JZKM6PGQ/p1776175643723599)

tests/flows/test_inference.py
```diff
@@ -724,6 +725,77 @@ async def test_run_classifier_inference_on_document(
...
+    classifier = MagicMock()
+    classifier.predict.return_value = [[]]
+
+    classifier_spec = ClassifierSpec(
+        wikibase_id=WikibaseID("Q788"),
+        classifier_id=ClassifierID("6vxrmcuf"),
+        wandb_registry_version="v7",
+    )
+    document = BaseParserOutputV2.model_validate(
+        {
```
kaparker: **thought:** I was wondering if this could be split out in a separate file to make it easier to update if any fields change, but I think it might end up being harder to maintain if it's not within the text file!

deployments.py
```diff
@@ -186,6 +192,7 @@ async def create_deployment(
         schedule=schedule,
         build=False,
         push=False,
+        parameters=deployment_parameters,
```
kaparker, line 195: **question:** will these be appended to the parameters set within the schedule?

kaparker (approved):
Looks good! Really clean addition and should be straightforward to clear up v1 when deprecating!

deployments.py
olaughter: Good question. I suspect they'll get overidden as they are described as defaults, I will make a point of checking this once deployed and may revisit if its confusing

tests/flows/test_inference.py
olaughter: Actually, thats a great point as there is already a very similar fixture for this that could just be reused instead. I'll switch over to it

PR #1124, opened 2026-04-20
fix: don't fail new sources by default

olaughter:
We previously were much more cautious about new corpora but this is now just causing disruption for no reason


kaparker (approved):
Looks good!

PR #1133, opened 2026-04-21
use entity-level metrics for BERT token classifier

kdutia:
eval loss and eval f1 [were both going up](https://wandb.ai/climatepolicyradar/Q32/runs/pgi68yk6?nw=nwuserkdutia) with a token-level f1 metric 🥴 

i'm not sure I fully understand why this is happening, but having an entity-level measure which better tracks the actual span metrics should help the training process to pick the correct model in the end

I chose this because it's what the [Huggingface token classification tutorial](https://huggingface.co/learn/llm-course/chapter7/2) uses

### DIFF
--- a/knowledge_graph/classifier/bert_token_classifier.py
+++ b/knowledge_graph/classifier/bert_token_classifier.py
@@ -6,13 +6,25 @@ import tempfile
 from collections.abc import Sequence
 from contextlib import contextmanager
 from datetime import datetime
+from typing import cast
 
 import numpy as np
 import pandas as pd
 import torch
 from datasets import Dataset
 from rich.logging import RichHandler
-from sklearn.metrics import accuracy_score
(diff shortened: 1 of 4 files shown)


juliesaigusaCPR (approved):
lgtm, but can we get a bit more in the description?

kdutia:
> lgtm, but can we get a bit more in the description?

@juliesaigusaCPR good point, i've added a link to the tutorial i got this idea from. is there any other detail you think is missing?

juliesaigusaCPR:
> eval loss and eval f1 [were both going up](https://wandb.ai/climatepolicyradar/Q32/runs/pgi68yk6?nw=nwuserkdutia) 🥴

was this following on from a previous PR?

kdutia:
@juliesaigusaCPR no, just what's on main at the moment

PR #689, opened 2025-09-16
Better bert based model training

harrisonpim:
We learned lots of lessons about model training in the [policy-mentions-classifier](https://github.com/climatepolicyradar/policy-mentions-classifier) project. As a result, the last training run was very fast, and produced a fairly solid passage-level classifier.

While this probably isn't the _perfect_ set of config/params/whatever for future BERT-based model training, it's almost definitely an improvement on what was here before, and it gives us a more modern template to start working from. Loads of improvements still to be made, loads of lessons still to incorporate from [targets classifier training](https://github.com/climatepolicyradar/classifier-training), etc, but I thought this was worth migrating now.

The highest priority in my mind is coming up with a method for reliably splitting a concept's labelled data into training vs eval sets... probably follow the targets methodology and use k-folds? not sure...

PR #1135, opened 2026-04-21
fix: set dataloader_num_workers to 0 for bert models

kdutia:
training modernBERT fails on coiled, with an exception ending in `RuntimeError: DataLoader worker (pid 135) exited unexpectedly with exit code 1. Details are lost due to multiprocessing. Rerunning with num_workers=0 may give better error trace.` - [run here](https://wandb.ai/climatepolicyradar/Q32/runs/bxuwis3h/logs?nw=nwuserkdutia)

#689 implemented this change as well as a bunch of others, but with no attached explanation. The [Huggingface docs](https://huggingface.co/docs/transformers/main_classes/trainer#transformers.Seq2SeqTrainingArguments.dataloader_num_workers) explain this value and state that its default is 0, which we're reverting to.

### DIFF
--- a/knowledge_graph/classifier/bert_based.py
+++ b/knowledge_graph/classifier/bert_based.py
@@ -611,7 +611,7 @@ class BertBasedClassifier(
                 load_best_model_at_end=True,
                 metric_for_best_model="eval_f1",
                 greater_is_better=True,
-                dataloader_num_workers=2,
+                dataloader_num_workers=0,
                 report_to=["wandb"] if enable_wandb else [],
                 disable_tqdm=True,
                 # W&B-specific settings when enabled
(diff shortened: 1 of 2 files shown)


juliesaigusaCPR:
does it work now with this change?

kdutia:
> does it work now with this change?

@juliesaigusaCPR unfortunately i have no way to know until this is deployed. running on staging [as the platform team suggested in Slack](https://climate-policy-radar.slack.com/archives/C06JZKM6PGQ/p1776778821377319?thread_ts=1776778448.664029&cid=C06JZKM6PGQ) does not work

juliesaigusaCPR:
sad! go forth

juliesaigusaCPR (approved):

PR #1151, opened 2026-04-29
Scripts for uploading geography concepts to wikibase

harrisonpim:
PRing a couple of scripts for archival purposes. See [the directory's README](https://github.com/climatepolicyradar/knowledge-graph/blob/094b53a29101265ce715363700fcd75d26ec107e/scripts/geography_concepts/README.md) for a bit more context 

I ran the second script this morning, which created a nice big hierarchy of concepts under [geography (Q2032)](https://climatepolicyradar.wikibase.cloud/wiki/Item:Q2032). The programmes team are now free to edit them, and downstream services can use them to create classifiers, or disambiguate free text data from the DIP to an authoritative source

One nice extra note - because all of these concepts are enriched with data from wikidata, we know their wikidata IDs - each new concept is tagged with a wikidata ID property eg https://climatepolicyradar.wikibase.cloud/wiki/Item:Q2039 ✨

closes SCI-722


knowledge_graph/wikibase.py
```diff
@@ -90,6 +90,7 @@ class WikibaseSession:
         "WIKIBASE_NEGATIVE_CONCEPT_PROPERTY_ID", "P11"
     )
     classifier_id_property_id = os.getenv("WIKIBASE_CLASSIFIER_ID_PROPERTY_ID", "P20")
+    wikidata_id_property_id = os.getenv("WIKIBASE_WIKIDATA_ID_PROPERTY_ID", "P13")
```
kdutia: will we ever want to override these values? i wonder whether they should just be defined as `Final[str]` values instead

knowledge_graph/wikibase.py
```diff
@@ -1357,6 +1358,218 @@ async def get_classifier_ids(
...
+        wait=wait_exponential_jitter(initial=RETRY_INITIAL_WAIT, max=RETRY_MAX_WAIT),
+        retry=retry_if_exception_type((HTTPStatusError, ReadTimeout)),
+    )
+    async def create_concept_async(
+        self,
+        concept: Concept,
+        subconcept_of: Optional[list[WikibaseID]] = None,
+        has_subconcept: Optional[list[WikibaseID]] = None,
+        related_to: Optional[list[WikibaseID]] = None,
+        wikidata_id: Optional[str] = None,
```
kdutia: nitpick: i've started using `type | None` where possible as we now use python versions that remove the need for `Optional`

knowledge_graph/wikibase.py
```diff
@@ -1357,6 +1358,218 @@ async def get_classifier_ids(
...
+        retry=retry_if_exception_type((HTTPStatusError, ReadTimeout)),
+    )
+    async def add_claim_async(
+        self,
+        entity_id: WikibaseID,
+        property_id: str,
+        target_id: WikibaseID,
+    ) -> None:
+        """
+        Add an item-valued claim to an existing Wikibase entity.
```
kdutia: the phrase "item-valued claim" feels a bit jargony to me – it could help to define it, or pick a clearer term?

scripts/geography_concepts/01_copy_geographies_from_wikidata.py
```diff
@@ -0,0 +1,302 @@
...
+        )
+    console.print(table)
+    write_output(geographies, "iso_3166-2", fmt)
+
+
+@app.command()
+def historical(
+    fmt: Annotated[Format, typer.Option("--format")] = Format.json,
+) -> None:
+    """Fetch ISO 3166-3 historical geography concepts from Wikidata."""
```
kdutia: could you define what 'historical' means here?

scripts/geography_concepts/README.md
```diff
@@ -0,0 +1,10 @@
...
+3. These concepts were then manually reviewed and their aliases were tweaked to most useful for searching and disambiguation. The refined sheet is in [google sheets, here](https://docs.google.com/spreadsheets/d/1f6YupXPK7DGcwXEnqCRoBHjumkg-pY0unj0S2_CRQPs/edit?usp=sharing). To get a local copy of this dataset for use in the second script, select the "ISO 3166-1 aliases_cleaned" sheet, and download it as a CSV via **File > Download > Comma Separated Values (.csv)**. Save it to the `scripts/geography_concepts/data` directory as `iso_3166-1_aliases_cleaned.csv`.
+4. [A second script](02_upload_geographies_to_wikibase.py) was written to upload the refined data to the concept store, preserving the hierarchy of the root 'geography' concept → region-level concepts → national-level concepts.
+
+This directory is maintained solely for archival purposes - the scripts here aren't maintained, and aren't intended to be re-run. Data in the concept store has likely changed since these scripts were first run, and thus the scripts here are not guaranteed to work.
```
kdutia, line 10: > the scripts here aren't maintained, and aren't intended to be re-run

It looks like the subnational path wasn't run, or at least these values didn't make it into the concept store. So, is this intended to be rerun? 

If so it might be worth calling out that there's this code path that hasn't been run yet, and might be run at some point – that way when the programmes team ask if possible a check of this readme should say so.

kdutia (commented):
exciting! some small comments about documentation, but the code looks great

knowledge_graph/wikibase.py
harrisonpim: nice! hadn't come across `Final` before. 

Taking this a step further, we might even be able to drop the ability to redefine these via env vars... When I originally wrote the `WikibaseSession` class I wanted to leave open the possibility of working with multiple wikibase instances - i was personally prototyping in another instance where the property IDs weren't the same - but at this point I'm pretty convinced that we'll be okay with our single instance and that these values can be treated as fixed 🤷‍♂️

kdutia: yeah, this is what I meant! sounds wise :)

kdutia (approved):

PR #1152, opened 2026-04-29
Effect of max passage tokens on training and inference time for BERT based classifier

juliesaigusaCPR:
We currently enforce a maximum length of 512 tokens per passage before truncation in our classifiers.  ModernBERT has a max length of 8192 tokens.  This experiment implements [dynamic padding](https://huggingface.co/docs/transformers/en/main_classes/data_collator#transformers.DataCollatorWithPadding) with increasing maximum lengths to observe the effects on local training and inference.

WandB project: https://wandb.ai/climatepolicyradar/max-length-experiment

The distribution of passage lengths (V2/chunked) in our full corpus (the longest passages are mainly tables and lists):
Screenshot 2026-05-21 at 16 32 21

Data: about 2000 passages were generated from LabelledPassages I already had locally.  For the largest 3 buckets (1025-2048, 2048-4096, 4097-8192), the number in each bucket was artificially inflated to 50.

I was not able to train any length above 512 with default batch size (16GB RAM).  I then cut batch size in half and used `group_by_length=True`.  [`group_by_length` ](https://huggingface.co/docs/transformers/main_classes/trainer#transformers.Seq2SeqTrainingArguments.train_sampling_strategy) "groups samples of roughly the same length together (to minimise padding and be more efficient)".  I found some people saying that this caused unexpected behaviour: https://discuss.huggingface.co/t/grouping-by-length-makes-training-loss-oscillate-and-makes-evaluation-loss-worse/82486.  Additionally, the time to start training can be reduced by passing a pre-computed list of lengths so that the passage lengths do not need to be computed to group them before training. https://discuss.huggingface.co/t/trainer-being-very-slow-to-init-training-setting-group-by-length-to-true/76754/2, https://github.com/huggingface/transformers/pull/10953

In my final iteration, maximum lengths of 512 and 1024 were quick, while 2048 took almost 11 hours for training and inference (6 hours for inference).  I wasn't able to go above 2048 with current settings locally.
Screenshot 2026-05-21 at 16 40 16

AI attribution:
Claude helped heavily with this code in terms of adapting it from the existing classifiers and generating the test data.


scripts/sample.py
```diff
@@ -122,6 +122,14 @@ def main(
...
         try:
             dataset = pd.read_feather(dataset_path)
+            if (
+                "corpus_type_name" in dataset.columns
+                and "document_metadata.corpus_type_name" not in dataset.columns
+            ):
+                dataset = dataset.rename(
+                    columns={"corpus_type_name": "document_metadata.corpus_type_name"}
+                )
+
```
kdutia: i think this is because you need to rebuild the dataset now it points at snowflake – it'd be worth doing this to get passages v2 in anyway!

kdutia (commented):
looks sensible to me! so, to summarise, your recommendation is that we try:

- halving batch size
- doubling max length to 1024
- enabling `group_by_length`?

I guess we try it, and run some experiments on classifiers we already have to see if it affects performance. 

can't approve this though, as it makes what i think is a breaking change to the sample script

kdutia (approved):

juliesaigusaCPR:
> looks sensible to me! so, to summarise, your recommendation is that we try:
> 
>     * halving batch size
> 
>     * doubling max length to 1024
> 
>     * enabling `group_by_length`?
> 
> 
> I guess we try it, and run some experiments on classifiers we already have to see if it affects performance.
> 
> can't approve this though, as it makes what i think is a breaking change to the sample script

clarification: the above, but in addition to implementing dynamic padding

PR #1157, opened 2026-05-11
Feature: Pla 1377 - Hanging Inference Jobs Response

THOR300:
Pull Request: 
---
The good news is I'm confident our inference jobs on GPU's aren't hanging, they're just taking a long time! Therefore, proposing the following. 

Why did I think they were hanging? 
- Host running for over 10 hours 
- No process logs for ~10 hours (just scheduler logs, many of which were connection errors)

## Proposal 

Batch Size: 
With the current batch size we have valid inference jobs that take 10+ hours to complete, this makes it quite difficult to differentiate between a hanging host and one that's genuinely still running. Therefore, proposing we drop the batch size to half the run time to more like 5 hrs. That way we can set a lower timeout. 

Time out: 
I reckon it's worth putting a timeout on the inference flow, if it's taking over a day we either need to speed up the process or something's gone wrong, this should be a very reserved estimate given jobs should take 5hr absolute max with this batch size. This also allows the topics pipeline to complete which would be stuck waiting for these gpu jobs to complete. It's also not the first time we've had very long running jobs (where we get notified by email). 

Logging: 
Then also adding logging so it's clearer from the logs that we're still running inference (we already log like this when writing result files to s3 as part of the inference step).

### DIFF
--- a/flows/inference.py
+++ b/flows/inference.py
@@ -84,7 +84,7 @@ BLOCKED_BLOCK_TYPES: Final[set[BlockType]] = {
 
 CLASSIFIER_CPU_CONCURRENCY_LIMIT: Final[PositiveInt] = 20
 CLASSIFIER_GPU_CONCURRENCY_LIMIT: Final[PositiveInt] = 10
-INFERENCE_BATCH_SIZE_DEFAULT: Final[PositiveInt] = 500
+INFERENCE_BATCH_SIZE_DEFAULT: Final[PositiveInt] = 200
 CLASSIFIER_PREDICT_BATCH_SIZE: Final[PositiveInt] = 32
 AWS_ENV: str = os.environ["AWS_ENV"]
 S3_BLOCK_RESULTS_CACHE: str = f"s3-bucket/cpr-{AWS_ENV}-prefect-results-cache"
--- a/flows/inference.py
+++ b/flows/inference.py
@@ -84,7 +84,7 @@ BLOCKED_BLOCK_TYPES: Final[set[BlockType]] = {
 
 CLASSIFIER_CPU_CONCURRENCY_LIMIT: Final[PositiveInt] = 20
 CLASSIFIER_GPU_CONCURRENCY_LIMIT: Final[PositiveInt] = 10
-INFERENCE_BATCH_SIZE_DEFAULT: Final[PositiveInt] = 1000
+INFERENCE_BATCH_SIZE_DEFAULT: Final[PositiveInt] = 500
 CLASSIFIER_PREDICT_BATCH_SIZE: Final[PositiveInt] = 32
 AWS_ENV: str = os.environ["AWS_ENV"]
 S3_BLOCK_RESULTS_CACHE: str = f"s3-bucket/cpr-{AWS_ENV}-prefect-results-cache"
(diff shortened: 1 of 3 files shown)


flows/inference.py
```diff
@@ -84,7 +84,7 @@
 
 CLASSIFIER_CPU_CONCURRENCY_LIMIT: Final[PositiveInt] = 20
 CLASSIFIER_GPU_CONCURRENCY_LIMIT: Final[PositiveInt] = 10
-INFERENCE_BATCH_SIZE_DEFAULT: Final[PositiveInt] = 1000
+INFERENCE_BATCH_SIZE_DEFAULT: Final[PositiveInt] = 500
```
olaughter:
```suggestion
INFERENCE_BATCH_SIZE_DEFAULT: Final[PositiveInt] = 200
```

kaparker (approved):
Looks good! 24 hours timeout makes sense!

flows/inference.py
```diff
@@ -783,6 +784,7 @@ async def run_classifier_inference_on_document(
     all_spans: list[list[Span]] = classifier.predict(
         all_text, batch_size=CLASSIFIER_PREDICT_BATCH_SIZE
     )
+    logger.info(f"Completed inference on {result.document_stem}")
```
olaughter: This is going to be so noisy if we log every time, suggest taking one of the following approaches:
- Bring the index number into scope and log every 50
- Add a check against a random number that means we log ~5% of the time (we actually do this in [indexing already](https://github.com/climatepolicyradar/knowledge-graph/blob/main/flows/index.py#L216-L235))

flows/inference.py
```diff
@@ -1131,6 +1133,7 @@ async def inference_batch_of_documents_cpu(
     result_storage=S3_BLOCK_RESULTS_CACHE,
     retries=INFERENCE_BATCH_RETRIES,
     retry_delay_seconds=INFERENCE_BATCH_RETRY_DELAY_S,
+    timeout_seconds=int(timedelta(hours=24).total_seconds()),
```
olaughter, line 1138: I recall this being flakey previously but that was along time ago now, have you tried testing this lately to make sure it behaves?

flows/inference.py
olaughter: Good idea, suggest going further

olaughter (approved):
Nice, this all makes sense. Some thoughts below fine tuning this

flows/inference.py
THOR300: Hmm honestly no I've not tested this based off of the assumption that prefect functionality works! Perhaps a wrong assumption after these lease renewals 😆

THOR300: it looks like there has previously been some issues with this feature, I expect getting prefect up to date should resolve any problems. 
- https://github.com/PrefectHQ/prefect/issues/10766

THOR300: I have done a very basic validation here which successfully timed out: 


```python 

import time

from prefect.flows import flow


@flow(
    timeout_seconds=3,  # Set the timeout to 3 seconds
)
def main():
    """
    Main function to run the flow.
    """
    # Simulate a long-running task
    time.sleep(5)
    print("Task completed.")


if __name__ == "__main__":
    main()
```

PR #1164, opened 2026-05-27
`ProbabilityCapableClassifier` always implements `predict_proba_batch`

kdutia:
As per the docstring in the new `abstractmethod`:

> Unlike `predict`, which only emits a `Span` (and thus a probability) for positive predictions, this returns a probability for *every* text, including negatives, independent of the threshold/argmax decision. This is needed for threshold calibration and for reporting probabilities on negative predictions.

This is so we can keep trusting mixins to implement the methods that are required to use the classifiers they describe downstream. E.g. we can use `if isinstance(classifier, ProbabilityCapableClassifier)` everywhere rather than `if hasattr(classifier, "predict_proba_batch")`.

### DIFF
--- a/knowledge_graph/classifier/classifier.py
+++ b/knowledge_graph/classifier/classifier.py
@@ -336,6 +336,18 @@ class ProbabilityCapableClassifier(ABC):
     overwrites probabilities – like the VotingClassifier.
     """
 
+    @abstractmethod
+    def predict_proba_batch(self, texts: Sequence[str]) -> list[float]:
+        """
+        Return per-text positive-class probability P(class=1) in [0, 1].
+
+        Unlike `predict`, which only emits a `Span` (and thus a probability) for
+        positive predictions, this returns a probability for *every* text, including
+        negatives, independent of the threshold/argmax decision. This is needed for
+        threshold calibration and for reporting probabilities on negative predictions.
+        """
+        ...
+
 
 @runtime_checkable
 class VariantEnabledClassifier(Protocol):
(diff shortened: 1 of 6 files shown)


knowledge_graph/classifier/classifier.py
```diff
@@ -336,6 +336,18 @@ class ProbabilityCapableClassifier(ABC):
...
+        """
+        Return per-text positive-class probability P(class=1) in [0, 1].
+
+        Unlike `predict`, which only emits a `Span` (and thus a probability) for
+        positive predictions, this returns a probability for *every* text, including
+        negatives, independent of the threshold/argmax decision. This is needed for
+        threshold calibration and for reporting probabilities on negative predictions.
+        """
+        ...
+
```
THOR300, line 350: **thought**: Seeing as this implements a new class method does this mean the existing models need retraining / pickling? I feel like we've hit a few occurrences before of changes to the python classifier class not being propagated to the pickled versions and I'm not entirely sure where we stand with that now?

THOR300 (approved):

knowledge_graph/classifier/classifier.py
kdutia: good question – as this isn't used in production inference paths i think we're in the clear though. also, [this PR](https://github.com/climatepolicyradar/knowledge-graph/pull/1078) should have fixed a lot of those issues for BERT classifiers 🤞

THOR300: Happy days!

PR #1165, opened 2026-05-27
fix: predict script outputs probabilities for negative predictions

kdutia:
- adds `include_prediction_probabilities` argument for `label_passages_with_classifier` method, which adds probabilities of negative predictions in labelled passage metadata
- predict and model evaluate generate prediction probabilities for negative examples
- test that probabilities of negatives are returned by `labelled_passages_to_dataframe`

you can see it work in the `runs["summary"].passages` table in [this W&B run](https://wandb.ai/climatepolicyradar/Q912/runs/xf4bjvqc?nw=nwuserkdutia)

image

### DIFF
--- a/knowledge_graph/classifier/bert_token_classifier.py
+++ b/knowledge_graph/classifier/bert_token_classifier.py
@@ -537,6 +537,51 @@ class BertTokenClassifier(
 
         return batch_labels, batch_probs, offset_mapping
 
+    def predict_proba_batch(self, texts: Sequence[str]) -> list[float]:
+        """
+        Return the per-text positive-class probability.
+
+        For this token-level model, the passage-level probability is the maximum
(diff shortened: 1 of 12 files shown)


ChloeBrennan-Card:
question (non-blocking): would `predict_proba_batch()` ever need batch size parameter as in `predict()`?

ChloeBrennan-Card (approved):
praise: although it's a bigger code change than you'd like, it was easy to follow :) Noted that for probability-capable classifiers inference is run twice per passage now.

kdutia:
> question (non-blocking): would predict_proba_batch() ever need batch size parameter as in predict()?

@ChloeBrennan-Card good question! i think yes and that this is something i missed 😳. but as this isn't prod code i'll merge as-is and we can add if it gets stuck. i haven't seen any signs of that yet though...

PR #1166, opened 2026-05-28
fix variance in predictions of LLM ensemble members

kdutia:
We use ensembles of LLM classifiers in our active learning workflows, but I'd noticed that an ensemble of LLM classifiers perturbed by only random seeds didn't seem to vary in their predictions enough.

So, an experiment with 5 different methods of creating ensembles: 

1. changing random seed only
2. fixed seed, varying temperature
3. changing random seed at temperature 0.7
4. prompt rewording
5. using different models

You can see the results in the CSV file on this PR. On Q32, our existing approach didn't actually lift F1 at all when 4/5 members of an ensemble agreed vs 1/5. The best approaches were 3 and 5.

This PR also implements method 3, as it was the easiest fix to apply here out of the two successful ones.

Also:

- adds some tests
- fixes a bug I found where a random seed of 0 would be silently coerced to 42
- bumps `griffe` in the lockfile to fix failing tests (it was unfrozen in #1034, but that PR didn't update it)


kdutia:
@AnneIsARealProgrammerNow fyi!

ChloeBrennan-Card (approved):
note: all makes sense!

PR #1167, opened 2026-06-01
feat(sci-836): convert sample.py to Prefect flow

ChloeBrennan-Card:
## Summary
- Adds `flows/sample.py` as a Prefect flow, wrapping the logic from `scripts/sample.py`
- Reads input dataset from S3 (`config.dataset_s3_bucket`) instead of local feather files
- Output unchanged: local JSONL file + W&B artifact 
- Adds `dataset_s3_bucket` field to `flows/config.py`
- Fixes type error in both `flows/sample.py` and `scripts/sample.py`: `list[list[Span]]` cast to `list[bool]` before DataFrame assignment

## Notes
- Local JSONL is only used for pushing to Argilla. The Argilla script will be converted to use W&B as part of [sci-837](https://linear.app/climate-policy-radar/issue/SCI-837/sample-eval-set-should-be-a-single-prefect-flow-accessible-via-the-ui)


flows/sample.py
```diff
@@ -0,0 +1,342 @@
...
+    - world bank region
+    - translated or untranslated
+    - type of document, eg CCLW, MCF, corporate disclosure
+
+    The sampled passages are saved to a local file and uploaded to W&B.
+
+    :param concept_override: List of concept property overrides in key=value format (e.g., description, labels)
+    :type concept_override: Optional[list[str]]
+    """
+    logger = get_logger()
```
kdutia: **suggestion:** could `scripts/sample.py` be refactored into functions which could then be imported into this flow, rather than copying the logic over to here? see [the pattern in predict](https://github.com/climatepolicyradar/knowledge-graph/blob/94acee37b195406b0a731293d90281e7b53f1403/flows/predict.py#L56)

ChloeBrennan-Card: **suggestion:** Thank you, I've updated accordingly.

scripts/sample.py
```diff
@@ -32,62 +32,26 @@
...
+    sample_size: int = 130,
+    min_negative_proportion: float = 0.1,
+    corpus_types_include: Optional[list[str]] = None,
+    corpus_types_exclude: Optional[list[str]] = None,
+    max_size_to_sample_from: int = 500_000,
+    max_negative_proportion: Optional[float] = None,
+    track_and_upload: bool = True,
+    concept_override: Optional[list[str]] = None,
+    wikibase_username: Optional[str] = None,
+    wikibase_password: Optional[str] = None,
```
kdutia: **polish:** if you wanted to, you could modernise these `Optional` types here too https://peps.python.org/pep-0604/

kdutia (commented):
nice work! this looks like it will work, but it's missing a deployment to run in ECS (AWS container workloads). Could you add a deployment for it? See `deployments.py`

flows/sample.py
```diff
@@ -0,0 +1,122 @@
...
+        config = await Config.create()
+
+    if track_and_upload:
+        wandb_api_key = SecretStr(get_aws_ssm_param("WANDB_API_KEY", aws_env=aws_env))
+        wandb.login(key=wandb_api_key.get_secret_value())
+
+    if dataset_name == "balanced":
+        dataset_filename = "sampled_dataset.feather"
+    elif dataset_name == "combined":
+        dataset_filename = "combined_dataset.feather"
```
kdutia: i think this will break, as the build_dataset flow stores the datasets in a subdir ([link](https://github.com/climatepolicyradar/knowledge-graph/blob/71cbebe51d37c5b13aa4bd2dbdcaa5a52183110e/flows/build_dataset_flow.py#L25-L26)). maybe import `COMBINED_S3_KEY` and `SAMPLED_S3_KEY` directly from that flow, now it's been merged to main?

ChloeBrennan-Card: Thank you for catching. Wondering whether the S3 load, w&b login and run_sampling could each be @tasks? Or is this not necessary here?

kdutia (approved):

PR #1169, opened 2026-06-01
flow to build vibe checker datasets

kdutia:
New flow replacing local vibe checker embeddings generation script, which runs monthly and uses the dataset feather file output by the build_dataset flow (#1170).

Also involved renaming some columns to catch the vibe checker up with the Snowflake data model (this would've usually been a separate PR but i'm still getting to grips with the vibe checker!).

### Breaking changes:

- page number is no longer displayed in the vibe checker. this is because it's too expensive to get from Snowflake, as `pages` is now stored as an array per text block


deployments.py
```diff
@@ -446,6 +447,20 @@ async def main() -> None:
...
+    await create_deployment(
+        flow=generate_vibe_checker_datasets,  # pyright: ignore[reportArgumentType]
+        description="Generate vibe-checker passage embeddings from the combined dataset in s3://cpr-kg-feather-files",
+        gpu=True,
+        flow_variables={
+            "cpu": 8,
+            "memory": "32 GiB",
+        },
+        env_schedules={
+            # 03:00 on the 1st of each month, before any 8am vibe-check run
```
abby11a, line 459: This is a really good point, it should definitely run before the vibe check. It has reminded me that build_dataset is currently scheduled at 8am too, but I can move that earlier to ensure the data is rebuilt before this flow runs.

abby11a (approved):

deployments.py
kdutia: don't worry about it, i believe there are [automations](https://github.com/climatepolicyradar/knowledge-graph/blob/4dc8c69e64cf1f843c10baaf778c08b9c4091c89/automations.py#L1) for us to be able to trigger this kind of thing but that's out of scope of us for now

PR #1170, opened 2026-06-01
feat: wrap build_dataset as a Prefect flow

abby11a:
Wraps `scripts/build_dataset.py` as a Prefect flow so it runs on a monthly schedule and writes to S3 instead of locally on a laptop.

[SCI-835](https://linear.app/climate-policy-radar/issue/SCI-835/convert-build-dataset-script-to-prefect-flow)

### What's changed

- Extracted `run_build_dataset()` from `scripts/build_dataset.py` so the flow can call it without touching the filesystem. The CLI (`just build-dataset`) still works exactly as before.
- New `flows/build_dataset_flow.py` — connects to Snowflake using the existing DbtBot key-pair SSM credentials, runs the queries in a thread pool, and uploads both feather files to `s3://cpr-kg-feather-files/build_dataset/`.
- Monthly deployment registered in `deployments.py` (8am on the 1st, production only).
- Added `dataset_s3_bucket` to `flows/config.py`.

### Tested

Ran the script locally - 9.3M rows combined, 10k balanced sample, columns and distribution look correct. I'll test the Prefect flow end-to-end once deployed to labs.


flows/build_dataset_flow.py
```diff
@@ -0,0 +1,112 @@
+"""
+Prefect flow wrapping scripts/build_dataset.py.
+
+Queries Snowflake for climate document passages, builds a combined (full corpus)
+and a balanced sampled dataset, and uploads both as feather files to S3 for use
+by the vibe checker and sampling flows.
+
+Runs on a monthly schedule (see deployments.py). DSs can pull the latest files
```
kdutia: polish: avoid abbreviations


```suggestion
Runs on a monthly schedule (see deployments.py). Data scientists can pull the latest files
```

flows/build_dataset_flow.py
```diff
@@ -0,0 +1,112 @@
...
+        get_aws_ssm_param(SNOWFLAKE_PRIVATE_KEY_SSM, aws_env=aws_env)
+    )
+
+    return config, snowflake_account, snowflake_user, snowflake_private_key
+
+
+@flow(
+    name="kg-build-dataset",
+    on_failure=[SlackNotify.message],  # pyright: ignore[reportUnknownMemberType]
+    on_crashed=[SlackNotify.message],  # pyright: ignore[reportUnknownMemberType]
```
kdutia: interesting... do we know where these notify? i'm definitely not in the channel, and we don't use Slack notifications in data science yet. keen to get this working if we can though!

flows/build_dataset_flow.py
```diff
@@ -0,0 +1,112 @@
...
+)
+async def build_dataset_flow(
+    n: int = 10000,
```
kdutia: polish: this could be worth renaming to be more descriptive, e.g. something like `sampled_dataset_target_num_rows`? Even though I know how this works, I found myself thinking it affected the combined dataset

also, a habit i'm trying to get into when putting large numbers into python:


```suggestion
    n: int = 10_000,
```

flows/build_dataset_flow.py
```diff
@@ -0,0 +1,112 @@
...
+    (
+        config,
+        snowflake_account,
+        snowflake_user,
+        snowflake_private_key,
+    ) = await _set_up_build_dataset_environment(config=config, aws_env=aws_env)
+
+    logger.info(f"Building dataset (n={n}, corpus_type={corpus_type})")
+
+    combined_df, sampled_df = await asyncio.to_thread(
```
kdutia: praise: smart! I had to ask claude why this was a good idea over prefect's `@task` and it agreed with you 🤝

scripts/build_dataset.py
```diff
@@ -194,26 +216,13 @@ def parse_geographies(x):
...
 
-    console.log(f"✅ Full dataset: {len(df_full):,} rows, presample: {len(df):,} rows")
-
-    console.log("🌍 Adding world bank region metadata")
     df_full["world_bank_region"] = df_full["document_metadata.geographies"].map(
         get_world_bank_region
     )
     df["world_bank_region"] = df["document_metadata.geographies"].map(
         get_world_bank_region
     )
```
kdutia, line 229: question: do we lose any ability to debug the code if we drop the logging statements here? (noting that if we kept them we'd want to switch them out to use `knowledge_graph.utils.get_logger`, to make them prefect-compatible)

flows/build_dataset_flow.py
```diff
@@ -0,0 +1,112 @@
...
+    logger.info(
+        f"Built datasets: combined={len(combined_df):,} rows, "
+        f"sampled={len(sampled_df):,} rows"
+    )
+
+    s3_client = cast(S3Client, get_s3_client(aws_env, config.bucket_region))
+
+    for df, key in [
+        (combined_df, COMBINED_S3_KEY),
+        (sampled_df, SAMPLED_S3_KEY),
```
kdutia, line 122: We lose the suffix of the filename that's added if `corpus_type` is set in the flow (so in S3), but not in the script. I think that could be less than ideal: if someone ran the flow manually with `corpus_type` set, they could overwrite the full dataset with a partial one.

For simplicity it could be better to remove the `corpus_type` option here rather than add the suffixes – but I'd do a quick double check that the dataset subsets aren't used downstream first

scripts/build_dataset.py
```diff
@@ -222,53 +231,87 @@ def parse_geographies(x):
...
-    dataset_path = processed_data_dir / f"sampled_dataset{corpus_suffix}.feather"
-    console.log("💾 Saving the dataset to feather")
-    df_balanced.to_feather(dataset_path)
-    console.log(f"✅ Saved the dataset to {dataset_path}")
+        if column in sampled_df.columns:
+            vc = sampled_df[column].value_counts()
+            console.log(f"  {column}:")
+            for val, cnt in vc.items():
+                console.log(f"    {val}: {cnt:,}")
+            console.log("")
```
kdutia, line 316: suggestion: it'd be neat if we could have these dataset metrics in the prefect flow too – they'd be super handy to debug dataset issues based on the last flow run

tests/scripts/test_build_dataset.py
```diff
@@ -0,0 +1,113 @@
...
+
+    for col in ["text_block.text", "document_id", "world_bank_region"]:
+        assert col in combined_df.columns
+
+
+def test_run_build_dataset_sampled_has_expected_columns(mock_snowflake_connection):
+    _, sampled_df = run_build_dataset(n=5)
+
+    for col in ["text_block.text", "document_id", "world_bank_region"]:
+        assert col in sampled_df.columns
```
kdutia: question: why this subset of columns? the the following ones are used in scripts/build_dataset.py, and in downstream metrics measurement. is it worth testing for them here?

```
"world_bank_region",
"document_metadata.corpus_type_name",
"translated",
```

kdutia (commented):
excellent work 👏 i've left a few comments for minor tweaks before we can merge and test this

flows/build_dataset_flow.py
abby11a: Good suggestion, I have now updated this

flows/build_dataset_flow.py
abby11a: This is a great catch, Claude was like do you want to use the existing slack notifier to notify on failure and I was like yeah I do! I should've checked which channel, turns out it goes to alerts-platform. So I'll remove for now, but we can eventually set one up for data science :)

flows/build_dataset_flow.py
abby11a: Yeah this is a good idea, I've now updated this

flows/build_dataset_flow.py
abby11a: Thanks! Although on second thought I feel like it should be using `@task` for better debugging haha, so I've now added `@task` and have kept the async thread logic. Please let me know if this isn't better though :)

scripts/build_dataset.py
abby11a: Yeah we do, great point. I've added them back using get_logger

flows/build_dataset_flow.py
abby11a: Great catch, scripts/sample.py and the vibe-checker both hardcode the fixed S3 filenames with no corpus-type suffix, so passing corpus_type to the flow would silently overwrite the full dataset with a filtered one. I've now removed it from the flow entirely :)

scripts/build_dataset.py
abby11a: Yeah I agree that would be nice, I've now added them to the flow :)

tests/scripts/test_build_dataset.py
abby11a: Yeah that would be a lot better to include all of them. I've now added them

flows/build_dataset_flow.py
kdutia: 🤷 let's give it a go and see!

kdutia (approved):
impressive first PR! excited to get this in :)

PR #1171, opened 2026-06-02
fix: adds geography csv to docker (SCI-835)

abby11a:
The `kg-build-dataset` flow crashed on first run because`knowledge_graph/geography.py` reads `geography-iso-3166.csv` at module import time, and the file wasn't present in the Docker image.

### What's changed

- Un-gitignored `data/raw/geography-iso-3166.csv`
- Added it to the Dockerfile so it's copied into the image

### Root cause

`build_dataset_flow.py` is the first flow to import from `scripts/`, which chains to `knowledge_graph.geography`, which does a top-level `pd.read_csv()` on import. Other flows never triggered this code path so it wasn't caught earlier.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -167,6 +167,7 @@ artifacts/*
 
 !data/raw/litigation-non-us.csv
 !data/raw/litigation-us.csv
+!data/raw/geography-iso-3166.csv
 
 tmp/
 scratch/
--- a/.dockerignore
+++ b/.dockerignore
@@ -51,6 +51,7 @@ htmlcov/
 
 # Local development
 data/
+!data/raw/geography-iso-3166.csv
 *.log
 .env
 .env.local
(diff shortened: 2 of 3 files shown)


kdutia (approved):
good catch! we now have geographies in the concept store, so i wonder whether we should update this...

.gitignore
```diff
@@ -167,6 +167,7 @@ artifacts/*
 
 !data/raw/litigation-non-us.csv
 !data/raw/litigation-us.csv
+!data/raw/geography-iso-3166.csv
```
kdutia, line 170: for my understanding, why do we need to un-gitignore this as well as removing it from `.dockerignore`?

abby11a: Docker builds from whatever files are in CI, so if its not in the git repo, docker can't see it. So geography-iso-3166.csv needs to be excluded from both .dockerignore and .gitignore. I had to double check with claude haha and it seems this is the case

PR #1176, opened 2026-06-03
fix (sci-836): separate out tasks for sample flow

ChloeBrennan-Card:
### Summary
- Fixes a `RuntimeError: Cannot run the event loop while another loop is running` bug that would cause the flow to fail in all environments (now know that Prefect always runs flows in an async loop, which conflicts with `async_bridge.py`'s `loop.run_until_complete()` call inside `run_sampling`)
- Wraps `run_sampling` in `asyncio.to_thread` inside a Prefect task, matching the pattern used in `build_dataset_flow.py`
- Extracts S3 dataset loading and W&B login into their own tasks for observability and consistency

### Tasks

- `load_dataset_from_s3`:  fetches the feather file from S3
- `login_to_wandb`: fetches W&B API key from SSM and logs in
- `run_sampling_task`: runs `scripts.sample.run_sampling` in a thread via `asyncio.to_thread`

### Tests
- Ran flow against files in `s3://cpr-kg-feather-files/build_dataset/` uploaded by `build_dataset_flow`
- Confirmed all three tasks appear in the Prefect Cloud UI


flows/sample.py
```diff
@@ -13,6 +14,107 @@
 from scripts.sample import run_sampling
 
 
+@task
```
kdutia, line 17: excellent - i am going to steal this for my fix PR 😆

kdutia (approved):

PR #1177, opened 2026-06-03
vibe checker dataset generation – logging fixes, code reuse

kdutia:
💐 A small bunch of fixes and improvements after the first run

- fixes logging by using a logger per task/flow
- adds logging into embeddings inference (a long running job). I tried ways of handing the `tqdm` output using a prefect-compatible logger, with no success
- logs the embedding model device, so we can check that it's actually running on a GPU 
- uses the dataset load method from `flows/sample.py`

### DIFF
--- a/flows/vibe_check_generate_datasets.py
+++ b/flows/vibe_check_generate_datasets.py
@@ -22,7 +22,7 @@ from flows.vibe_check import (
     get_bucket_name_from_ssm,
     push_object_bytes_to_s3,
 )
-from knowledge_graph.utils import get_logger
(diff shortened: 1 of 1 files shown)


ChloeBrennan-Card (approved):
**note:** I need to look at the wider vibe checker code, but everything here and the changes makes sense :)

kdutia:
> note: I need to look at the wider vibe checker code, but everything here and the changes makes sense :)

@ChloeBrennan-Card so do I! Once it's all working maybe i could do a run-through in a weekly sync?

PR #1179, opened 2026-06-03
feat: adds prefect flow for push new argilla dataset

abby11a:
## Description
Extracts the core Argilla push logic from the CLI script into a standalone `push_passages_to_argilla` function so it can be called from both the local script and a new Prefect flow.

The new `push_new_dataset` flow loads labelled passages from a W&B artifact and delegates to the extracted function, passing credentials explicitly from Config rather than relying on environment variables.

## Testing
Tests cover W&B login behaviour and that the flow passes the correct credentials to the push function.


flows/push_new_dataset.py
```diff
@@ -0,0 +1,62 @@
...
+    ],
+    workspace_name: Annotated[
+        str,
+        Field(description="The name of the existing workspace in Argilla"),
+    ] = "knowledge-graph",
+    limit: Annotated[
+        Optional[int],
+        Field(description="Limit the number of passages loaded to Argilla"),
+    ] = 130,
+    aws_env: AwsEnv = AwsEnv.production,
```
kdutia: question: do we need `aws_env` in here? it looks like it's not used below

scripts/argilla/push_new_dataset.py
```diff
@@ -5,12 +5,52 @@
 from knowledge_graph.identifiers import WikibaseID
 from knowledge_graph.labelled_passage import LabelledPassage
 from knowledge_graph.labelling import ArgillaSession
+from knowledge_graph.utils import get_logger
```
kdutia: as we don't want to encourage pushing datasets to argilla from the hardcoded directory, i wonder whether we delete this script entirely, and replace it with the flow... wdyt?

tests/flows/test_push_new_dataset.py
```diff
@@ -0,0 +1,104 @@
+from unittest.mock import AsyncMock, patch
```
kdutia: praise: a great set of tests!

kdutia (commented):
looking great overall 🌟 – just a couple of small questions

flows/push_new_dataset.py
abby11a: Great catch, I forgot to remove when copying from another flow. This has now been removed :)

scripts/argilla/push_new_dataset.py
abby11a: Yeah I think that's a good shout, I've now removed the script and moved it all to the flow

tests/flows/test_push_new_dataset.py
abby11a: Thanks :)

kdutia (approved):

PR #1181, opened 2026-06-04
fix: fetches WANDB_API_KEY from SSM in push_new_dataset

abby11a:
## Description
Fixes the following error for the push-new-dataset Prefect flow:
```
Encountered exception during execution: UsageError('No API key configured. Use `wandb login` to log in.')
```
Config.create() does not fetch WANDB_API_KEY from SSM, so config.wandb_api_key was always None and wandb.login() was never called, causing wandb.Api() to fail with "No API key configured".

Now it fetches the key directly from SSM (consistent with the sample flow), and warns rather than silently debug-log when the fetch fails so the issue is visible in Prefect logs.

### DIFF
--- a/flows/push_new_dataset.py
+++ b/flows/push_new_dataset.py
@@ -2,9 +2,10 @@ from typing import Annotated, Optional
 
 import wandb
 from prefect import flow
-from pydantic import Field
+from pydantic import Field, SecretStr
 
 from flows.config import Config
+from knowledge_graph.cloud import get_aws_ssm_param
 from knowledge_graph.identifiers import WikibaseID
 from knowledge_graph.labelling import ArgillaSession
 from knowledge_graph.utils import get_logger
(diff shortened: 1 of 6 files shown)


flows/push_new_dataset.py
```diff
@@ -44,8 +45,13 @@ async def push_new_dataset(
     if not config:
         config = await Config.create()
 
-    if config.wandb_api_key:
-        wandb.login(key=config.wandb_api_key.get_secret_value())
+    try:
+        wandb_api_key = SecretStr(
+            get_aws_ssm_param("WANDB_API_KEY", aws_env=config.aws_env)
+        )
```
kdutia: it looks like a previous oversight that this is not populated in `Config.create()` to me! what do you think about putting this loading in there? I imagine we might be able to delete a few lines of code from a few different flows after that.

abby11a: Yeah that's a good point, I've now added it to config instead. I've left the other flows that use w&b key as is (sample.py and train.py) for now but if you'd like me to convert them to use this there too please let me know :)

Edit: I just read your comment properly haha I'll convert the others too

kdutia (approved):
this is great - if you could check the code for anywhere else we could use this new config helper that'd be v much appreciated!

PR #1183, opened 2026-06-08
feat: adds sample and push flow

abby11a:
### Description
Adds flows/sample_and_push.py - a flow that runs sample followed by push_new_dataset in sequence, passing the newly created W&B artifact (climatepolicyradar/{wikibase_id}/labelled-passages:latest) to the push step automatically. If sampling fails, the push is skipped.

Exposes all parameters from sample (with the same defaults), plus workspace_name and limit from push_new_dataset. track_and_upload is forced to True since the artifact must exist in W&B for the push step to work.


flows/create_evaluation_dataset_in_argilla.py
```diff
@@ -0,0 +1,107 @@
...
+from pydantic import Field
+
+from flows.config import Config
+from flows.push_new_dataset import push_new_dataset
+from flows.sample import sample
+from knowledge_graph.cloud import AwsEnv
+from knowledge_graph.identifiers import WikibaseID
+
+
+@flow
```
kdutia, line 13: can we call this `create_evaluation_dataset_in_argilla` or something similar? I think `sample_and_push` might be too generic

flows/create_evaluation_dataset_in_argilla.py
```diff
@@ -0,0 +1,107 @@
...
+    ],
+    workspace_name: Annotated[
+        str,
+        Field(description="The name of the existing workspace in Argilla"),
+    ] = "knowledge-graph",
+    dataset_name: Annotated[
+        str,
+        Field(
+            description="Dataset to use",
+            json_schema_extra={"enum": ["balanced", "combined"]},
```
kdutia, line 29: interesting, is `Literal` or a direct `enum` not supported here because of prefect hashing?

flows/sample_and_push.py
```diff
@@ -0,0 +1,107 @@
...
+        corpus_types_exclude=corpus_types_exclude,
+        max_size_to_sample_from=max_size_to_sample_from,
+        max_negative_proportion=max_negative_proportion,
+        track_and_upload=True,
+        concept_override=concept_override,
+        aws_env=aws_env,
+        config=config,
+    )
+
+    wandb_artifact_path = f"climatepolicyradar/{wikibase_id}/labelled-passages:latest"
```
kdutia: we could end up with a race condition here if another person is doing any task that produces labelled passages. I was hoping, but didn't check, that we could get `@flow`-decorated functions to output variables i.e. the W&B artifact path in this case. do you know whether that's possible?

tests/flows/test_sample_and_push.py
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+
+@pytest.mark.asyncio
+async def test_sample_and_push_calls_sample_then_push(patched_subflows, test_config):
+    """Sample should be called before push_new_dataset."""
+    await sample_and_push(wikibase_id=WikibaseID("Q787"), config=test_config)
+
+    assert patched_subflows["sample"].called
+    assert patched_subflows["push"].called
```
kdutia: i think this asserts that both are called but not the order, as the test name and docstring imply?

kdutia (commented):
looking great - a few small comments

flows/create_evaluation_dataset_in_argilla.py
abby11a: Yeah good idea, I'll change this

flows/create_evaluation_dataset_in_argilla.py
abby11a: Using json_schema_extra={"enum": [...]} on a str parameter tells Prefect's UI to show an enum dropdown. This is how it currently is in the sample flow which seemed good, so copied over :)

flows/sample_and_push.py
abby11a: Good catch, it is possible I'll make small edits to the other files (i.e. sample) so it can actually return the path

tests/flows/test_sample_and_push.py
abby11a: Very true, that would be a better test. I will add

flows/sample.py
```diff
@@ -174,12 +174,15 @@ async def sample(
     ] = None,
     aws_env: AwsEnv = AwsEnv.production,
     config: Optional[Config] = None,
-) -> None:
+) -> str | None:
```
kdutia, line 177: thought (non-blocking): you might be able to use overload here to make the type checker aware that the output type should always be str if track_and_upload is True. I'm not sure if this plays nicely with the @task decorator and prefect more generally, though. See [example here](https://github.com/climatepolicyradar/knowledge-graph/blob/d36ce6679af2194209d35e9cc1c328eab0693b54/knowledge_graph/classifier/classifier.py#L89-L110)

kdutia (approved):

kdutia (approved):
great! one non-blocking comment

PR #1186, opened 2026-06-09
fix: adds artifact wait to sample and push flow

abby11a:
### Description
Fixes the following issue in create-evaluation-dataset-in-argilla. After sampling, accessing artifact.version immediately raised:
```
ArtifactNotLoggedError: 'Artifact.version' used prior to logging artifact or while in offline mode. Call Artifact.wait() before accessing logged artifact properties.
```

Fixed by calling logged_artifact.wait() after the W&B run context exits, blocking until the artifact is fully committed server-side before reading the version.

### DIFF
--- a/scripts/sample.py
+++ b/scripts/sample.py
@@ -277,6 +277,7 @@ def run_sampling(
             logger.info("✅ Labelled passages uploaded successfully")
 
     if track_and_upload and logged_artifact is not None:
+        logged_artifact.wait()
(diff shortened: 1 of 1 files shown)


kdutia (approved):

kdutia:
@abby11a having thought about this a bit more, I’m guessing this issue would’ve come up when running the flow locally? Maybe something to factor into our project milestone around thinking about local testing of prefect if so

PR #1189, opened 2026-06-09
manage feather files bucket in pulumi

kdutia:
`pulumi up` succeeds and adds the required cross-account access to the existing bucket. 

Not verified that access works for the vibe checker yet, but now this is in code this can be managed in a separate PR!

### DIFF
--- a/infra/README.md
+++ b/infra/README.md
@@ -1,5 +1,5 @@
 # AWS S3 bucket infra-as-code
 
-The code in this repo manages the bucket `s3://cpr-kg-feather-files` which exists in production and has cross-account access to labs.
+The code in this repo manages the bucket `s3://cpr-kg-feather-files` which exists in production and has cross-account access to labs. Labs cross-account access is needed as the feather files stored in this bucket are used by processes related to model training (AWS production) and the vibe checker (AWS labs).
 
 Vibe checker infra on AWS labs is managed separately in `vibe-checker/infra`.
--- a/infra/README.md
+++ b/infra/README.md
@@ -1,0 +1,5 @@
+# AWS S3 bucket infra-as-code
+
+The code in this repo manages the bucket `s3://cpr-kg-feather-files` which exists in production and has cross-account access to labs.
+
+Vibe checker infra on AWS labs is managed separately in `vibe-checker/infra`.
--- a/infra/__main__.py
+++ b/infra/__main__.py
@@ -1,0 +1,58 @@
+import pulumi
+import pulumi_aws as aws
+
+config = pulumi.Config()
+labs_aws_account_id = config.get("labs_aws_account_id")
+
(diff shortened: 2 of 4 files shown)


infra/README.md
```diff
@@ -0,0 +1,5 @@
+# AWS S3 bucket infra-as-code
+
+The code in this repo manages the bucket `s3://cpr-kg-feather-files` which exists in production and has cross-account access to labs.
```
katybaulch: nit: rationale on why we need the cross account perms might be useful here

infra/__main__.py
```diff
@@ -0,0 +1,58 @@
+import pulumi
+import pulumi_aws as aws
+
+config = pulumi.Config()
+labs_aws_account_id = config.get("labs_aws_account_id")
```
THOR300: **suggestion**: Use `config.require` here so `null` can't be passed in.

infra/__main__.py
```diff
@@ -0,0 +1,58 @@
...
+
+production_knowledge_graph_feather_files_bucket = aws.s3.Bucket(
+    "production-knowledge-graph-feather-files-bucket",
+    bucket="cpr-kg-feather-files",
+    grants=[
+        aws.s3.BucketGrantArgs(
+            id="0fedc730a2af259d90402b1197e87cf40c4014a20851f540cac4269c0156abb9",
```
THOR300, line 12: **thought**: We could add this to the pulumi config with a name indicating what / who this canonical user id is for?

THOR300 (commented):
Looking good! 

Further thoughts just whether you want a lifecycle policy to manage file number over time as well as bucket versioning perhaps to protect against overwriting existing data.

THOR300 (approved):

infra/__main__.py
katybaulch: if you're gonna do this, use `require_secret` instead to make sure it HAS to be secret and present

infra/__main__.py
kdutia: I don't understand this sorry! This was generated when I imported the already created bucket

infra/__main__.py
kdutia: done, thanks both!

infra/__main__.py
katybaulch: we could, but in the name of simplicity i'd say this is fine!

kdutia:
> Further thoughts just whether you want a lifecycle policy to manage file number over time as well as bucket versioning perhaps to protect against overwriting existing data.

@THOR300 i thought about it (after a brief intro to lifecycle policies from @katybaulch), but this bucket is likely to remain with a handful of files in it that can be generated in <30 seconds from snowflake - so it seemed like overkill a bit

katybaulch (approved):
nice work!! 🪄

abby11a (approved):

PR #1194, opened 2026-06-11
Feature/sci 842 manage custom prompts descriptions and definitions with yaml

ChloeBrennan-Card:
## Summary

For custom concept training, text is hardcoded in `scripts/custom_concept_training/qXXXX_train_llm.py`. This is the first step to managing custom prompts, descriptions and definitions with YAML, not code. `scripts/train.py` loads a Pydantic validated YAML config via `--from-yaml-config`. This is tested with Q1829 (finance flow), which uses a bespoke system prompt and guidelines.

## Changes

- `scripts/train.py` — minimal `CustomClassifierConfig` at the top of the file for now
-  `--from-yaml-config` option (mutually exclusive with `--wikibase-id`), resolved via `_resolve_training_inputs`
- `concept_overrides`: a definition/description that fits in the concept store (<= 2500 chars) raises a `ValidationError` with a "put it in Wikibase" message on load
- `scripts/custom_concept_training/configs/q1829.yaml` - bespoke system prompt and guidelines 
- `tests/test_train_from_yaml.py` - loads the yaml(s) end-to-end (offline, no W&B)

## Note on definition and description validation placement
I thought about having this as a separate config-validation script, but went with model validation so that an invalid config can't be constructed anywhere, and it surfaces as a standard Pydantic `ValidationError` with a clear message. Open to discussion on what is 'better' here!

## Next steps 

1. Extract `CustomClassifierConfig` into `knowledge_graph/custom_classifier_config.py`.
2. Add `SamplingConfig`
3. Add `--from-yaml-config` into `sample.py` and update `concept_overide: list[str]` in `run_sampling` (sample.py) to `concept_overides: dict | None` so dict can be used directly
4. Related-definition facilitation and Q32/Q911/Q912 configs
5. Update test and verify against the existing qXXXX runs, then retire those Python files
6. Scaling to classifiers beyond the existing Python ones; CI to enable writing a clean YAML with validation

11/06: working on 1-4, 15/06: will work on 5-6

## Testing
`uv run pytest tests/test_train_from_yaml.py -vv`   # all configs
`uv run pytest tests/test_train_from_yaml.py -k q1829 -vv `  # specific config


scripts/train.py
```diff
@@ -49,6 +54,108 @@
...
+            """Return the set override fields as a dict (drops unset/None)."""
+            return self.model_dump(exclude_none=True)
+
+    class LLMConfig(BaseModel):
+        """Model + prompt for the LLM-labelling stage."""
+
+        model_config = ConfigDict(extra="forbid")
+        model_name: str | None = (
+            None  # set in the YAML; if omitted then LLMClassifier's own default
+        )
```
kdutia: I think it could be better to enforce this is set. `LLMClassifier`'s default value could otherwise change the properties of the classifier as stored. Maybe we should also remove the default `model_name` in `LLMClassifier` - wdyt?

scripts/train.py
```diff
@@ -49,6 +54,108 @@
 
 app = typer.Typer()
 
+DESCRIPTION_WIKIBASE_LIMIT = int(os.getenv("WIKIBASE_DESCRIPTION_LIMIT", "2500"))
+DEFINITION_WIKIBASE_LIMIT = int(os.getenv("WIKIBASE_DEFINITION_LIMIT", "2500"))
```
kdutia: nitpick: maybe call these `_LENGTH_LIMIT`?

and, I think no need to allow override using env. I think that pattern is leftover elsewhere from when we were testing with different instances of local wikibase.

kdutia (commented):
a great start! two small comments

scripts/train.py
ChloeBrennan-Card: Done:)

scripts/train.py
ChloeBrennan-Card: I've enforced it and added some error handling. I've made a note to remove the default model_name in LLMClassifier as a separate pr as I want to understand were this will impact a bit better.

kdutia (approved):

PR #1196, opened 2026-06-15
feat(sci-842): Seperate CustomClassifierConfig, add SamplingConfig, related definitions handling

ChloeBrennan-Card:
## Summary

For custom concept training, text is hardcoded in `scripts/custom_concept_training/qXXXX_train_llm.py`. This is the second step to managing custom prompts, descriptions and definitions with YAML, not code. `scripts/train.py` and `scripts/sample.py` load a Pydantic validated YAML config via `--from-yaml-config`.

## Changes

- Extracted `CustomClassifierConfig` into `knowledge_graph/custom_classifier_config.py`
- Added `SamplingConfig`
- Added `--from-yaml-config` into `sample.py` and updated `concept_overide: list[str]` in `run_sampling` (sample.py) to `concept_overides: dict | None` so dict can be used directly (and in the Prefect flow)
- Related-definition facilitation and Q32/Q911/Q912 configs

## Next steps 

1. Update test and verify against the existing qXXXX runs, then retire those Python files
2. Scaling to classifiers beyond the existing Python ones; CI to enable writing a clean YAML with validation

## Testing
`uv run pytest tests/test_train_from_yaml.py -vv`   # all configs
`uv run pytest tests/test_train_from_yaml.py -k q1829 -vv `  # specific config

## Note

Q32 uses a different definition and description from the concept store but they're <2500. Checking with programmes team whether the concept store should be updated with these definitions.


knowledge_graph/custom_classifier_config.py
```diff
@@ -0,0 +1,163 @@
...
+                    "put it in Wikibase instead of overriding here."
+                )
+            return v
+
+        def as_overrides(self) -> dict[str, Any]:
+            """Return the set override fields as a dict (drops unset/None)."""
+            return self.model_dump(exclude_none=True)
+
+    class SamplingConfig(BaseModel):
+        """Maps to run_sampling() params (sample.py)."""
```
kdutia: nitpick: not using nested classes here would be more readable

(continues below, but the github UI wouldn't let me select all of it)

knowledge_graph/custom_classifier_config.py
```diff
@@ -0,0 +1,163 @@
...
+
+        @model_validator(mode="before")
+        @classmethod
+        def _require_model_name(cls, data: Any) -> Any:
+            """Error when model_name is missing/empty."""
+            if isinstance(data, dict) and not data.get("model_name"):
+                raise ValueError(
+                    "model_name is required — set llm.model_name in the config YAML"
+                )
+            return data
```
kdutia: I don't think you need this – `model_name` should raise an error if not defined as it's not got a default value

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "cSpell.words": ["pred", "wikibase"]
+}
```
kdutia: i'm not sure this should be pushed? what spellchecker do these settings apply to?

knowledge_graph/custom_classifier_config.py
```diff
@@ -0,0 +1,163 @@
...
+            """Exactly the run_sampling() sampling kwargs (drops None so run_sampling defaults win)."""
+            return self.model_dump(exclude_none=True)
+
+    class LLMClassifierConfig(BaseModel):
+        """Model + prompt ({slot}s filled from related_definitions)."""
+
+        model_name: str
+        system_prompt_template: str = DEFAULT_SYSTEM_PROMPT
+        labelling_guidelines: str | None = None
+        related_definitions: dict[str, WikibaseID] = Field(default_factory=dict)
```
kdutia: nitpick: it's not clear here what the `str` keys of related definitions refer to. renaming the variable descriptively (`thingone_to_thingtwo_mapping`) could be an easy way around this

knowledge_graph/custom_classifier_config.py
```diff
@@ -0,0 +1,163 @@
...
+            """Fill {slot} placeholders with the related concepts' definitions (passed by the caller)."""
+            if self.labelling_guidelines is None:
+                return None
+            if not self.related_definitions:
+                return self.labelling_guidelines
+            slots = {
+                name: definitions[wid] for name, wid in self.related_definitions.items()
+            }
+            return self.labelling_guidelines.format(**slots)
+
```
kdutia: would you ever expect for this method to be used interactively or in a script? If so, it might be helpful to describe the input and output arguments in the docstring. If not, then it could be better to make it a private method

scripts/sample.py
```diff
@@ -291,13 +292,12 @@ def run_sampling(
 @app.command()
 def main(
     wikibase_id: Annotated[
-        WikibaseID,
+        WikibaseID | None,
         typer.Option(
-            ...,
-            help="The Wikibase ID of the concept to sample passages for",
+            help="The Wikibase ID of the concept to sample passages for. Required unless --from-yaml-config is given.",
```
kdutia, line 297: I wonder whether you can use `typing.overload` (e.g. [here](https://github.com/climatepolicyradar/knowledge-graph/blob/e76c51d831591c83962bd9c3a603d739798852d4/knowledge_graph/ensemble/__init__.py#L68)) to enforce the constraint that either `wikibase_id` or `from_yaml_config` is present in the type checker

scripts/train.py
```diff
@@ -48,121 +44,37 @@
...
+
+def fetch_related_definitions(related: dict[str, WikibaseID]) -> dict[WikibaseID, str]:
+    """Fetch each related concept's definition from the store for {slot} interpolation."""
+    if not related:
+        return {}
+    session = WikibaseSession()
+    return {
+        wid: (session.get_concept(wid).definition or "")
+        for wid in set(related.values())
+    }
```
kdutia: nitpick: i don't think we gain anything from this being a separate method, as it's only used once and its implementation is only a couple of lines

tests/test_from_yaml.py
```diff
@@ -0,0 +1,133 @@
...
+    assert cfg.wikibase_id == WikibaseID(path.stem.upper())
+    stub = {wid: f"DEF::{wid}" for wid in cfg.llm.related_definitions.values()}
+    prompt = cfg.llm.to_classifier_kwargs(definitions=stub)["system_prompt_template"]
+    assert isinstance(prompt, LLMClassifierPrompt)
+    assert "{concept_description}" in prompt.system_prompt_template
+    for wid in cfg.llm.related_definitions.values():
+        assert f"DEF::{wid}" in (prompt.labelling_guidelines or "")
+
+
+def test_resolve_config_inputs_yaml_returns_cfg():
```
kdutia:
```suggestion
def test_resolve_config_inputs_yaml_returns_customclassifierconfig():
```

kdutia (commented):
looking great! left a bunch of comments, a minority of which are blocking

.vscode/settings.json
ChloeBrennan-Card: Ah yeah my bad I think this was added by mistake when I was trying to fix the bad merge

knowledge_graph/custom_classifier_config.py
ChloeBrennan-Card: Keen to keep the `related_definitions` naming so it's consistent with the yaml. Would an alias be appropriate here? E.g. `concept_name_placeholder: TypeAlias = str`. The labelling guidelines in the yaml contain {placeholder} slots and a map from each placeholder to a concept id. At train the script fetches each concept's current definition from the store and .format()s it into the guidelines.

knowledge_graph/custom_classifier_config.py
ChloeBrennan-Card: Removed

knowledge_graph/custom_classifier_config.py
ChloeBrennan-Card: Done

knowledge_graph/custom_classifier_config.py
ChloeBrennan-Card: Made private for now. Might come back to when implementing the CI to create a YAML.

scripts/train.py
ChloeBrennan-Card: Done

knowledge_graph/custom_classifier_config.py
kdutia: maybe just a single line comment is enough for now then – I originally thought the `str` was a description and then saw `name` below, hence the confusion!

separately: how do we use the name? would this be better pulled from the concept store?

scripts/sample.py
ChloeBrennan-Card: I don't think `typing.overload` facilitates Runtime enforcement from what I've read this afternoon? But maybe if the options were separately defined with overload and then validation logic goes in main?

kdutia: not runtime enforcement, just helps with viewing the code in an IDE which has type checking on. no big deal though :)

kdutia (approved):

knowledge_graph/custom_classifier_config.py
ChloeBrennan-Card: Yes, upon reflection, using just wikibase id for related definition is simpler and avoids potential user typos causing errors. Thank you!

scripts/sample.py
ChloeBrennan-Card: Ah I see! Thank you, I've had a go at using.

scripts/train.py
```diff
@@ -51,6 +51,27 @@
...
+    wikibase_id: None, from_yaml_config: Path
+) -> tuple[WikibaseID, CustomClassifierConfig]: ...
+
+
+# main() hands in two *optional* values, which matches neither strict overload above
+@overload
+def resolve_config_inputs(
+    wikibase_id: WikibaseID | None, from_yaml_config: Path | None
+) -> tuple[WikibaseID, CustomClassifierConfig | None]: ...
+
```
kdutia, line 73: nice!

kdutia (approved):

PR #1202, opened 2026-06-16
feat (sci 843): CI should enforce adding custom elements to concept store

ChloeBrennan-Card:
## Summary

Custom-classifier training config now lives in per-concept YAML files (SCI-842). This PR adds validation in CI and a first pass at an interactive generator so a user can create a YAML in CI.

The Pydantic model remains and is used at sample/train time too; the CI command invokes it and layers on filename/duplicate checks. Manual YAML authoring is fully supported whether or not the generator is used.

I used claude code to plan a skeleton interactive generator using Typer (for consistency with existing CI) and then heavily edited and iterated on it through testing.

## Limitations

The interactive generator is heavily biased towards the existing custom classifiers; someone with more engineering expertise could hopefully improve it's scalability and flexibility. 

## Changes

- `knowledge_graph/custom_classifier_config.py` - `SUPPORTED_LLM_MODELS` + a `model_name` validator
- `scripts/custom_concept_training/validate.py` (new)
- `scripts/custom_concept_training/configs_cli.py` (new) - Typer app with `validate` and `create`: live concept confirmation, bespoke definition and description with guardrails (enforces adding to concept store wherever possible - model does this too), model menu, sampling, related definitions, a guidelines template, optional bespoke system prompt, and self-validates the file it writes
- `scripts/custom_concept_training/templates/labelling_guidelines.md` (new) - a starter template based on Anne's work for climate justice custom classifiers
- `pyproject.toml` - `classifier-configs` entry point
- `scripts/scripts.just` - `validate-configs`, `validate-config path`, `validate-configs-live`, `new-classifier-config`
- `.pre-commit-config.yaml`
- `tests/test_from_yaml.py` - validation tests: committed configs pass, filename mismatch, duplicate concept, unsupported model, the guardrail on bespoke definition and description, and a mocked-session live check

## Testing

- `uv run classifier-configs validate` all committed configs (or use `just validate-config`)
- `uv run classifier-configs validate <config_file_path>` a specific config file (or use `just validate-config <config_file_path>`
- Ran `just new-classifier-config` end-to-end against a live concept, produced a YAML that re-validates clean. Errors raised when expected.
- `just validate-configs-live` 

## Questions

- `SUPPORTED_LLM_MODELS` is a hand-maintained list - could this be automated?
-  The just train wrapper forces --wikibase-id, so the mutually exclusive handling of --wikibase-id and --from-yaml-config will error. Should just be updated or is it ok to use uv run train?

## Next Steps

- Update concept store with definitions and descriptions in code
- Test end to end (what's the cheapest and lowest compute to test?) and remove custom classifier scripts
- Improve interactive generator for more flexible and scalable use
- Documentation and user guidance


knowledge_graph/custom_classifier_config.py
```diff
@@ -21,6 +21,13 @@
 DESCRIPTION_WIKIBASE_LENGTH_LIMIT = 2500
 DEFINITION_WIKIBASE_LENGTH_LIMIT = 2500
 
+# These are from the existing custom classifiers, need to check which are actually supported, whether keeping this up to date is feasible or can be automated or if this is necessary
+SUPPORTED_LLM_MODELS: set[str] = {
+    "openrouter:openai/gpt-5",
+    "openrouter:google/gemini-3.1-pro-preview",
+    "openrouter:google/gemini-3-flash-preview",
+}
+
```
kdutia: i wouldn't worry about this – openrouter has thousands (?) of models, so we can wait for the classifier to throw the appropriate error. it might be worth enforcing the `openrouter:` prefix in code, though

scripts/custom_concept_training/configs_cli.py
```diff
@@ -0,0 +1,280 @@
+import string
```
kdutia: I worry that this is a lot of logic to maintain, and I'm not clear on its advantage over a manual copy-paste or a script which runs a method which creates a new yaml file from the `BaseModel` with sensible defaults, e.g. something like 

``` py
def to_example_dict(model_class: type[BaseModel]) -> dict:
    result = {}
    for field_name, field_info in model_class.model_fields.items():
        default = field_info.default
        annotation = field_info.annotation

        if default is not PydanticUndefined:
            result[field_name] = default
        elif field_info.default_factory is not None:
            result[field_name] = field_info.default_factory()
        elif isinstance(annotation, type) and issubclass(annotation, BaseModel):
            result[field_name] = to_example_dict(annotation)  # recurse
        else:
            result[field_name] = get_example_value(field_name, field_info)

    return result
```

scripts/custom_concept_training/validate.py
```diff
@@ -0,0 +1,63 @@
...
+
+from knowledge_graph.custom_classifier_config import CustomClassifierConfig
+from knowledge_graph.identifiers import WikibaseID
+
+CONFIG_DIR = Path(__file__).parent / "configs"
+
+
+def expected_wikibase_id(path: Path) -> WikibaseID:
+    """Derive the WikibaseID a file must declare from its name (q32.yaml -> Q32)."""
+    return WikibaseID(path.stem.upper())
```
kdutia: i don't think we need this as a method, as it's a one line implementation and only used once

scripts/custom_concept_training/validate.py
```diff
@@ -0,0 +1,63 @@
...
+    errors: list[str] = []
+    try:
+        expected = expected_wikibase_id(path)
+    except ValueError:
+        return [f"filename {path.name!r} is not of the form q<number>.yaml"]
+    if cfg.wikibase_id != expected:
+        errors.append(
+            f"wikibase_id {cfg.wikibase_id} does not match filename-derived {expected}"
+        )
+    return errors
```
kdutia: what's the benefit of this over raising the errors and handling them downstream, rather than returning a list of strings?

.pre-commit-config.yaml
```diff
@@ -40,3 +40,11 @@ repos:
...
       - id: typos
         exclude: ^(tests/fixtures|tests/flows/fixtures|tests/local_vespa/test_documents|tests/__snapshots__|flows/classifier_specs|.*\.csv$)
+  - repo: local
+    hooks:
+      - id: validate-classifier-configs
+        name: validate classifier configs
+        entry: uv run classifier-configs validate --check-all-configs
+        language: system
+        pass_filenames: false
+        files: ^(scripts/custom_concept_training/configs/.*\.yaml|knowledge_graph/custom_classifier_config\.py)$
```
kdutia: nice attempt! i think we want to run this in CI though – [see here](https://github.com/climatepolicyradar/knowledge-graph/blob/ed1defd66d4140216623ab56701d465c2cecc9f3/.github/workflows/ci.yml#L1). and, ideally just run it when a file changes. adding too many steps to the pre-commit hooks slows down each commit for every developer, which is slow

kdutia (changes requested):
some initial comments - let me know if you'd like to huddle about this

.pre-commit-config.yaml
ChloeBrennan-Card: I've removed from the pre-commit and not yet added to CI. Would be helpful to chat about this one as I think I'm lacking some understanding. Also keen to know what you think the best way to run only when a file changes is. Chat requested:)

scripts/custom_concept_training/validate.py
ChloeBrennan-Card: My mistake, this was to see all failures at once across all configs before I realised validating a single file would be needed and I didn't make the changes correctly. I've updated and hopefully error raising is now more consistent.

scripts/custom_concept_training/configs_cli.py
ChloeBrennan-Card: I was worried too, but I was simultaneously overthinking trying to make it easy and intuitive for the programmes team to create. This is much cleaner; I've had a go at implementing this approach, thank you.

ChloeBrennan-Card: I decided to remove the labelling guidelines template as the comment block was making things look messy in the yaml, but I can add this back in if necessary.

scripts/custom_concept_training/configs_cli.py
```diff
@@ -50,230 +39,77 @@ def validate(
...
-
-    if not ok:
+            raise typer.BadParameter("Pass a config file path or --check-all-configs.")
+
+        if check_wikibase:
+            session = WikibaseSession()
+            for p in paths:
+                check_wikibase_ids(CustomClassifierConfig.from_yaml(p), session)
+    except ValueError as e:
+        console.print(f"[red]FAIL[/red] {e}")
```
kdutia: sorry for not catching this in the last PR, but it might help if we use `knowledge_graph/utils.py:get_logger` here to keep it consistent with elsewhere. that logger means we still get logging in prefect, should we decide to run functions from here on prefect

kdutia (commented):
looking a lot simpler - nice one! one small comment

scripts/custom_concept_training/configs_cli.py
ChloeBrennan-Card: That makes sense, thank you for catching and sorry I missed this.

.pre-commit-config.yaml
ChloeBrennan-Card: I've added a separate workflow, as this seemed the simplest implementation. I've used the logic from ci.yml.

kdutia (approved):

PR #1203, opened 2026-06-17
classifier training: thread `seed` through training script/flow; add threshold-independent metrics

kdutia:
- random seed can be set in train script and flow (see #1201)
- add passage-level PR-AUC and ROC-AUC to evaluation for probability capable classifiers
- add a test that these new metrics appear in the run summary


ChloeBrennan-Card (approved):
**question (non-blocking):** when should we use pytest vs unittest?

PR #1205, opened 2026-06-17
feat: replace predict/build_dataset scripts with an operations layer

abby11a:
### Context
A few of our operations (build-dataset, predict) currently keep three things in one file under `scripts/`: the domain logic, the Prefect flow, and a Typer CLI. 

That's caused some problems:
- **Imports point the wrong way.** `flows/` and `knowledge_graph/` both reach *up* into `scripts/` for the logic (e.g. `autollm.py` needs `evaluate`/`get_concept`). Lower layers depending on the top layer is backwards and risks an import cycle.
- **The logic isn't very testable.** It's tied to Prefect and Typer, so you can't unit-test it without the Prefect runtime
- **Duplicated, divergent logic.** For instance, both scripts talk to Snowflake, but each had its own connection code: build-dataset did proper key-pair auth, while predict just hardcoded the local `connection_name="local_dev"`, so predict couldn't authenticate in the cloud at all.

### What changed 
I've separated what an operation *does* from how it's *run*, with imports only pointing downward:

- New `knowledge_graph/operations/` layer for the pure, Prefect-free logic (`run_build_dataset`,  `run_prediction`, shared `connect_to_snowflake`).
- `flows/` is now thin - resolve credentials, call the operation, save the results.
- Removed the old `scripts/` files and their CLIs for these two operations.
- Added a `predict_documents` flow (+ on-demand deployment) so that path can run in the cloud.
- Along the way: parameterised the SQL (`corpus_type`, `document_ids`) instead of building query strings, and moved Snowflake connection + credentials into one shared place.
- Re-added thin `just` recipes (`just build-dataset`, `just predict`, `just predict-documents`) that call the operation functions directly, so local use stays easy. Documented in `flows/README.md`.

**Why this way?** 
The logic is now unit-testable without Prefect (see `tests/operations/`), the import direction is fixed, and there's a clear home for each kind of code. 

I also considered the following options:

- Put everything in flows/ (no separate layer). Simplest, but knowledge_graph/ already imports this logic, so moving it up would create an import cycle, and the logic would stay tied to Prefect.
- Make every operation a @ flow. Consistent, but it forces orchestration overhead onto code that doesn't need it, and a flow still can't be tested without Prefect.
- The operations layer (this PR). More work up front, but the only option that fixes the inversion and makes the logic testable.

### Testing
Unit tests for the new modules pass (fully mocked, no Snowflake/W&B/AWS needed). Imports and the new `just` recipes verified locally.

### What we need to do next
This PR establishes the pattern on build-dataset and predict. Still to migrate:
- sample: move its logic into operations/ (currently imported up by flows).
- train: scripts/train.py is ~933 lines mixing all three concerns; the careful split, best done last.
- inference.py: extract its pure per-document helpers into operations/infer.py.
- Enforce the boundary in CI with import-linter so the inversion can't silently return.


knowledge_graph/operations/snowflake.py
```diff
@@ -0,0 +1,78 @@
...
+    NoEncryption,
+    PrivateFormat,
+    load_pem_private_key,
+)
+
+from knowledge_graph.utils import get_logger
+
+SNOWFLAKE_WAREHOUSE = "PRODUCTION_DBT_WH"
+SNOWFLAKE_DATABASE = "PRODUCTION"
+SNOWFLAKE_SCHEMA = "PUBLISHED"
```
THOR300, line 27: **thought**: I can see these were carried over from the scripts, perhaps worth a follow up to remove the prod hard coding?

knowledge_graph/operations/snowflake.py
THOR300, line 1: praise: For lifting this out 🙌

THOR300 (approved):
Such a nice refactor, let's gooo! 🚀 

Splitting out changes into separate commits / stacked PR's would help the review from my perspective 🙏 !

knowledge_graph/operations/snowflake.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def connect_to_snowflake(
+    snowflake_user: str | None = None,
+    snowflake_private_key: str | None = None,
+    snowflake_account: str | None = None,
+):
```
THOR300, line 55: thought: A `match/case` statement could be nice here! 

```python
snowflake_config = (snowflake_user, snowflake_private_key, snowflake_account)

match snowflake_config:
    case (str(), str(), str()):
        ...

    case (None, None, None):
        ...

    # Any other combination means it's partially filled
    case _:
        raise ValueError(
            "Partial Snowflake credentials supplied: snowflake_user, "
            "snowflake_private_key, and snowflake_account must all be provided "
            "together for key-pair authentication, or all omitted to use the local "
            "config.toml connection."
        )

```

scripts/scripts.just
```diff
@@ -1,7 +1,11 @@
 
-# build a dataset of passages for sampling
+# local equivalent of build_dataset_flow (which writes to S3); writes to data/processed/
 build-dataset n="10000":
-    uv run python scripts/build_dataset.py --n {{n}}
+    uv run python -c "from knowledge_graph.operations.build_dataset import build_dataset_locally; build_dataset_locally(n={{n}})"
+
+# build a dataset filtered to a specific corpus type, written locally to data/processed/
+build-dataset-corpus corpus n="10000":
+    uv run python -c "from knowledge_graph.operations.build_dataset import build_dataset_locally; build_dataset_locally(n={{n}}, corpus_type='{{corpus}}')"
```
kdutia, line 8: what are your thoughts about whether we keep or remove these? they seem brittle, and i guess we could replace them with a `just get-datasets` command which pulls from s3?

scripts/scripts.just
```diff
@@ -78,9 +82,15 @@ deploy-classifiers ids aws_env +OPTS="":
...
+# just predict Q123 "climatepolicyradar/Q123/abc:v0" "climatepolicyradar/Q123/abc-labelled-passages:v0"
+# find instances of a concept in passages from a W&B artifact (runs the operation directly, no Prefect)
+predict id classifier_wandb_path labelled_passages_wandb_path:
+    uv run python -c "import asyncio; from knowledge_graph.operations.predict import run_prediction; asyncio.run(run_prediction(wikibase_id='{{id}}', classifier_wandb_path='{{classifier_wandb_path}}', labelled_passages_wandb_path='{{labelled_passages_wandb_path}}'))"
+
+# just predict-documents Q123 "climatepolicyradar/Q123/abc:v0" "CCLW.executive.1.1 CCLW.executive.2.2"
+# find concept instances in passages for Snowflake document IDs (uses local ~/.snowflake/config.toml)
+predict-documents id classifier_wandb_path document_ids:
+    uv run python -c "import asyncio; from knowledge_graph.operations.predict import load_passages_from_snowflake, run_prediction; passages = load_passages_from_snowflake('{{document_ids}}'.split()); asyncio.run(run_prediction(wikibase_id='{{id}}', classifier_wandb_path='{{classifier_wandb_path}}', input_passages=passages))"
```
kdutia, line 93: similarly, are these in here because you figured out we _can't_ run the prefect flows locally as-is?

kdutia (commented):
really helpful refactor! I have a couple of questions about whether we need the `uv run python -c ...` `just` commands or whether we could just locally run the flows

knowledge_graph/operations/snowflake.py
abby11a: Yeah, that's a great call! I'll definitely add that as a follow up

knowledge_graph/operations/snowflake.py
abby11a: Thank you 🙏

knowledge_graph/operations/snowflake.py
abby11a: Yeah great idea, now added :)

flows/snowflake.py
olaughter, line 1: suggestion: Wondering if this should this be combined with the snowflake file in `knowledge_graph/operations`?

flows/predict.py
```diff
@@ -50,13 +73,63 @@ async def predict_adhoc(
...
+        exclude_training_data=exclude_training_data,
+        prediction_threshold=prediction_threshold,
+        stop_after_n_positives=stop_after_n_positives,
+        restart_from_wandb_run=restart_from_wandb_run,
+        aws_env=aws_env,
+    )
+
+
+@flow()
+async def predict_documents(
```
olaughter: question: wondering on the naming here. Is `predict_document_passages` more accurate?

olaughter (approved):

scripts/scripts.just
abby11a: I'd be happy to remove them if you'd prefer. I mainly kept build-dataset as it seemed convenient for pulling the dataset locally rather than via S3, especially if you wanted a different sample size or corpus. But a get-datasets command is a great idea, I'll add this! :)

scripts/scripts.just
abby11a: Great question! The flows do run locally, these just scripts just give a thinner entry point that skips the Prefect @ flow wrapper and the extra prediction-environment setup, calling the operation directly. They still use W&B and AWS, so it's about convenience and preserving the quick workflow we had before, not because the flows can't run locally. Happy to drop them if they're not worth maintaining though!

flows/snowflake.py
abby11a: Good call, I think merging makes sense. Both files are really about getting a Snowflake connection, so keeping them in one place is more discoverable than hopping between flows/snowflake.py and operations/snowflake.py. I'd only split them to keep operations AWS-free, but boto3 is already a dependency so that's pretty marginal. I'll merge them!

flows/predict.py
abby11a: Yeah that would be clearer, I will rename

scripts/scripts.just
kdutia: given the flow is running on a schedule i think pulling from there will always be a decent option, and ensures we're all working with the same version of the data. but happy to leave this in too if you think you'd find it helpful! too early to tell

scripts/scripts.just
kdutia: i think calling the flows directly would help – i'm a bit wary of `just` commands as they tend to be untested so can drift pretty quickly from python implementation. @ChloeBrennan-Card flagged the same issue happening earlier to her, and wrote [this ticket](https://linear.app/climate-policy-radar/issue/SCI-879/should-we-delete-just-files)

flows/snowflake.py
olaughter: Nice!

flows/predict.py
olaughter: Awesome!

PR #1208, opened 2026-06-18
Implement changes from max length experiment (dynamic padding, etc)

juliesaigusaCPR:
Implementing changes recommended here: https://github.com/climatepolicyradar/knowledge-graph/pull/1152

Tested:
can we run inference on an existing model from W&B?
- model loaded using `evaluate.py` when setting `device="cpu"` in `new_classifier.move_model_to_device(device="cpu")` (`knowledge_graph/classifer/classifier.py`)

Same precision/recall/f1 when loading Q1829 before and after changes: (top is branch, bottom is main)
Screenshot 2026-06-23 at 14 17 33

To be tested:
Can we train a new model, that gets the same performance as before, with the new settings (https://linear.app/climate-policy-radar/issue/SCI-882/train-model-q1829-with-new-changes-and-compare-metrics)


knowledge_graph/classifier/bert_based.py
```diff
@@ -434,7 +437,7 @@ def fit(
         """
         Fine tune the base model using the labelled passages of the supplied concept.
 
-        The model is fine-tuned using several techniques for faster and more stable
+        The model is fine-tuned usg several techniques for faster and more stable
```
kdutia: typo
```suggestion
        The model is fine-tuned using several techniques for faster and more stable
```

kdutia (approved):

PR #1209, opened 2026-06-18
feat(sci-876): update train flow to use openrouter api secret in ssm

ChloeBrennan-Card:
## Summary

This PR updates the train flow to use openrouter api secret in SSM. It mirrors the way predict flow does. 

I also added a guarded fetch for the local CLI path that is gated on the `openrouter: model` prefix and skipped when OPENROUTER_API_KEY is already set. Not sure if this is required or if there is a better way of doing this.

## Changes

- `flows/train.py`: replicated the logic from `flows/predict.py`
- `scripts/train.py`: added a guarded fetch for local runs


scripts/train.py
```diff
@@ -863,6 +864,18 @@ async def run_training(
         classifier_kwargs=classifier_kwargs or {},
     )
 
+    # For local LLMCLassifier runs, remote runs fetches via flows/train/py
+    model_name = getattr(classifier, "model_name", None)
+    if (
+        isinstance(model_name, str)
+        and model_name.startswith("openrouter:")
+        and not os.environ.get("OPENROUTER_API_KEY")
+    ):
```
kdutia: i would say we may as well _always_ fetch the secret from SSM. i can't think of a valid situation where someone would have a key that's different from the one we all use, except for if someone not at CPR was using this code to train a classifier - which i think is unlikely!

ChloeBrennan-Card: Done, thank you! Also updated train tests with mock SSM.

flows/train.py
```diff
@@ -42,6 +43,11 @@ async def _set_up_training_environment(
     if config.wandb_api_key:
         wandb.login(key=config.wandb_api_key.get_secret_value())
 
+    openrouter_api_key = SecretStr(
+        get_aws_ssm_param("/OpenRouter/KGApiKey", aws_env=aws_env)
+    )
+    os.environ["OPENROUTER_API_KEY"] = openrouter_api_key.get_secret_value()
```
kdutia: could it be better to put this in `flows.config.Config` (line 31 above)? then, this PR could ensure all flows load the openrouter key from that config rather than manually. 

this would make it a bit more DRY to update future flows with this API key

ChloeBrennan-Card: Thank you!

kdutia (approved):

PR #1211, opened 2026-06-22
refactor: replace vibe checker specific methods with ones designated for reuse

kdutia:
a small 🧹 of the vibe checker, reusing some methods defined elsewhere – in an effort to make sure we use the tested surface of the code wherever we can

it looks like SCI-880 was a red herring – the failing `train-for-vibe-checker-labs` was actually removed in #1050, but removing a deployment from the CD script doesn't delete the existing deployment. I've now manually deleted that deployment in the Prefect UI in all environments

### DIFF
--- a/flows/vibe_check.py
+++ b/flows/vibe_check.py
@@ -20,13 +20,10 @@ import os
 import random
 from datetime import datetime
 from pathlib import Path
-from typing import Any, Optional
+from typing import Any, Optional, cast
 
-import boto3
 import numpy as np
 import pandas as pd
-import yaml
-from botocore.exceptions import ClientError
 from mypy_boto3_s3 import S3Client
 from prefect import flow, task
 from prefect.cache_policies import NO_CACHE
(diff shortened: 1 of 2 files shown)


flows/vibe_check.py
```diff
@@ -151,27 +131,6 @@ def load_passages_dataset(
     return dataset
 
 
-@task(retries=3, retry_delay_seconds=5)
```
juliesaigusaCPR, line 154: question: are we removing this retry behaviour [here](https://github.com/climatepolicyradar/knowledge-graph/pull/1211/changes#diff-2368f3f7cb28e4705f203b8b146ae5d86438dfd94feb637d9135176091cb63fdR147)?

juliesaigusaCPR (approved):
lgtm! just one non blocking question

flows/vibe_check.py
kdutia: yes, i think it wasn't needed as the task just loads a file from local storage. which is deterministic, so if it fails once it'll fail every time

PR #1213, opened 2026-06-23
refactor: extract evaluate & get_concept logic into knowledge_graph/o…

abby11a:
…perations

## Description

Extracts the reusable, Prefect/CLI-free logic from `scripts/evaluate.py` and `scripts/get_concept.py` into a new `knowledge_graph/operations/` package, and repoints all callers at the new home so there's a single canonical import path.
The scripts are reduced to thin Typer CLI wrappers.

## Why

`knowledge_graph.classifier.autollm` imported *up* into `scripts/` (`from scripts.evaluate import ...`, `from scripts.get_concept import ...`), which inverts the dependency direction - library code should not depend on `scripts/`. 

Moving the domain logic into `knowledge_graph/operations/` lets `autollm` and `scripts/train.py` import it directly, and establishes `operations/` as the canonical home for *what* an operation does, independent of
Prefect orchestration (`flows/`) or CLI entrypoints (`scripts/`).

## Changes

**New domain modules**
- `knowledge_graph/operations/evaluate.py`, `knowledge_graph/operations/get_concept.py` now own the logic (gold-standard construction, metrics, W&B charts, concept fetch, …).

**Scripts reduced to CLI wrappers**
- `scripts/evaluate.py` / `scripts/get_concept.py` keep only their Typer `app` + `main()` and import what they need from `operations/`. No re-export shim.
- `scripts/train.py` and `classifier/autollm.py` import from `operations/`.

**Callers repointed to `operations/`** (no more `from scripts.{evaluate,get_concept} import ...`)
- `static_sites/labelling_librarian/checks.py`
- `scripts/benchmarks/{embeddingclassifier,token_vs_sequence_bert,llm_ensemble_eval}.py`
- `scripts/active_learning/plot_ensemble_metrics.py`
- `scripts/classifier_experiments/batching_llm_classifier_using_sep_tag.py`
- removed a stale unused `import scripts.get_concept` from `scripts/deploy.py`

**Tests**
- `tests/test_evaluate.py` + syrupy snapshots moved to `tests/operations/`; patch  targets updated to `knowledge_graph.operations.*`.

## Behaviour

Pure refactor - no behaviour change. `just evaluate` / `just get-concept` work exactly as before (entrypoints and recipes unchanged). Verified:
- imports resolve with no cycle (`autollm` is imported lazily by `classifier/__init__.py`)
- no remaining references to `scripts.{evaluate,get_concept}` anywhere
- `ruff` passes on all touched files
- `tests/operations/test_evaluate.py` + `tests/test_train.py` green (46 passed)

## Known follow-ups (not in this PR)

This is the first slice of a broader `scripts/* → operations/` migration:

- **`sample` and `train`** still live in `scripts/` and are imported by flows (`train` by four flows), which violates `flows/* → scripts/*`. Next migration: move their logic into `knowledge_graph/operations/`.
- **`inference.py`** hasn't been split. Its per-document pure helpers (`document_passages`, `is_noop_document`, `_validate_spans`, …) still sit in `flows/inference.py` alongside orchestration. Follow-up: extract the pure helpers into
  `operations/infer.py` (+ `tests/operations/test_infer.py`); helpers that need `flows.*` types stay in `flows/`.
- **`analyse-classifier` just recipe** has no prediction step: `just predict` now needs a classifier + passages artifact path, which the recipe chain (only has an `id`) can't supply. Run `just predict` / `just predict-documents` separately for that step.


knowledge_graph/operations/evaluate.py
```diff
@@ -0,0 +1,660 @@
...
+
+Loads a classifier and a concept's gold-standard labelled passages, runs the classifier
+over them, and computes passage- and span-level performance metrics (grouped by equity
+strata) plus optional W&B charts and tables. This module knows nothing about Prefect or
+the CLI; `evaluate_classifier`, `create_gold_standard_labelled_passages` and
+`create_validation_predictions_dataframe` are imported directly by
+`knowledge_graph.classifier.autollm` (so this must live at/below `knowledge_graph` to
+avoid an import cycle), and the `scripts/evaluate.py` CLI wraps these for terminal use via
+`just evaluate`.
+"""
```
kdutia, line 9: nitpick: a lot of the contents of this seems to be generally about operations, rather than specific to this file. that could be better placed in an `operations/README.md`?

knowledge_graph/operations/evaluate.py
```diff
@@ -0,0 +1,660 @@
...
+from knowledge_graph.labelling import label_passages_with_classifier
+from knowledge_graph.metrics import (
+    ConfusionMatrix,
+    count_passage_level_metrics,
+    count_span_level_metrics,
+)
+from knowledge_graph.span import Span, group_overlapping_spans
+
+console = Console()
+logger = logging.getLogger(__name__)
```
kdutia: suggestion: should we use `knowledge_graph.utils.get_logger` here to ensure logs are picked up in prefect?

suggestion (2): we could make a `CLAUDE.md` file with instructions to always do that in this repo in a separate PR?

knowledge_graph/operations/evaluate.py
```diff
@@ -0,0 +1,660 @@
...
+    """Load a concept from local storage by its Wikibase ID."""
+    try:
+        concept = Concept.load(concept_dir / f"{wikibase_id}.json")
+        return concept
+    except FileNotFoundError as e:
+        raise typer.BadParameter(
+            f"Data for {wikibase_id} not found. \n"
+            "If you haven't already, you should run:\n"
+            f"  just get-concept {wikibase_id}\n"
+        ) from e
```
kdutia: nitpick: do we want to call this `load_concept_local`?

kdutia: this also could better sit in `operations/get_concept`?

knowledge_graph/operations/get_concept.py
```diff
@@ -0,0 +1,103 @@
...
+Get-concept operation: reusable, Prefect-free domain logic.
+
+Fetches a concept's metadata from Wikibase and its labelled passages from Argilla, saves
+the concept to disk, and returns it. This module knows nothing about Prefect or Typer; it
+is imported directly by `knowledge_graph.classifier.autollm` (so it must live at/below
+`knowledge_graph` to avoid an import cycle) and wrapped by the `scripts/get_concept.py`
+CLI for terminal use via `just get-concept`.
+"""
+
+from typing import Optional
```
kdutia: nitpick: we shouldn't need to import this any more https://peps.python.org/pep-0604/. this happens in a few places in this PR. maybe another one for a `CLAUDE.md` if you end up writing one?

scripts/get_concept.py
```diff
@@ -1,100 +1,29 @@
+"""
+CLI wrapper for the get-concept operation.
+
+The reusable logic (`get_concept_async`, `parse_wikibase_config`) lives in
+`knowledge_graph.operations.get_concept` and is imported directly by
+`knowledge_graph.classifier.autollm`. This module only adds the Typer command used by
+`just get-concept`.
```
kdutia: nitpick: i worry that this precise docstring might easily drift from implementation. wdyt?

kdutia (commented):
✨

knowledge_graph/operations/evaluate.py
abby11a: Yeah that would be clearer, I've moved this context to a md file here :)

knowledge_graph/operations/evaluate.py
abby11a: Very good point, I've now updated this. And yeah I agree a claude file with the best practices included would be a great idea.

knowledge_graph/operations/evaluate.py
abby11a: Yes and yes, also good points, a get_concepts file would be cleaner. I've now changed this

knowledge_graph/operations/get_concept.py
abby11a: Oh nice, cool I've now removed optional from these files and will add it to the claude md file

scripts/get_concept.py
abby11a: yeah thats a good point, I've now dropped the specific function names from the scripts/get_concept.py docstring so it won't drift.

knowledge_graph/operations/README.md
```diff
@@ -0,0 +1,31 @@
...
+## Modules
+
+See each module's docstring for what it does:
+
+- `build_dataset.py`
+- `evaluate.py`
+- `get_concept.py`
+- `predict.py`
+- `snowflake.py`
+
```
kdutia: nitpick (sorry, extra nitpicky): could this drift from the modules in this directory? is it needed for the reader to make sense of this directory?

knowledge_graph/operations/README.md
```diff
@@ -0,0 +1,31 @@
...
+## Conventions
+
+- Operations must live at or below `knowledge_graph/` so that library code
+  (e.g. `autollm`) can import them without an import cycle — library code must
+  never import *up* into `scripts/`.
+- Use `knowledge_graph.utils.get_logger` (not `logging.getLogger`) so logs are
+  picked up when an operation runs under Prefect.
+- Prefer PEP 604 unions (`X | None`) over `typing.Optional`.
```
kdutia: these are conventions for the whole codebase, not just the operations!

kdutia (approved):
nice :) approved pending your thoughts on a couple more nitpicky comments...

knowledge_graph/operations/README.md
abby11a: Agreed will remove

knowledge_graph/operations/README.md
abby11a: Very true. I'll remove for now and create a seperate pr with something more claude friendly with our conventions :)

PR #1215, opened 2026-06-30
fetch classifier ids as part of concept

kdutia:
This has been built to enable fetching "all concepts which have classifiers" in concept store MCP. See context in ticket: https://linear.app/climate-policy-radar/issue/SCI-902/allow-concept-store-mcp-to-access-classifier-id-and-not-to-be-confused

### DIFF
--- a/knowledge_graph/identifiers.py
+++ b/knowledge_graph/identifiers.py
@@ -186,3 +186,11 @@ class ConceptID(Identifier):
 
     This is intended for typing clarity rather than extending the Identifier class.
     """
+
+
+class StatementRank(Enum):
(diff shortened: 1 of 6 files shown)


knowledge_graph/identifiers.py
```diff
@@ -186,3 +186,11 @@ class ConceptID(Identifier):
 
     This is intended for typing clarity rather than extending the Identifier class.
     """
+
+
+class StatementRank(Enum):
```
kdutia, line 191: moved to avoid circular import between `concept` and `wikibase`

olaughter (approved):

PR #1214, opened 2026-06-25
refactor: extract train & sample logic into operations

abby11a:
## Description

Moves the reusable, Prefect-free logic for `train` and `sample` out of `scripts/` and into `knowledge_graph/operations/`, leaving the scripts as thin CLI wrappers. This completes the `scripts/* → operations/*` migration tracked in the `flows/README.md` known-follow-ups (resolves the `flows/* → scripts/*`import-direction violation). This is a pure code move + relocation; no behavioural change to training/sampling.

PR #1216, opened 2026-06-30
refactor: extracts promote/demote/metadata/deploy logic into flows

abby11a:
## Description

This wraps up the migration, apologies for the long review the changes were pretty dependent on eachother, I can break down into two if needed. 

`flows/classifiers_profiles.py` needed to promote, demote, and edit metadata on classifiers, but that logic only existed in `scripts/`. So the flow was doing `import scripts.promote` and calling `scripts.promote.main(...)` - a production
flow depending on CLI scripts, which drags in Typer and CLI-specific behaviour (`typer.BadParameter`, exit codes) where it doesn't belong.

This is a continuation of the scripts→operations/flows effort (#1214): push reusable logic down into importable modules and keep `scripts/` as thin entry points.

## What this PR does

Moves the reusable cores out of `scripts/` and into `flows/`, leaving the scripts as thin Typer wrappers:

| New core (`flows/`) | Scripts wrapper |
|---|---|
| `promote.py` → `run_promotion` | `scripts/promote.py` |
| `demote.py` → `run_demotion` | `scripts/demote.py` |
| `classifier_metadata.py` → `update` / `update_entire_env` | `scripts/classifier_metadata.py` |
| `update_classifier_spec.py` → `refresh_all_available_classifiers` | `scripts/update_classifier_spec.py` |
| `deploy.py` → `run_deploy_existing` / `run_deploy_new` | `scripts/deploy.py` |

- `classifiers_profiles` now imports the cores directly from `flows/`.
- `deploy` trains via `knowledge_graph.operations.train.run_training` instead of   `scripts.train.main`.
- Cores raise plain `ValueError`; CLI wrappers catch and re-raise as `typer.BadParameter`, so the cores carry no CLI dependency while CLI UX is unchanged.
- Logging standardised on `knowledge_graph.utils.get_logger` (Prefect-compatible).
- Tests moved to `tests/flows/` with mock patch targets repointed (`scripts.promote.main` → `flows.classifiers_profiles.run_promotion`, etc.).
- Minor: `Version | None` over `Optional`, dropped dead constants.

## Behaviour

No intended behaviour change - this is a move-and-rewrap. The local-training path in `deploy` resolves to the same `run_training` call the old `scripts.train.main` made for its defaults (no YAML config, `evaluate=True`,
`force=True`). Error messages and CLI exit codes are preserved.

## Risk / verification

- No `flows/` or `knowledge_graph/` import of `scripts.*` remains; no `knowledge_graph → flows` import.
- All cores + CLI wrappers + `classifiers_profiles` import cleanly (no cycles).
- `update-inference-classifiers` CLI rename (`refresh_all_available_classifiers` → `refresh`) is invisible — it's a single-command Typer app.
- Nothing catches `typer.BadParameter`, so the ValueError swap is safe.
- 94 tests pass across the affected files; ruff clean. (The one `__real_vespa` failure is a pre-existing integration test needing a live Vespa.)
- A real `promote`/`deploy` run against W&B + AWS can't be verified in CI (tests mock W&B)


flows/classifier_metadata.py
```diff
@@ -0,0 +1,172 @@
+"""
+Reusable logic for editing classifier metadata in W&B.
+
+These functions depend on `flows.classifier_specs` types, so they live in `flows/`
```
olaughter, line 4: This makes sense, and their output is primarily used in the flows as well so seems like good responsibility bereakdown 👍

scripts/classifier_metadata.py
```diff
@@ -131,102 +111,22 @@ def update(
...
-                else:
-                    artifact.metadata["classifiers_profiles"] = classifiers_profiles
-            else:
-                artifact.metadata.pop("classifiers_profiles", None)
-
-        artifact.save()
-
-    if update_specs:
-        refresh_all_available_classifiers([aws_env])
+    """Update the metadata for a classifier to determine behaviour at inference time."""
```
olaughter, line 114: Niiice, this looks much cleaner

olaughter (approved):
Nice work, this looks great!

PR #1223, opened 2026-07-01
delete just commands for model training

kdutia:
deletes all `just` commands related to model training, except for the ones that provide handy wrappers around python functions for local running. 

these are untested and undocumented, unlike the `typer` CLIs or flows – proven by the fact that a few of these don't actually have corresponding scripts!


ChloeBrennan-Card (approved):
Yay:)

PR #1225, opened 2026-07-14
fix: classifier_ids tuple expecting enum for both values

kaparker:
A new field has been added to the Concept model for classifier_ids. This is a tuple of enum (statement rank) and string (classifier_id).

In polars, the type of both fields in the tuple is inferred as an enum causing a failure in the sync_concepts flow trying to retrieve ClassifierID.value for a string.

This fix sets each value and prevents this issue.


kaparker:

- #1225 ◀

olaughter (approved):
Nice one!

PR #1226, opened 2026-07-15
Revert "fix: classifier_ids tuple expecting enum for both values (#1225)" and ignore new classifier_ids column

kaparker:
Revert "fix: classifier_ids tuple expecting enum for both values (#1225)"
and remove classifier_ids from being written to s3 to avoid schema conflicts between historical files and new files

This avoids historical data not having the classifier_ids field and new data going forward having this field. 
It is not used downstream in snowflake as the classifier_specs already contains this information and avoids incomplete data for historical classifiers.


kaparker:

- #1226 ◀

THOR300 (approved):

PR #1227, opened 2026-07-20
fix(enri-1505): make concept sync robust to field schema changes

mattwilliams00:
concepts_to_dataframe let Polars infer the schema from model_dump() output. A populated classifier_ids (list[tuple[StatementRank, ClassifierID]]) crashed inference and took down the whole sync flow.

- dump concepts with mode="json" so enums and custom types flatten to primitives Polars can handle
- cast every nullable/list column to an explicit dtype so empty and populated concepts produce one stable schema across Parquet files
- add regression tests: a roundtrip on a fully populated concept, and a cross-file Parquet schema-stability check

### DIFF
--- a/flows/sync_concepts.py
+++ b/flows/sync_concepts.py
@@ -61,19 +61,30 @@ def concepts_to_dataframe(concepts: list[Concept]) -> pl.DataFrame:
         ]
     )
 
-    # Cast Null columns to proper types for consistent schema Polars
-    # infers Null type when all values are None, but we want explicit
-    # types.
+    # Force explicit dtypes so empty/None columns don't infer as Null or
+    # List(Null), which differ from populated columns and break unioning the
+    # archive's Parquet files across runs.
     schema_casts = {
         "description": pl.Utf8,
         "definition": pl.Utf8,
+        "wikibase_id": pl.Utf8,
+        "wikibase_revision": pl.Int64,
+        "alternative_labels": pl.List(pl.Utf8),
+        "negative_labels": pl.List(pl.Utf8),
+        "subconcept_of": pl.List(pl.Utf8),
+        "has_subconcept": pl.List(pl.Utf8),
+        "related_concepts": pl.List(pl.Utf8),
+        "negative_concepts": pl.List(pl.Utf8),
         "recursive_subconcept_of": pl.List(pl.Utf8),
         "recursive_has_subconcept": pl.List(pl.Utf8),
     }
 
-    for col, dtype in schema_casts.items():
-        if col in df.columns and df[col].dtype == pl.Null:
-            df = df.with_columns(pl.col(col).cast(dtype))
+    if casts := [
+        pl.col(col).cast(dtype)
+        for col, dtype in schema_casts.items()
+        if col in df.columns and df[col].dtype != dtype
+    ]:
+        df = df.with_columns(casts)
 
     # Add sync timestamp to track when this version was synced
     df = df.with_columns(pl.lit(datetime.now(timezone.utc)).alias("synced_at"))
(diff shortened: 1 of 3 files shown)


tests/flows/test_sync_concepts.py
```diff
@@ -98,6 +98,22 @@ def test_dataframe_to_concepts__empty_dataframe():
...
+    A concept with every field populated survives the roundtrip (ENRI-1505).
+
+    classifier_ids is deliberately excluded from what is written to S3 (see
+    concepts_to_dataframe), so it is not expected to round-trip.
+    """
+    df = concepts_to_dataframe([fully_populated_concept])
+    recovered = dataframe_to_concepts(df)
+
+    assert len(recovered) == 1
+    assert recovered[0].model_dump(exclude={"classifier_ids"}) == (
```
kaparker, line 112: **question:** is the exclude needed here or should it already be excluded in the output of `dataframe_to_concepts` ?

tests/flows/test_sync_concepts.py
```diff
@@ -98,6 +98,22 @@ def test_dataframe_to_concepts__empty_dataframe():
     assert [] == dataframe_to_concepts(df)
 
 
+def test_dataframe_to_concepts__roundtrip_fully_populated(fully_populated_concept):
```
kaparker, line 101: **praise:** super helpful and clean test!

tests/flows/conftest.py
```diff
@@ -44,7 +44,12 @@
 from knowledge_graph.cloud import AwsEnv
 from knowledge_graph.concept import Concept
```
kaparker, line 45: **suggestion:** In the future, if another schema change is made to `Concept` are there any tests that could be added that will catch this and flag that further exclusions are needed to keep the output schema consistent?

flows/sync_concepts.py
```diff
@@ -52,28 +52,41 @@ def concepts_to_dataframe(concepts: list[Concept]) -> pl.DataFrame:
...
-    for col, dtype in schema_casts.items():
-        if col in df.columns and df[col].dtype == pl.Null:
-            df = df.with_columns(pl.col(col).cast(dtype))
+    casts = [
+        pl.col(col).cast(dtype)
+        for col, dtype in schema_casts.items()
+        if col in df.columns and df[col].dtype != dtype
+    ]
+    if casts:
+        df = df.with_columns(casts)
```
kaparker: **suggestion:** this is a nice addition to this code to keep the output schema consistent  to avoid any conflicts due to type changes in the future so might be worth keeping setting the schema structure in.

tests/flows/test_sync_concepts.py
mattwilliams00: Yeah I think it's needed, unfortunately! dataframe_to_concepts hands back full Concept objects, and since we don't write classifier_ids out, it just comes back as the default [] rather than being missing. So the recovered concept isn't equal to the original (which has it populated) unless we drop classifier_ids from both sides for the comparison. 

Please correct me if I'm wrong though! This is just how I thought it worked :)

flows/sync_concepts.py
mattwilliams00: I agree, do we want to do that in this ticket? Or keep this as a test related on and create a new one for the code change?

tests/flows/test_sync_concepts.py
mattwilliams00: Thanks!

tests/flows/conftest.py
mattwilliams00: Yea there currently isn't anything for that, a good guard would be a small test that introspects Concept.model_fields and fails if a field isn't accounted for in concepts_to_dataframe's handling. Do we want to incorporate that ?

tests/flows/test_sync_concepts.py
kaparker: ah yes, nice! this sounds great!!

tests/flows/conftest.py
kaparker: That sounds great, thanks! It will then force a change to `concepts_to_dataframe`  if the Concept model changes

kaparker (approved):
Looks great and super clean!

PR #1228, opened 2026-07-21
feat: remove indexing from sync concepts and classifiers

annaCPR:
- remove Vespa updates from concept and classifier sync
- tidy up unused functions and related tests


flows/classifiers_profiles.py
```diff
@@ -1685,35 +1346,13 @@ async def sync_classifiers_profiles(
...
-
-    vespa_errors = [unwrap_err(r) for r in vespa_results if isinstance(r, Err)]
+        logger.warning("Error creating and merging PR or syncing to s3")
 
-    # The default, assuming there were no Vespa successes
     event: Result[Event | None, Error] = Ok(None)
-    if any(map(is_ok, vespa_results)) and auto_train:
-        logger.info("found at least 1 Vespa success, emitting finished event")
+    if auto_train:
+        logger.info("Emitting finished event")
```
kaparker: **suggestion:** Looking at this again, this might need to have an additional condition too, or be nested in another block.

The event should only be emitted if there are changes to the classifier specs and the PR was merged successfully (in addition to auto_train). The logic before inherited those conditions as vespa would only run if there were no errors. This could potentially be added within an else after ` if is_err(cs_pr_results) or is_err(s3_result):`?

flows/classifiers_profiles.py
```diff
@@ -1274,97 +1045,6 @@ def emit_finished(
         )
 
 
-def maybe_allow_retiring(
```
kaparker, line 1277: **thought:** `maybe_allow_retiring and `concept_present_in_vespa` might actually be needed here, or adding some alternative.

These functions are checking the documents vespa instance (rather than the classifier profiles vespa) to see if this classifier exists as a topic on any documents and only will allow a classifier to have the profile of "retired" if it has already run through the topics pipeline before.

If we want to avoid using vespa, we could also check in s3 whether the labelled_passages for that wikibase id and classifier has any documents in the bucket - long term that might be the best option since that is the source of the topics data in the data lake.

flows/sync_concepts.py
```diff
@@ -1,36 +1,20 @@
-import json
```
kaparker, line 1: **praise:** this looks great! new version is super clean

flows/classifiers_profiles.py
annaCPR: Done

flows/classifiers_profiles.py
```diff
@@ -1045,6 +1045,89 @@ def emit_finished(
...
+
+def concept_has_results_in_S3(
+    wikibase_id: WikibaseID,
+    classifier_id: ClassifierID,
+    archive_path: S3Uri,
+) -> Result[bool, Error]:
+    """Check if a Concept<>Classifier has some results in S3."""
+    try:
+        s3 = boto3.client("s3")
+        results = s3.list_objects_v2(
```
kaparker, line 1112: **question:** if this prefix doesnt't exist at all (e.g. the wikibase_id folder doesn't exist in that path), will it throw an exception or just return 0?

flows/classifiers_profiles.py
```diff
@@ -1045,6 +1045,89 @@ def emit_finished(
...
+                    Error(
+                        msg="no results found in S3, so can't retire",
+                        metadata=op.classifiers_profile_mapping.model_dump(mode="json"),
+                    )
+                )
+            )
+            return False, wandb_results
+        case Err(e):
+            logger.info(
+                f"{op.classifiers_profile_mapping.wikibase_id}, {op.classifiers_profile_mapping.classifier_id} failed to be checked for in Vespa: {str(e)}"
```
kaparker: **fix:** typo for `Vespa` -> `S3`

flows/classifiers_profiles.py
```diff
@@ -1045,6 +1045,89 @@ def emit_finished(
...
+        archive_path=archive_path,
+    ):
+        case Ok(True):
+            logger.info(
+                f"{op.classifiers_profile_mapping.wikibase_id}, {op.classifiers_profile_mapping.classifier_id} has results in Vespa, and can be retired"
+            )
+            return True, wandb_results
+        case Ok(False):
+            logger.info(
+                f"{op.classifiers_profile_mapping.wikibase_id}, {op.classifiers_profile_mapping.classifier_id} has no results in Vespa, and can't be retired"
```
kaparker: **fix**: typo for `Vespa` -> `S3`

flows/classifiers_profiles.py
kaparker: Please ignore the above, I can see in the tests that it will return 0 results :)

flows/classifiers_profiles.py
kaparker: Looks great!

flows/classifiers_profiles.py
annaCPR: Well spotted!

flows/classifiers_profiles.py
```diff
@@ -1685,53 +1434,29 @@ async def sync_classifiers_profiles(
...
-        # the update will overwrite the doc if the field are unchanged
-        # or if there are changes to the fields but same doc_id
-        # or create a new document if a new doc_id is generated or it does not exist
-        async with vespa_search_adapter.client.asyncio(
-            connections=VESPA_CONNECTION_POOL_SIZE,
-            timeout=httpx.Timeout(VESPA_MAX_TIMEOUT_M.total_seconds()),
-        ) as vespa_connection_pool:
-            vespa_results = await update_vespa_with_classifiers_profiles(
-                updated_classifier_specs, vespa_connection_pool, upload_to_vespa
+        if auto_train:
```
kaparker: **thought:** my only concern here is if there are no changes to the classifiers, so no PR is opened and no updates to s3, will this still trigger since the result will be Ok(none)? This is the same logic as previous so I assume this is ok? Might be worth an extra test here to check though?

kaparker: Just to clarify, the event should only be emitted if there are changes to the classifier specs and those were merged successfully as in this case the topics pipeline (knowledge graph) needs rerunning.

tests/flows/test_sync_concepts.py
```diff
@@ -80,38 +73,47 @@ def test_concepts_to_dataframe(mock_concepts):
...
     df = concepts_to_dataframe([fully_populated_concept])
-    recovered = dataframe_to_concepts(df)
 
-    assert len(recovered) == 1
-    assert recovered[0].model_dump(exclude={"classifier_ids"}) == (
-        fully_populated_concept.model_dump(exclude={"classifier_ids"})
-    )
+    assert df.height == 1
+
+    # Values
```
kaparker, line 95: **suggestion:** the classifier_ids field should be in the fully_populated_concept but dropped during concepts_to_dataframe, might be worth an extra check here to confirm that column doesn't exist in the output df

kaparker (approved):
Looks great! A few minor comments that might need an extra test/check, otherwise all good!!

flows/classifiers_profiles.py
annaCPR: Good catch - the event was indeed emitted even if there were no classifier changes. I added an extra condition to prevent that plus the missing test coverage

kaparker (approved):

THOR300 (approved):
This clear down looks great @annaCPR, thanks a bunch for pushing this forwards! 🙌 

I'd say it would be worth deploying to staging via this gha workflow and running some integration tests there to build confidence in the change before merge: 

Image

PR #1231, opened 2026-07-27
fix: tolerate legacy schema drift when scanning the concept sync archive

THOR300:
## Summary
- `sync_concepts` was crashing every run because `pl.scan_parquet` enforces one identical schema across every file it globs in the archive, and one legacy file (`concepts_20260715_080922.parquet`) still has a `classifier_ids` column that later versions of `concepts_to_dataframe` stopped writing.
- Since the archive scan only ever needs the `id` column, extract it into `load_existing_ids()` and pass `extra_columns="ignore"` / `missing_columns="insert"` so schema drift between old and new archive files no longer fails the whole scan.
- This fixes the failure without deleting or rewriting any existing S3 archive data, and without losing sync history (see Slack thread on `sync-concepts/true-skylark`).
- Staging run [here](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/06a676e2-7da2-7333-8000-7d926d6820e4?preview=true&tab=logs).

## Test plan
- [x] Added `test_load_existing_ids__tolerates_extra_column_in_legacy_file`, which reproduces the incident (one archive file with an extra column) and asserts the scan succeeds. This is now removed as I didn't feel we needed it going forwards. 
- [x] `pytest tests/flows/test_sync_concepts.py -k "load_existing_ids or get_new_versions or concepts_to_dataframe or dataframe_to_concepts"` — 11 passed.
- [x] CI
- [x] Staging deployment and run 

### DIFF
--- a/flows/sync_concepts.py
+++ b/flows/sync_concepts.py
@@ -251,6 +251,32 @@ async def s3_prefix_has_objects(s3_uri: S3Uri, region: str, aws_env: AwsEnv) -> 
         return "Contents" in response and len(response["Contents"]) > 0
 
 
+def load_existing_ids(
+    parquet_pattern: str,
+    credential_provider: pl.CredentialProvider | None,
+    storage_options: dict[str, str] | None,
+) -> pl.LazyFrame:
+    """
+    Load concept IDs from the archive's Parquet files.
+
+    Only `id` is needed here, so schema drift between files written by
+    different versions of concepts_to_dataframe (e.g. a column that's since
+    been excluded, like classifier_ids) is tolerated rather than treated as
+    a fatal error.
+    """
+    return (
+        pl.scan_parquet(
+            parquet_pattern,
+            credential_provider=credential_provider,
+            storage_options=storage_options,
+            extra_columns="ignore",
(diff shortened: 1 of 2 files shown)


THOR300:
Update: reworked this to keep the `scan_parquet` fix inline in `sync_concepts` rather than splitting it into a separate `load_existing_ids` function. During development I wrote a quick regression test that reproduced the incident (an archive with one file containing an extra column) and confirmed `extra_columns="ignore"` / `missing_columns="insert"` fixes it — that test passed locally, giving confidence in the fix, but has been removed here since we don't need to carry it forward permanently.

kaparker (approved):
Great idea! Looks good!

kaparker:
In the description it mentions a new test but that isn't included in the PR?

THOR300:
> In the description it mentions a new test but that isn't included in the PR?

Ahh yes I'll remove that from the description, I added one to validate the change worked and then reverted it 👍

PR #1232, opened 2026-07-27
Feature/enri 1514 feat remove indexing from sync concepts and classifiers

annaCPR:
- fixes scheduled Prefect deployments which were still passing in a deprecated `upload_to_vespa` parameters and crashing


THOR300 (approved):
Let's goo!

kaparker (approved):

PR #1240, opened 2026-08-05
fixes for span metrics and merging

kdutia:
two fixes that Claude found while I was doing something else:

- misused `break`s meant that FP and FN spans were undercounted
- `Span.timestamp` is optional, and span merging crashed when timestamps were missing (fixing the robustness of the code rather than worrying about the root cause for now, as `Span` is used everywhere)


ChloeBrennan-Card (approved):
🔧 ✨

PR #1241, opened 2026-08-06
refactor: remove vespa indexing from topics pipeline

odrakes-cpr:
per title, remove vespa indexing from topics pipeline. 

this is part of the migration to make the processing pipeline more streamlined and push the indexing step to vespa further downstream


flows/topic_pipeline.py
odrakes-cpr: @THOR300 is the aggregation_result_raw here purely for for the vespa indexing step, trying to figure out what related functions to move

THOR300: Hey hey, looking at the code this is what I'm seeing as well! 

So yeh after aggregation it looks like we just need to run `create_topic_pipeline_summary_artifact` and then raise if it's an exception as is implemented on line 251. 

This would mean the match / case statement could be removed as well!

flows/topic_pipeline.py
```diff
@@ -186,62 +183,6 @@ async def topic_pipeline(
...
-        run_output_identifier=inference_run_output_identifier,
-        config=config,
-        n_documents_in_batch=aggregation_n_documents_in_batch,
-        n_batches=aggregation_n_batches,
-        return_state=True,
-        # Make sure we never wipe any concepts for users. Otherwise, it's allowed.
-        classifier_specs=classifier_specs
-        if config.aws_env != AwsEnv.production
-        else None,
-    )
```
THOR300, line 199: **change**: We definitely don't want to stop running the aggregation step! My prior comment was around the downstream use of the run output identifier not the job itelf.

flows/topic_pipeline.py
```diff
@@ -207,41 +184,14 @@ async def topic_pipeline(
...
-            # Partial success - extract AggregateResult from Fault data
-            if isinstance(aggregation_result_raw.data, AggregateResult):
-                aggregation_run_output_identifier = (
-                    aggregation_result_raw.data.run_output_identifier
-                )
-            else:
-                raise ValueError(
-                    f"Expected Fault.data to contain AggregateResult, got {type(aggregation_result_raw.data)}"
-                )
+            raise aggregation_result_raw
```
odrakes-cpr: I am not too comfortable with this swich as it feels redundant due to the line above and even line 253. @THOR300 but I am not sure I am understanding what we are trying to ascertain with this match statement.

THOR300: Hey hey! So this is what the match statement was doing 

```
match aggregation_result_raw:
         # We returned an Unknown Exception
        case Exception() if not isinstance(aggregation_result_raw, Fault):
            ...
        # We returned a Fault Exception (we do this if we had a mixture of successes and failures)
        case Fault():
            ...
        # We returned an aggregation result (we do this if we had total successes)
        case AggregateResult():
            ...
        # Handle an edge case (we don't expect this to get hit)
        case _:
            ...
```

THOR300: So we want to return an exception in the form of a Fault for a partial failure, but this means we have to distinguish between a raw Exception and a Fault.

THOR300: thought: Can we simplify to this now that we don't need the aggregation_result_raw downstream? 

```
         ...
        case _:
            raise ValueError(f"unexpected result {type(inference_result_raw)}")

    await create_topic_pipeline_summary_artifact(
        config=config,
        successful_document_stems=successful_document_stems,
    )

    aggregation_run: State = await aggregate(
        run_output_identifier=inference_run_output_identifier,
        config=config,
        n_documents_in_batch=aggregation_n_documents_in_batch,
        n_batches=aggregation_n_batches,
        return_state=True,
        # Make sure we never wipe any concepts for users. Otherwise, it's allowed.
        classifier_specs=classifier_specs
        if config.aws_env != AwsEnv.production
        else None,
    )
    await aggregation_run.result()
    logger.info("Aggregation completed!")

    logger.info("Checking after the fact whether inference was a partial failure.")
    if isinstance(inference_result_raw, Fault):
        raise inference_result_raw
    logger.info("Topic pipeline run completed successfully!")
```

THOR300: The Fault handling was to ensure we continued to indexing should aggregation have been a partial success, now that indexing is gone we probably don't need that.

THOR300: Then shifting up the artifact creation as that's not dependent upon aggregation.

odrakes-cpr: understood, but for partial results and Aggregate Result we had the `aggregation_run_output_identifier` which we used for the indexer. So whats the other path in this insstance.
```
 if isinstance(aggregation_result_raw.data, AggregateResult):
                aggregation_run_output_identifier = (
                    aggregation_result_raw.data.run_output_identifier
                )
            else:
                raise ValueError(
                    f"Expected Fault.data to contain AggregateResult, got {type(aggregation_result_raw.data)}"
                )
        case AggregateResult():
            aggregation_run_output_identifier = (
                aggregation_result_raw.run_output_identifier
            )

```

odrakes-cpr: ^^^ ignore the above, my github was being slow to reload

THOR300 (dismissed):
This is looking great! A couple actions below: 

One change required here in the `deployments.py ` to remove the param.
- https://github.com/climatepolicyradar/knowledge-graph/blob/b0f6ef15574fd992862709d14087b9c07ad6a52f/deployments.py#L365

We also shouldn't need to bring dependency changes like pre-commit and pyright into the repo to clear this code down so the changes in the `uv.lock` and `pyproject.toml` can go? 

Now we're not using the `index` flow this and all the related deployments, tests and documentation etc. can be cleaned up as well! Happy for this to be on a follow up PR so we can get this indexing job removed now but it defo needs to be done as part of the clean up 🙌

THOR300 (approved):
Looking cleaaaan, thanks so much for the testing iteration! 🚀

PR #1242, opened 2026-08-06
run LLM ensemble as eval set labeller experiment

kdutia:
Code to run experiment – see [Notion](https://app.notion.com/p/climatepolicyradar/RFC-Update-classifier-evaluation-set-labelling-3499109609a480978611c17b1e2bec69?source=copy_link#3b49109609a480c38306c4698a2ab237) for results including screenshots from the CLI.

Also: 

- adds span aggregators for ensembles: `knowledge_graph/ensemble/aggregation.py`
- adds negative predictive value (negative precision) and specificity (negative recall) to `ConfusionMatrix`

Sorry for the large PR, but this is experiment code, so no need to look too closely. Next week's implementation of this will be the one to pay closer attention to as it'll touch the pipelines we use for classifier dev!

### DIFF
--- a/knowledge_graph/ensemble/aggregation.py
+++ b/knowledge_graph/ensemble/aggregation.py
@@ -1,0 +1,81 @@
+"""
+Aggregation of an ensemble's per-classifier spans into a single set of spans.
+
+`Ensemble.predict` returns one list of spans per classifier, which is what the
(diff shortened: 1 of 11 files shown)


ChloeBrennan-Card (approved):
Didn't review everything line by line, but the approach makes sense:)

PR #1243, opened 2026-08-10
integrate llm ensemble labelling into validation set creation pipeline

kdutia:
Following #1242, adds the best performing ensemble as a provider of Argilla suggestions [see docs](https://docs.argilla.io/latest/reference/argilla/records/suggestions/). 

Recording of how this looks in the UI:

https://github.com/user-attachments/assets/f4e0b238-9189-4397-98fa-b6a318aa453b

This still needs to be tested with `LLMClassifier`s (blocked by [OpenRouter credits issue](https://linear.app/climate-policy-radar/issue/FUS-225/get-openrouter-access-unblocked)) and the deployed Prefect flow, but these can be done after this PR is merged.


knowledge_graph/labelling.py
```diff
@@ -640,14 +786,42 @@ def add_labelled_passages(
...
 
+            suggestions = None
+            if suggested_spans_and_score is not None:
+                spans, score = suggested_spans_and_score[idx]
+                span_values = self._spans_to_argilla_value(spans, wikibase_id)
+                # an empty suggestion is still attached, so that a passage the model
+                # judged negative is distinguishable from one it never saw
+                suggestions = [
+                    Suggestion(
+                        question_name="entities",
```
ChloeBrennan-Card: **question(non-blocking)**: Took me a while to work out what this refers to, is this correct? Every concept dataset has a span ‘question’ to highlight mentions of the concept in the text and each human response and each model suggestion is tied back to the ‘question’ by its name? Appreciate this was used pre this pr, but is it worth a comment explaining or is it more obvious than I thought?

ChloeBrennan-Card (approved):
One non-blocking comment, otherwise 🧑‍🍳  🤌

knowledge_graph/labelling.py
kdutia: good catch – i was also confused by this at first! 

I've added a clarifying comment and turned this magic value into a variable – [see commit](https://github.com/climatepolicyradar/knowledge-graph/pull/1243/changes/ac3d8a13976c6f75926968c0d044298dd295b5ed)

PR #1245, opened 2026-08-11
fix vibechecker inference for recursive negative labels

kdutia:
Vibe checker inference was stuck in an infinite loop ([see example flow](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/06a744df-474c-7860-8000-b3eeb2caca3f?preview=true&tab=logs&selected_id=019fea7b-7173-78ad-8915-c97f85a4ed85&selected_kind=task-run)), because recursive negative labels had been added (6th August). 

This PR fixes that by **not fetching the negative labels of a negative label** – otherwise we the process continues forever!

Also:

- adds a timeout and slack alert on failure of any concept's inference to the Prefect flow (to help us catch issues like this before the programmes team in future)
- gives the subtasks of the flow names that incorporate their `wikibase_id` for easier debugging

### DIFF
--- a/flows/vibe_check.py
+++ b/flows/vibe_check.py
@@ -148,7 +148,12 @@ def load_embeddings_metadata(
     return json.load(io.BytesIO(bytes_from_s3))
 
 
-@task(retries=2, retry_delay_seconds=10, cache_policy=NO_CACHE)
+@task(
+    retries=2,
+    retry_delay_seconds=10,
+    cache_policy=NO_CACHE,
+    task_run_name="process_single_concept-{wikibase_id}",
+)
 async def process_single_concept(
     wikibase_id: WikibaseID,
     passages_dataset: pd.DataFrame,
--- a/flows/vibe_check.py
+++ b/flows/vibe_check.py
@@ -50,6 +50,11 @@ from knowledge_graph.wikibase import WikibaseConfig, WikibaseSession
 
 aws_region = os.getenv("AWS_REGION", "eu-west-1")
 
+# Bounds to stop a hanging concept from holding a task runner slot, and the whole flow
+# run, forever. A healthy concept takes minutes at most.
+CONCEPT_TIMEOUT_SECONDS = 30 * 60
+FLOW_TIMEOUT_SECONDS = 6 * 60 * 60
+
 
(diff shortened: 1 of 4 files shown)


flows/vibe_check.py
```diff
@@ -148,7 +154,13 @@ def load_embeddings_metadata(
     return json.load(io.BytesIO(bytes_from_s3))
 
 
-@task(retries=2, retry_delay_seconds=10, cache_policy=NO_CACHE)
+@task(
+    retries=2,
+    retry_delay_seconds=10,
+    cache_policy=NO_CACHE,
+    task_run_name="process_single_concept-{wikibase_id}",
```
jamesgorrie, line 161: ✨  Nice, didn't know you could use string interpolation like this

knowledge_graph/wikibase.py
```diff
@@ -402,24 +409,29 @@ async def _fetch_concept_async(
...
                 logger.warning("❌ %s", str(e))
+                return None
             except ValidationError as e:
                 logger.warning("❌ Failed to validate concept %s: %s", wikibase_id, e)
-            return None
+                return None
+
+        # Negative concepts are fetched after releasing the semaphore, otherwise
+        # one fetch consumes 2/3 semaphore permits at a given time.
+        if incorporate_negative_concepts:
```
jamesgorrie, line 431: q: for my clarity here, this is only `True` when called from `_incorporate_negative_concepts` to avoid e.g. tax => tax advantage => tax ➿ ?

jamesgorrie (approved):
Ah graph recursion, what a jolly! Nice catch.
Nice solution to avoid it happening again!

knowledge_graph/wikibase.py
kdutia: yes, exactly!

flows/vibe_check.py
kdutia: thanks, me neither until now – found it poking around manually whilst Claude was Clauding some other stuff, and I prefer it to the convention used elsewhere in the repo of defining a specific task-naming function I think

PR #1247, opened 2026-08-11
fix vibechecker W&B parallelism; enable per-concept retries

kdutia:
two fixes for a [second failure](https://app.prefect.cloud/account/4b1558a0-3c61-4849-8b18-3e97e0516d78/workspace/1753b4f0-6221-4f6a-9233-b146518b4545/runs/flow-run/06a7b277-d1b6-7b19-8000-65374160212c?preview=true&tab=logs&left-panel-tab=run&selected_id=019ff115-32d9-7a2f-98a7-38ad04dc474c&selected_kind=task-run) (after #1245):

1. allow the `process_single_concept` task to fail, so it can be retried
2. apply a `threading.Lock()` to the Weights and Biases `Run`, otherwise several concepts try to use the same run (unsurprisingly Claude found this one, not me!)

details on 2:

> Both come from querying the W&B API about the actual failed run, not from a reproduction:
> 
> 1. Project Q2420 has no run on 2026-08-11 at all — only 08-05, 08-06 and 08-10. Yet the flow logs show Q2420 went all the way through train_classifier that day, and wandb.init() is called unconditionally there when track_and_upload=True. So Q2420's init() created no run. It must have been handed an existing one.
> 2. Project Q2421 contains a collection named 2hz23cj2, with versions=[] — and 2hz23cj2 is Q2420's classifier id, per the log line Saved KeywordClassifier("energy supply mitigation") to /app/data/processed/classifiers/Q2420/2hz23cj2/.... The only code that creates that collection is create_and_link_model_artifact calling run.log_artifact(name=classifier.id), and an artifact lands in the project owned by the run that logs it. So the run Q2420's thread was holding belonged to project Q2421. The empty version list is the commit that never landed; Q2420/2hz23cj2:latest doesn't exist.
> 
> Those two can only both be true if Q2420's thread was operating on Q2421's run. That's the claim, and it's an observation about the failed run itself.

It seems like it's happened more times than just this, but we've been saved by classifiers auto-retraining every day if they don't find their ID already in W&B – i.e. if W&B logging fails, the classifier is just trained again the next day

### DIFF
--- a/flows/vibe_check.py
+++ b/flows/vibe_check.py
@@ -175,182 +175,166 @@ async def process_single_concept(
     """
     Process inference for a single concept.
 
-    This task is designed to be isolated - if it fails, it won't affect the other
-    concept processing tasks.
+    Failures are left to propagate, so that Prefect retries the task and the flow can
+    record the concept as failed. Concepts are still isolated from each other: the flow
+    collects each task's outcome separately, so one failing concept doesn't stop the
+    others.
     """
     logger = get_logger()
-    try:
-        wikibase = WikibaseSession(
-            username=wikibase_config.username,
-            password=wikibase_config.password.get_secret_value(),
-            url=str(wikibase_config.url),
-        )
-        concept = await wikibase.get_concept_async(wikibase_id)
-        logger.info(f"Loaded concept: {concept}")
+    wikibase = WikibaseSession(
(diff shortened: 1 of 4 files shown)


jamesgorrie (approved):
nice, thready!

PR #1248, opened 2026-08-13
Exploratory analysis of the three climate justice classifiers

AnneIsARealProgrammerNow:
Word-count analysis of the three new justice classifiers for the Topics blog post. Everything lives under `scripts/climate_justice_analysis/`; the write-up is [FINDINGS.md](scripts/climate_justice_analysis/FINDINGS.md) and the ten charts are in `figures/`.

**Written by Claude (Anthropic) with @AnneIsARealProgrammerNow.** Every number comes from the scripts in this directory and regenerates from them, but the analysis and framing are machine-written and have not been independently reviewed. Treat it as a starting point for discussion, and check anything you intend to publish against the source data.

| Concept | Label | Body-text passages |
|---|---|---:|
| `Q32` | climate justice | 420,780 |
| `Q911` | distributive justice | 748,528 |
| `Q912` | procedural justice | 230,667 |

## What's here

`pull_data.py` (Snowflake → gitignored `data/`) → `text_stats.py` + `specificity.py` (→ `results/`) → `viz.py` (→ `figures/`). `house_style.py` holds the CPR palette, typography, logo and caption conventions. `viz.py` carries a PEP 723 header so it resolves its own dependencies rather than adding any to the shared repo environment.

## What it found

**The three classifiers separate cleanly.** Q32 is recognition and rights (`women`, `gender`, `rights`), Q912 is the institutional furniture of consultation (`stakeholders`, `consultation`, `civil society`, `local communities`), Q911 is mitigation vocabulary (`energy`, `emissions`, `renewable`).

**Q911 reading as mitigation language is a finding, not a defect.** It says *where* distributive claims sit in this corpus — in decisions about who builds and pays for the transition — rather than in explicit talk of payment or risk-sharing. A keyword yardstick (§2, all four lexicons printed in full) would have recovered most of Q912, two-thirds of Q32 and under a third of Q911. Whether the rest is signal or noise needs a labelled precision audit; word counts cannot settle it.

**Q32 behaves as an umbrella, by design.** P(Q32) is 47% across justice passages but 78% once both other classifiers fire. The concept store confirms it: Q911 and Q912 are formally `subconcept_of` Q32.

**Procedural justice travels with people; distributive travels with no one.** Against the just transition and impacted group concept families, Q912 is enriched 10–12× for legal safeguards, social inclusion and indigenous people, while Q911 sits flat at 2–3.5× against everything.

**Both countries' current NDCs carry far more justice language than their long-term strategies** — Türkiye 51.9% vs 27.9%, Australia 45.2% vs 29.1%, holding on every classifier. Caveated in the figure: NDCs are short political documents, LT-LEDS long technical ones.

## The one result most likely to be misread

The raw justice share falls from 33.1% in 2016 to 26.9% in 2025. **That decline is composition, not discourse.** Hold the 2016 corpus mix fixed and 2025 reads **37.2%** — above 2016. Multilateral fund projects, the densest corpus at 55.3% and still rising, fell from 37.6% of volume to 13.2%, while UN submissions, the least dense and falling, went from 25.0% to 62.3% as national inventory reporting grew. A bare "share of climate discourse" line drawn from this corpus would show a fall that does not exist in any actual body of policy text.

## Method choices a reviewer should check

- **Litigation excluded everywhere.** All three specs carry `dont_run_on: ["sabin"]` and every Litigation document is a Sabin record, so its 0% is an absence of inference.
- **Corporate disclosure excluded from the time series** but kept in the corpus heatmap: that dataset has not been refreshed this year, so its passages all sit in the past and would distort a trend.
- **One- and two-word terms ranked in separate vocabularies.** Pooled, a bigram is always rarer than its parts and the prior shrinks it harder, so almost none survive.
- **Concept co-occurrence is lift, not raw rate**, so a large concept cannot dominate every column.

## Two data issues worth passing on

- **`DOCUMENTS.passage_count` is rolled up from the v1 passages table, not `PASSAGES`.** It matches `PASSAGES_V1` row counts exactly on every document checked and overstates the real v2 count 7–11× (Uganda NBSAP 4,996 vs 441). Unusable as a denominator — looks like a second v1 dependency beyond the documented `labels[].passages_id` one.
- **`geographies` values are prefixed** (`country::UGA`) and `countries`/`regions` hold names, not ISO-3, so the obvious filter silently returns nothing.

Also: `fetch_pandas_all` returns Snowflake 0/1 flags as **int8**, so a groupby `sum()` over more than 127 rows wraps negative. `pull_data.query()` now widens them.

## Notes on the diff

`data/` and `figures/*.svg` are gitignored. `text_samples.parquet` alone is ~290MB of raw passage text, and the SVGs were ~400KB each of machine-generated XML making up 95% of the diff. Both regenerate in one command.

Two `pyproject.toml` lines add `allocat` and `participat` to the typos allowlist — deliberate substring prefixes in a lexicon list that the hook rewrote to `allocate` and `participants`, silently changing what the probe matched. Worth knowing: that hook edits string literals in place, and it also rewrote `PNGs` to `ONGs` in a README before I reworded around it.

Pyright suppressions are scoped to the analysis scripts rather than weakening `[tool.pyright]` repo-wide.


AnneIsARealProgrammerNow:
All of this is Claude-written so far. I'm just playing with the data, in part to generate insights for a blog post, in part because we'll want to share findings like this at a conference (CC @kdutia )

demos/climate_justice_analysis/README.md
```diff
@@ -0,0 +1,107 @@
+# Climate justice classifier analysis
+
+Written by Claude (Anthropic) with Anne Sietsma, August 2026.
+
+Exploratory analysis of the three climate justice classifiers — `Q32` (climate
```
kdutia, line 5: maybe worth linking to the justn0w conference here?

kdutia:
@AnneIsARealProgrammerNow if you move this into a `demos/climate_justice_analysis` directory then happy to approve :)

kdutia (approved):

PR #1249, opened 2026-08-13
feat: add extend_existing_dataset prefect flow

ChloeBrennan-Card:
## what does this do

this moves the functionality of the previous `extend_existing_dataset` script to the operations layer with a Prefect flow wrapper and deployment. The script is kept for CLI if a local sample needs using (thinking about previous google sheets labeling, so thought this might still be useful?).

Initially I used the existing oversampling approach (over sample then de-duplicate with what's in Argilla based on the passage text), but then tried excluding what is already in Agilla at sampling based on the existing `LabelledPassage.id` (`Identifier.generate(text)`). This works, but a real passage ID from Snowflake would be the better long-term fix.

`run_sampling` died with `ValueError: No objects to concatenate` whenever both classifiers found zero positives so added a clearer error for that too.

## why was it needed?

to make extending existing data sets consistent with other knowledge graph bau flows

## how was it tested

created and ran some new unit tests.

I created a tiny dataset of 10 locally which I then extended by 5 locally for unused Q1189 in [this](https://wandb.ai/climatepolicyradar/Q1189/runs/v8z57wwa/overview?nw=nwuserchloebrennancard) run with the new `n_excluded_passages` config parameter. You can also see the 15 in Argilla.

## AI attribution

claude helped write the unit tests for speed and check that the identifier was consistent and stable.


flows/extend_existing_dataset.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+@flow
+async def extend_existing_dataset(
+    wikibase_id: Annotated[
+        WikibaseID,
+        Field(description="The Wikibase ID of the concept whose dataset to extend"),
+    ],
+    wandb_artifact_path: Annotated[
+        Optional[str],
```
kdutia: nitpick: no need for `Optional` in modern python. use `str | None` instead

flows/extend_existing_dataset.py
```diff
@@ -0,0 +1,201 @@
...
+        bool,
+        Field(
+            description="Whether to attach LLM ensemble predictions as annotator "
+            "suggestions. Defaults to False."
+        ),
+    ] = False,
+    aws_env: AwsEnv = AwsEnv.production,
+    config: Optional[Config] = None,
+) -> int:
+    """Add more labelling passages to a concept's existing Argilla dataset."""
```
kdutia: could you please add a summary of the approach in the docstring?

flows/extend_existing_dataset.py
```diff
@@ -0,0 +1,201 @@
...
+        wandb_artifact_path = await sample(
+            wikibase_id=wikibase_id,
+            dataset_name=dataset_name,
+            sample_size=resolved_sample_size,
+            min_negative_proportion=min_negative_proportion,
+            max_negative_proportion=max_negative_proportion,
+            corpus_types_include=corpus_types_include,
+            corpus_types_exclude=corpus_types_exclude,
+            max_size_to_sample_from=max_size_to_sample_from,
+            exclude_passage_ids=sorted(existing_passage_ids),
```
kdutia: why do we need to sort here?

flows/sample.py
```diff
@@ -149,6 +153,13 @@ async def sample(
             description="Corpus types to exclude. Can be specified multiple times.",
         ),
     ] = None,
+    exclude_passage_ids: Annotated[
+        Optional[list[str]],
```
kdutia:
```suggestion
        list[str] | None,
```

knowledge_graph/operations/sample.py
```diff
@@ -97,6 +98,17 @@ def run_sampling(
             f"{len(dataset)} passages remain"
         )
 
+    if exclude_passage_ids:
+        before = len(dataset)
+        passage_ids = dataset["text_block.text"].map(Identifier.generate)
```
kdutia, line 106: this relies on `Identifier.generate(text)` being the way that `LabelledPassage.id` is generated which is true, but would be good to make explicit with a short comment

knowledge_graph/operations/sample.py
```diff
@@ -194,17 +207,16 @@ def run_sampling(
                 positive_samples_list.append(sampled_df)
 
         # Combine positive samples
+        if not positive_samples_list:
+            raise ValueError(
+                f"No positive passages found for {wikibase_id} in {len(dataset)} "
+                f"candidates: neither the keyword nor the embedding classifier matched "
+                f'anything. Try dataset_name="combined" for a much larger pool, or '
+                f"check the concept's labels in Wikibase."
+            )
```
kdutia, line 219: nice!

scripts/argilla/extend_existing_dataset.py
```diff
@@ -25,6 +20,11 @@ def main(
     limit: int | None = typer.Option(
         130, help="Limit the number of passages loaded to Argilla."
     ),
+    require_full_limit: bool = typer.Option(
+        False,
+        help="Fail, rather than warn and add fewer, when fewer than --limit of the "
+        "sampled passages are new.",
+    ),
```
kdutia: this _might_ be clearer named something like `raise_on_insufficient_passages` or `raise_when_limit_not_met` as it tells you _what_ will happen, as well as the conditions that need to be met. the current variable name doesn't really tell you what happens if the limit is not met

kdutia (approved):
nice work! a few small comments. it'd be great if you could address these but approving to unblock when they're done (and because i might need to take the afternoon off 🤧)

PR #1251, opened 2026-08-18
reads all alias language buckets from wikibase

abby11a:
## Description

`_parse_wikibase_entity` hardcoded the `en` alias bucket, so any alias tagged with another language tag was silently dropped before reaching the `Concept`.

Wikibase keeps aliases in buckets keyed by language tag. British spellings have been added under `en-gb` - [Q1432 `license`](https://climatepolicyradar.wikibase.cloud/wiki/Item:Q1432) has `licence` and `licencing` there - so none of them were matchable.

Fixes [FUS-252](https://linear.app/climate-policy-radar/issue/FUS-252).

```python
# before
for alias in entity.get("aliases", {}).get("en", [])
if alias.get("language") == "en"

# after
for language in ALIAS_LANGUAGES          # ("en", "en-gb", "en-us", "mul")
for alias in aliases.get(language) or []
```

Preferred label and description deliberately stay `en`-only, so a regional variant can't override the canonical preferred label.

## Impact

28 concepts gain 92 labels in total — `labour`/`labourer`, `neighbourhood`, `marginalised`, `programme`, `centre`, `fertiliser`, plus `licence`/`licencing` on Q1432.

**None of the 28 are currently in `v2/production.yaml`**, so no deployed classifier changes and there's no re-inference cost from merging. They'll pick the new labels up whenever they're next trained. Classifier IDs do change (they're derived from labels), so the affected concepts will need retraining and a spec refresh at that point.

## Testing

- three new tests in `tests/test_wikibase.py`: all English buckets read, non-English (`fr`) still ignored, entity with no aliases
- `pytest tests/test_wikibase.py tests/test_concept.py tests/test_classifier.py tests/operations`  → 170 passed, 1 skipped
- verified against live Q1432: `"an environmental licence was granted"` returns a span where it previously returned `[]`

### DIFF
--- a/knowledge_graph/wikibase.py
+++ b/knowledge_graph/wikibase.py
@@ -39,10 +39,7 @@ MAX_RETRIES = 5
 RETRY_INITIAL_WAIT = 1.0
 RETRY_MAX_WAIT = 60.0
 
-# Wikibase stores labels and aliases in buckets keyed by language tag. Concepts carry
-# their English variants across several of these, eg British spellings like "licence"
-# are tagged "en-gb", so we read all of them into the concept's alternative labels.
-# "mul" is Wikibase's tag for values which apply across languages.
+# Aliases are bucketed by language tag, eg "licence" sits under "en-gb" not "en".
 ALIAS_LANGUAGES: Final[tuple[str, ...]] = ("en", "en-gb", "en-us", "mul")
 
 
--- a/knowledge_graph/wikibase.py
+++ b/knowledge_graph/wikibase.py
@@ -39,6 +39,12 @@ MAX_RETRIES = 5
 RETRY_INITIAL_WAIT = 1.0
 RETRY_MAX_WAIT = 60.0
 
+# Wikibase stores labels and aliases in buckets keyed by language tag. Concepts carry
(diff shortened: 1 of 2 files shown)


jamesgorrie (commented):
Nice tests! And simple solution.

Q (probably due to my lack of context): this method [is used extensively through the codebase via `get_concepts_async`](https://github.com/search?q=repo%3Aclimatepolicyradar%2Fknowledge-graph+get_concepts_async&type=code) <= do we know if it will have any unintended affects there?

kdutia (approved):
thank you!

abby11a:
@jamesgorrie Thanks! Great question.

It does feed everything that fetches concepts, but only 28 concepts have aliases outside the en bucket, so those are the only ones that change at all. Everything else parses exactly as before.

The main thing I wanted to rule out was a new British label clashing with a negative label and accidentally cancelling out matches. I checked the whole store and there aren't any, and if there ever was one the new label would just get ignored rather than breaking anything.

Nothing changes on its own either, since the new labels only take effect once a concept is retrained. None of the 28 are in the prod spec at the moment anyway. I did run train Q1432 locally to check, and it now matches licence and licencing where it used to return nothing :)

PR #1252, opened 2026-08-18
keyword classifiers are optionally flexible to plurals and superscripts/subscripts

kdutia:
See context [in Notion](https://app.notion.com/p/climatepolicyradar/keyword-classifiers-that-are-sophisticated-enough-to-ignore-typos-plurals-subscripts-27f9109609a48040a1c2d78b57f79200?d=ce09109609a483338a0e83b4bfcfa25e).

I also experimented with an edit-distance (rather than dictionary) based typo approach but it didn't perform well.

## Test results

Running the two modes here plus typos detection on a bunch of concepts for which we have keyword classifiers, these were the results. I'm assuming the `fold_subscripts` result changed nothing because we've already got subscripts in the concept store where relevant!

| Variant | ΔF1 passage | ΔF1 span (0.5) | Better/Same/Worse (span) | Predict time |
|---|---|---|---|---|
| `fold_subscripts` | +0.0000 | +0.0000 | 0 / 56 / 0 | 0.04s |
| `match_word_forms` | −0.0001 | **+0.0030** | 15 / 35 / 6 | 0.04s |
| `typo_tolerance_1` (removed) | +0.0060 | +0.0026 | 15 / 26 / 15 | 0.78s |

The `typo_tolerance_1` config performed on average about the same (15 better, 15 worse, 26 the same) – but for far more time, and there were a couple of catastrophic cases (`oil and` & `oil sand`)

### `match_word_forms` by concept

The `match_word_forms` setting turns single labels into duplicate, but not vice-versa. Claude foudn that 29.9% of labels are already plural though. **I'm assuming this is to account for _not_ having this feature so far?**.

All performance gains are recall-driven, which is what we'd expect.

I'd suggest retraining all of the classifiers down to `Q979 | heavy precipitation` (~1%+ more recall), although we could go as far as `Q385 | drought`.

| Concept | Label | ΔF1 span | ΔF1 passage | ΔPrec | ΔRec |
|---|---|---|---|---|---|
| Q695 | youth | **+0.0357** | +0.0073 | +0.0135 | +0.0543 |
| Q374 | extreme weather | **+0.0235** | +0.0130 | +0.0032 | +0.0392 |
| Q2420 | energy supply mitigation | **+0.0235** | −0.0059 | +0.0033 | +0.0377 |
| Q1275 | subsidy removal | +0.0173 | 0.0000 | +0.0175 | +0.0171 |
| Q1273 | loan | +0.0115 | 0.0000 | +0.0106 | +0.0125 |
| Q1280 | tradable permit | +0.0110 | 0.0000 | −0.0035 | +0.0230 |
| Q979 | heavy precipitation | +0.0075 | 0.0000 | +0.0041 | +0.0095 |
| Q567 | renewable energy | +0.0068 | 0.0000 | +0.0074 | +0.0063 |
| Q47 | just transition | +0.0065 | 0.0000 | +0.0056 | +0.0073 |
| Q69 | green jobs | +0.0065 | 0.0000 | +0.0036 | +0.0093 |
| Q618 | solar energy | +0.0062 | 0.0000 | +0.0040 | +0.0083 |
| Q639 | oil | +0.0056 | 0.0000 | +0.0020 | +0.0092 |
| Q385 | drought | +0.0054 | 0.0000 | +0.0054 | +0.0055 |
| Q650 | gas | +0.0051 | **−0.0186** | −0.0054 | +0.0147 |
| Q684 | indigenous people | +0.0040 | 0.0000 | −0.0023 | +0.0098 |
| Q2008 | water adaptation | −0.0002 | +0.0033 | −0.0007 | 0.0000 |
| Q557 | adaptation/resilience | −0.0007 | 0.0000 | −0.0018 | 0.0000 |
| Q1834 | infrastructure and settlements adaptation | −0.0009 | 0.0000 | −0.0021 | 0.0000 |
| Q661 | coal | −0.0011 | 0.0000 | −0.0024 | 0.0000 |
| Q1833 | health adaptation | −0.0011 | 0.0000 | −0.0034 | 0.0000 |
| Q638 | fossil fuel | −0.0039 | −0.0052 | −0.0079 | 0.0000 |

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -15,3 +15,8 @@ extend-ignore-re = [
   # block-scoped: `# spellchecker:off` ... `# spellchecker:on`
   "(?s)(#|//)\\s*spellchecker:off.*?(#|//)\\s*spellchecker:on",
 ]
+
+# Domain abbreviations which typos would otherwise "correct".
+[default.extend-words]
+fpr = "fpr"
+FPR = "FPR"
--- a/_typos.toml
+++ b/_typos.toml
@@ -1,0 +1,17 @@
+# Configuration for the `typos` spell checker (crate-ci/typos), run via pre-commit.
+#
+# Enables the `spellchecker:off` / `spellchecker:on` block markers, so that deliberate
+# misspellings can be fenced off in place rather than corrected.
+#
+# This matters more than it looks: typos *errors* on ambiguous corrections but silently
+# rewrites unambiguous ones. Tests which assert that a wrong spelling is NOT matched
+# were therefore being edited by the hook until the assertion no longer tested
+# anything - e.g. ("branch", matches=["branches"], does_not_match=["branchs"]) was
+# rewritten to does_not_match=["branches"], contradicting the line above it.
+[default]
+extend-ignore-re = [
+  # line-scoped: trailing `# spellchecker:disable-line`
+  "(?Rm)^.*(#|//)\\s*spellchecker:disable-line$",
+  # block-scoped: `# spellchecker:off` ... `# spellchecker:on`
+  "(?s)(#|//)\\s*spellchecker:off.*?(#|//)\\s*spellchecker:on",
+]
--- a/knowledge_graph/classifier/__init__.py
+++ b/knowledge_graph/classifier/__init__.py
@@ -101,7 +101,12 @@ def create_classifier(
     """
 
     try:
-        classifier_class = __getattr__(classifier_type)
+        # Classifiers which are imported at the top of this module never reach the
+        # lazy __getattr__ hook, so look in the module globals first. Without this,
+        # `--classifier-type KeywordClassifier` fails.
+        classifier_class = globals().get(classifier_type) or __getattr__(
+            classifier_type
+        )
         return classifier_class(concept=concept, **classifier_kwargs)
 
     except (ImportError, AttributeError) as e:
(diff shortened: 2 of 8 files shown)


AnneIsARealProgrammerNow:
> Claude foudn that 29.9% of labels are already plural though. I'm assuming this is to account for not having this feature so far?.

Yep! 

Overall, typos aren't worth it, but plurals are. The numbers to me suggest it's worth it to turn this on by default, though I agree probably, too, that it's probably not worth re-training beyond heavy precipitation purely for this.

Would the logic also carry through for `not to be confused with` and `negative label` keywords?

arminellovell-source:
Cool thanks Kalyan! Ok agree that let's not bother with the typos (thanks for exploring it anyway). But lets keep updates to plurals (as default) and sub/superscripts. I'll update the concept store style guide and documentation to reflect these updates. And will add into our plan for two sprints time (because of leave) to rerun/retrain relevant classifiers.

kdutia:
@AnneIsARealProgrammerNow 

> The numbers to me suggest it's worth it to turn this on by default

I hadn't spotted that – I guess we might need a process update for figuring out when to turn it off?

> Would the logic also carry through for `not to be confused with` and `negative label` keywords?

Yes – the switch turns on/off the feature for both positive and negative labels for a concept symmetrically

AnneIsARealProgrammerNow:
> I hadn't spotted that – I guess we might need a process update for figuring out when to turn it off?

I suspect this  is best done at the concept level, rather than at the classifier level, particularly as more concepts end up feeding multiple classifiers. 

Would need to have a closer look at failure modes to say anything more than that.

juliesaigusaCPR (approved):

abby11a (approved):

PR #1254, opened 2026-08-19
docs: add operational runbook

ChloeBrennan-Card:
Adds `RUNBOOK.md` — operational procedures for the knowledge graph: sampling evaluation sets in Argilla, the classifier training and promotion loop, and how to check whether a deploy has landed.

Every command, parameter, log line and error path in it was checked against the source rather than written from memory.

## Prefect-first

Each section leads with the deployment. The few steps that genuinely have no deployment — classifier metadata (`flows/classifier_metadata.py` has no `@flow`) and listing Argilla datasets — say so explicitly, so the local CLI route reads as deliberate rather than as an oversight.

## Things it documents that weren't written down anywhere

- **`vibe-check-inference` does not evaluate a classifier you just trained.** It builds one from `ClassifierFactory` defaults every run (`flows/vibe_check.py:247-259` → `knowledge_graph/operations/train.py:659-663`) and never downloads a trained model from W&B, so for almost every concept it vibe-checks a `KeywordClassifier`. Only the four concepts in `bespoke_classifier_map` get anything else.
- **The Docker classifier-spec mount needs `/app/flows/classifier_specs/v2`.** `Dockerfile:27` sets `WORKDIR /app` and `SPEC_DIR` is relative, so a root-level `/flows/...` mount is written to by nothing.
- **`topic_pipeline` no longer runs indexing** (removed in f7e5a82). `index` remains a separate flow with its own deployment.
- **A deployment's `sha:` tag proves which commit it carries**, not the image tag — that's the hand-bumped `pyproject.toml` version, so two merges without a bump share a tag.
- **CI ignores markdown-only changes** (`paths-ignore: ['**/*.md']`), so a docs-only merge produces no CI run and therefore no deploy.
- The `sync-classifiers-profiles` deployment is named, so the existing "let the profiles sync pipeline handle it" guidance now has a command attached.

Also corrects the `sample` and `extend_dataset` rows in `flows/README.md` to match current flow behaviour.

## Known follow-ups, not fixed here

- `README.md:114-135` still describes indexing as part of `topic_pipeline`. The runbook states the steps itself rather than relying on that section.
- `scripts/scripts.just:22` passes the ID positionally to `uv run deploy new`, which only accepts `--wikibase-id`, so `just deploy-classifier` errors out. The runbook documents the working plural `just deploy-classifiers` instead.


ChloeBrennan-Card:
Closing — superseded. Reopening a fresh PR from the same branch.

PR #1255, opened 2026-08-19
docs: add runbook and pipeline diagram

ChloeBrennan-Card:
adds runbook and diagram. Claude created the skeleton of the runbook for speed and I verified and changed the bits I'm familiar with (evaluation set and training). Would appreciate review of the promote, demote and update specs and pipeline parts (from Enrichment maybe?).


RUNBOOK.md
```diff
@@ -0,0 +1,402 @@
...
+Expect, in order:
+
+| log line | meaning |
+| --- | --- |
+| `✅ Found N passages already in dataset '<WIKIBASE_ID>'` | current record count |
+| `Excluded N passages already in the labelling dataset` | the pool filter worked. **If this line is missing entirely, the exclusion list was empty - stop and investigate** |
+| `📊 Sampled X positive passages, Y negative passages` | the split actually drawn |
+| `<n>/<n>> input passages remaining after deduplication` | more passages as intended |
+| `✅ Successfully added <n>> passages to dataset '<WIKIBASE_ID>'` | done |
+
```
kdutia: this seems like a sign that the docs are too brittle to change – a bit like writing `if` statements or tests based on log lines in code. 

could it be worth updating the code to more dramatically warn instead – and include the caveat around "you might want to stop and investigate" directly in a log message?

RUNBOOK.md
```diff
@@ -0,0 +1,402 @@
...
+`AWS_ENV` must also be set. `Config` reads it directly, and `.env.example` ships it empty,
+which fails with `ValueError: '' is not a valid AwsEnv` before anything else happens.
+
+Also needs `uv run prefect cloud login`. `Config.create()` reads a Prefect Variable for the cache
+bucket. If AWS calls fail with `TokenRetrievalError`, re-run `aws sso login --profile prod`.
+
+## Classifiers
+
+Training uses the passages labelled in Argilla for the concept, so if a concept has too little
+training data, [Add more labelling passages to a concept that already has an evaluation set](#add-more-labelling-passages-to-a-concept-that-already-has-an-evaluation-set).
```
kdutia: this is confusing adding to the training set with adding to the evaluation set. we don't have a neat way of adding to the evaluation set yet

RUNBOOK.md
```diff
@@ -0,0 +1,402 @@
...
+
+### Train a classifier
```
kdutia, line 122: I think it'd be better to suggest using scripts/train.py here with the appropriate `--compute` flag. The CLI is self-documenting, so not too much extra documentation should be needed here

RUNBOOK.md
```diff
@@ -0,0 +1,402 @@
...
+
+**1. Train an initial classifier** on a GPU. `--classifier-type` overrides whatever
+`ClassifierFactory` would pick:
+
+```bash
+uv run train \
+  --wikibase-id <WIKIBASE_ID> \
+  --classifier-type BertBasedClassifier \
+  --compute remote-gpu
+```
```
kdutia: let's find some time to go through this process (but good job giving it a go!). this should cover training an `LLMClassifier`, using the `AutoLLMClassifier` to iterate on its prompt, then generating training data. this is what's echoed in the Whimsical that you've included a screenshot of here – which maybe could serve as documentation itself, rather than repeating the process here?

training_pipeline.png
kdutia, line 1: i think it's best to either:

- delete this screenshot, and link to the whimsical in `RUNBOOK.md`, or
- remove references to the whimsical, and update the `RUNBOOK` (or a separate file) to have a description of the process.

What do you think? Which would be easier for you had you never run any steps in this process before?

kdutia (commented):
good job! sorry for the very late review – let's chat about this this week if helpful

training_pipeline.png
ChloeBrennan-Card: James specifically requested the whimsical be in the repo, but happy to link instead if you disagree

RUNBOOK.md
ChloeBrennan-Card: Had a go at updating, but can discuss and if we decide to remove the screenshot then can add more detailed steps?

RUNBOOK.md
```diff
@@ -130,121 +130,37 @@ bucket. If AWS calls fail with `TokenRetrievalError`, re-run `aws sso login --pr
...
+1. **Train an `LLMClassifier`** for the concept, a prompt-based classifier
+   (`knowledge_graph/classifier/large_language_model.py`).
+2. **Iterate the prompt with `AutoLLMClassifier`** (`knowledge_graph/classifier/autollm.py`). Its
+   `fit()` runs `n_trials` optimisation trials against the concept's labelled passages, using an
+   optimiser model to rewrite the labelling guidelines and keeping the best trial by f-beta score.
+3. **Generate training data** — run the tuned LLM classifier over unlabelled passages (the predict
+   flow / `run_prediction`) to produce LLM-labelled passages, uploaded as a W&B artifact.
+4. **Train a BERT classifier on that data** — point a per-concept YAML config's
+   `training_data_wandb_path` (`BERTClassifierConfig`) at that artifact, then
+   `uv run train --from-yaml-config <config> --classifier-type BertBasedClassifier --compute remote-gpu`.
```
kdutia: nearly! two extra steps:

- after 2, someone from programmes team has a look at the final prompt and tweaks it a bit. this usually gives us a few more % F1
- after that, we vibe check the LLMClassifier as an extra safeguard before running it to create training data (expensive)

this can fall down if we don't get enough positives for the BERT classifier, which defaults to a balanced positive/negative dataset based on previous experiments. that's an issue with the sample step that it might be worth looking into in the future.

@AnneIsARealProgrammerNow wrote a really detailed summary of the process from the programmes team angle [here](https://app.notion.com/p/climatepolicyradar/Concept-development-LLM-classifiers-31f9109609a480c89a81ed4a7169ccb7). i think both can exist as they're for different audiences, but it might be worth linking to that one from here

training_pipeline.png
kdutia: > James specifically requested the whimsical be in the repo

i think this means we can link to the Whimsical? the important thing IMO is that the diagram is editable from here, which obviously a screenshot doesn't achieve

RUNBOOK.md
kdutia: i think linking is sensible – a screenshot is guaranteed to go out of date, and we need the link from the docs to be able to update the screenshot without having to recreate it from scratch. this process is likely to change a lot!

ChloeBrennan-Card: Good point! 🤓

kdutia (approved):

odrakes-cpr (approved):

PR #1256, opened 2026-08-20
adds integration tests for build_dataset and vibe_check flows

abby11a:
## What
Adds integration tests for `build_dataset` and `vibe_check`, the two BAU flows whose existing tests mocked out their own internal wiring rather than just external services.

## Why
- `build_dataset`: the existing flow tests bypassed `_set_up_build_dataset_environment` entirely, so `get_snowflake_credentials`'s SSM lookups never ran in any flow test.
- `vibe_check`: the existing isolation test (one concept failing shouldn't affect others) only checked a mock's call count, not that the real S3 read/write paths (`load_passages_dataset`, `push_object_bytes_to_s3`, `get_bucket_name_from_ssm`) actually worked.

## What's still mocked, deliberately
Snowflake, Wikibase, the embedding model, and W&B - genuine external services. Everything else (S3, SSM) runs against moto for real.

## Test plan
- [x] `uv run --extra dev --extra transformers pytest tests/flows/test_build_dataset_flow.py tests/flows/test_vibe_check.py -v`
- [x] All existing tests in both files still pass unmodified in behavior

### DIFF
--- a/tests/flows/test_build_dataset_flow.py
+++ b/tests/flows/test_build_dataset_flow.py
@@ -1,6 +1,7 @@
+import base64
 import io
 from contextlib import contextmanager
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
 
 import pandas as pd
 import pytest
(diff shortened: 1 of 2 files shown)


henry-cpr (approved):
Lgtm!

PR #1257, opened 2026-08-25
chore: remove leftover topic indexing code

olaughter:
We have consolidated our inexers and so no longer run the index flow or code path. Here we clean it up and attempt to also purge all related code.

Judging by github action runtimes, this makes tests roughly x2 as fast to run. It also gives a minor improvement to setup burden & image size by removing the dependency.


THOR300 (approved):
Let's gooo! Amazing to get this cleared down 🙌 

Only reference I see remaining is the following config object which I believe we can remove: 
- INDEX_RESULTS_PREFIX / index_results_prefix

olaughter:
> * INDEX_RESULTS_PREFIX

Good catch will get this one cleared too

PR #1258, opened 2026-08-25
Feature/enri 1639 migrate topics pipeline off parseroutput to the simplified

olaughter:
We're cutting down on the data that we pass around in S3. This simplification lets us save storage as well as unblocks mapper -> processing interface. Here the sdk code is vendored, the extra fields are then removed, and associated test code cleaned up.

We have chosen to reproduce the models in the knowledge graph at this stage so we can break the dependency on the sdk. We're not looking at making a new, common abstraction at this time but also haven't necessarily ruled it out in the future.

### DIFF
--- a/flows/models.py
+++ b/flows/models.py
@@ -1,0 +1,314 @@
+from datetime import date
+from enum import Enum
+from typing import (
+    Annotated,
+    Any,
+    Generic,
+    List,
+    Optional,
+    Protocol,
+    Sequence,
+    Tuple,
+    TypeVar,
+    Union,
+)
+
+from pydantic import BaseModel, Field
+
+
+class VerticalFlipError(Exception):
+    """Exception for when a vertical flip fails."""
+
+    pass
+
+
+class BlockType(str, Enum):
+    """
+    List of possible block types from the PubLayNet model.
+
+    https://layout-parser.readthedocs.io/en/latest/notes/modelzoo.html#model-label-map
+    """
+
+    TEXT = "Text"
+    TITLE = "Title"
+    LIST = "List"
+    TABLE = "Table"
+    TABLE_CELL = "TableCell"
+    FIGURE = "Figure"
+    INFERRED = "Inferred from gaps"
+    AMBIGUOUS = "Ambiguous"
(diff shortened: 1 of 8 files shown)


flows/models.py
```diff
@@ -0,0 +1,314 @@
...
+
+
+class _BaseParserOutputFieldsMixin(BaseModel, Generic[PDFDataT, PDFTextBlockT]):
+    """Shared fields and methods for BaseParserOutput* classes."""
+
+    document_id: str
+    document_content_type: Optional[str] = None
+    languages: Optional[Sequence[str]] = None
+    html_data: Optional[HTMLData] = None
+    pdf_data: Optional[PDFDataT] = None
```
olaughter, line 281: The minimal version of the fields, see for comparison: [the sdk version](https://github.com/climatepolicyradar/cpr-sdk/blob/f74ae76c6cb0cd3706a56ece77e8a8eb11bf35ad/src/cpr_sdk/parser_models.py#L315) content_type is produced within processing & is still used here to check for html data, although I have suspiscions we can remove this too with some more refactoring here

THOR300: Hey hey! Are these minimal fields distinct from those that we expect to output from the document processing pipeline? 

E.g. `content_type`, `cdn_object` we'd continue to output from processing as they're required to replace the rds write back but here in the topics pipeline we just don't load them?

THOR300 (approved):
Huge! Well this is amazingly clean! 

I suppose there's no change to the output of the topics pipeline so nothing in snowflake breaks because of this 🙌

flows/models.py
olaughter: We're safe here because we don't actually pass any of them through, we make a new object in inference that really just needs the document_id, text block_id and the topics. The others are all still used here to determine what gets run. I'd love to come back and get rid of the html_data & content_type, we only have these still because there are some html documents stuck in the system that we havent been able to clear out. Hoping we'll be able to before long though and we can take a few more out

olaughter: From how things look to me, keeping them won't block work upstream, but happy to talk through this more if there's any uncertainty on your side on this

PR #1259, opened 2026-08-26
chore: add lifecycle policy for ECR

olaughter:
This involves importing the resources and applying a tag based & untagged based policy. I've already applied all of them as the infra isnt run in ci.

### DIFF
--- a/infra/__main__.py
+++ b/infra/__main__.py
@@ -3,15 +3,9 @@ import json
 import pulumi
 import pulumi_aws as aws
 
-# One stack per AWS environment, named after that environment. Each stack must be
-# run against the matching account's credentials, eg `AWS_PROFILE=labs pulumi up
-# -s labs`.
 stack = pulumi.get_stack()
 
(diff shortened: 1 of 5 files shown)


infra/__main__.py
```diff
@@ -1,58 +1,122 @@
...
-# Grant the labs AWS account read-only access to the feather files bucket
-labs_cross_account_policy = aws.iam.get_policy_document(
-    statements=[
-        aws.iam.GetPolicyDocumentStatementArgs(
-            sid="LabsCrossAccountReadOnly",
-            effect="Allow",
-            principals=[
-                aws.iam.GetPolicyDocumentStatementPrincipalArgs(
-                    type="AWS",
-                    identifiers=[f"arn:aws:iam::{labs_aws_account_id}:root"],
```
olaughter, line 39: This was erroring because the f string wasnt handling the secret value, so I've updated below using `get_policy_document_output` & apply instead

THOR300 (approved):
LGTM!

PR #1263, opened 2026-09-02
add llm reviewed status metadata to argilla

kdutia:
I realised that when pre-labelling Argilla records with an LLM, there's no way to tell the difference between a record that hasn't been labelled and a record that has been labelled as having no spans.

This PR adds an `llm_prelabelled` boolean metadata key to every record to say whether it's been labelled or not, which can be used to sort and filter records in the Argilla UI.

Follows #1243.


henry-cpr (approved):
lg!

PR #1269, opened 2026-09-16
feat: improve inference output data

olaughter:
Takes the existing data being written out and adds:
- The `etag` of the document — snowflake uses this as a content_key
- The `last_modified` of the document — this helps us validate which document ran when checking downstream
- The `idx` of the passage letting us preserve & validate ordering downstream

For v2 passages this now also no longer writes out blocks with no spans, which will roughly half the sizes

## Still to do

- [ ] Update current outputs to reflect new structure


kaparker (approved):
Looks good. 
Will there be a way to know that a document ran inference successfully but just didn't get any spans vs. was not successful in inference so has no file if we can no longer check if the file exists in s3?

flows/inference.py
```diff
@@ -569,24 +582,24 @@ def _stringify(text: list[str]) -> str:
 
 def document_passages(
     document: BaseParserOutput,
-) -> Generator[tuple[str, str], None, None]:
+) -> Generator[tuple[str, str, Optional[int]], None, None]:
```
THOR300, line 585: **thought**: Just type hint as `None`?

THOR300 (approved):
This change looks great and I can see the downstream uses of this, questions from me are around how we intend to run the backfill (assuming a simple script over s3), and also the wider context of what we're specifically intending on doing with these fields (the ticket is a lil minimal)?

olaughter:
> This change looks great and I can see the downstream uses of this, questions from me are around how we intend to run the backfill (assuming a simple script over s3)

Yeah good question, a script was my original plan but I actually don't think its an option because the results lose track of which document they used. It also would be risky with making sure the state is perfect when a script would require different code paths. And the downstream snowflake update would be messy with the existing data. In the end I'm looking to just do a full rerun, its a bit more resource intensive but feels like the only approach that we could take in this case

> and also the wider context of what we're specifically intending on doing with these fields (the ticket is a lil minimal)?

Ah yeah the parent ticket has the opposite problem because claude wrote most of it so its mega verbose. Basically we were having problems reliably joining up labelled passages to regular passages, particularly when passages change as a subset did during the backfill. This gives us more concrete join keys — the specific file etag & last modified date for the document, and the specific idx for the passage.

With the idx we also can drop the ones that are empty making the files smaller and easier to manage in snowflake.

olaughter:
> Looks good. Will there be a way to know that a document ran inference successfully but just didn't get any spans vs. was not successful in inference so has no file if we can no longer check if the file exists in s3?

I've left it still outputting the file for now if it is successful but has no spans. Might be worth revisiting this point in the future though, ideally the prefect artefacts would tell us everything we need to know on that but I think they could do with some improvements generally.

THOR300:
Sounds great, thanks for the context!

PR #1272, opened 2026-09-16
chore: switch off v1 for new classifiers

olaughter:
We have some new classifiers that will be running soon, but have agreed these won't need to run through our v1 pipeline. This means we can disable this automation.

See: https://climate-policy-radar.slack.com/archives/C078K2QNU0L/p1789562372259799?thread_ts=1789561399.394719&cid=C078K2QNU0L

FIXES ENRI-1717


THOR300 (approved):
Confirming my understanding: 

Still running v1 and v2 passages through every passage we process as part of document processing, just turning off the automation that triggers a backfill / full run from a new classifier temporarily?

